PR #4, opened 2025-10-29
add hfopenllm_v2

sanchit-ahuja:


data/hfopenllm_v2/0-hero_Matter-0.2-7B-DPO
```diff
@@ -0,0 +1,101 @@
...
+    "source_organization_name": "Hugging Face",
+    "evaluator_relationship": "third_party"
+  },
+  "model_info": {
+    "name": "0-hero/Matter-0.2-7B-DPO",
+    "developer": "0-hero",
+    "inference_platform": "unknown",
+    "precision": "bfloat16",
+    "architecture": "MistralForCausalLM",
+    "params_billions": 7.242
```
akornilotrust: We do not use these 3 keys in the current schema

akornilo (approved):

PR #11, opened 2025-11-27
Automated workflow for syncing GitHub to HF dataset via GitHub actions.

nelaturuharsha:
In this PR, the following functionality is added:
When either a new leaderboard is added to data/ or a previous leaderboard is updated with new entries then the action is triggered.

The modified leaderboards are detected by git diff (and currently compare only the previous and current commit). These modified leaderboards (if previously in the hf dataset as a split, are then downloaded to the action env).

There are two cases:

**New Leaderboard**
The JSON(s) are aggregated into a single .parquet for the leaderboard, and is ready for upload to HF.

**Existing Leaderboard**
In case there are new JSON(s) added to a pre-existing leaderboard, they are appended to the downloaded .parquet of the leaderboard.

There are some nice features/fail-safes which can be described in detail if required. The following use-cases have been tested:
- Empty dataset, uploading new leaderboards
- Adding a new leaderboard
- Adding to pre-existing leaderboard.


evijit:
@nelaturuharsha -- the schema has changed. Will your dataset fields also change? If so, please revise

evijit:
@nelaturuharsha seems like there are merge conflicts still

PR #16, opened 2025-12-16
Moved inspect converter scripts from EvalHub repo to EEE

damian1996:


evijit:
@damian1996 pinging for viz

scripts/eval_converters/inspect/adapter.py
```diff
@@ -0,0 +1,159 @@
...
+            retrieved_timestamp=retrieved_unix_timestamp,
+            source_data=source_data,
+            source_metadata=source_metadata,
+            model_info=model_info,
+            evaluation_results=evaluation_results,
+            detailed_evaluation_results_per_samples=detailed_evaluation_results_per_samples
+        )
+        
+    def _load_file(self, file_path) -> EvalLog:
+        return read_eval_log(file_path)
```
jjallaire: Ideally we could get it so that you can pass header_only here (and not have to deserialize all of the samples which can be expensive for long agentic evals). We have a sample summaries cache which can provide individual samples scores. I'll propose a way to do this as a follow up shortly.

damian1996: Thanks for your insightful proposition! I've modified my script to use log and sample summaries. There is only a one problem for me. I've extracted a model name from two places:
1) from `EvalSpec` where is a full model path like`"model": "openai/azure/gpt-4o-mini"` or `"model": "anthropic/claude-sonnet-4-0"`
2) from `ModelOutput` in `Sample` details where is more detailed model name like `"model": "claude-sonnet-4-20250514"` or `"model": "gpt-4o-mini-2024-07-18"`

Could I get this second info from some other place to omit deserialization all of the samples (maybe I should deserialize only the first sample to extract this info)?

data/ai2_arc/anthropic/claude-sonnet-4-0/d548f20c-7c07-4c0e-8c77-39093f383a34.json
```diff
@@ -0,0 +1,108 @@
+{
+  "schema_version": "0.1.0",
+  "evaluation_id": "inspect_ai/anthropic/claude-sonnet-4-0/allenai/ai2_arc/1761000045.0",
```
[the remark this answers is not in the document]
damian1996: Fixed it, I've forgot to change this line. Thanks!

andrewtran117 (approved):

PR #23, opened 2026-01-20
Fast follow up on return type and ModelInfo

andrewtran117:
Quick follow up addressing @damian1996 's comments.
- ModelInfo type
- Return type tuple

Verified HELM:
```
uv run python scripts/validate_data.py data/helm_capabilities -s eval.schema.json

Validating 61 JSON files...
61 file(s) passed; 0 file(s) failed
```


damian1996 (approved):
LGTM

PR #27, opened 2026-01-31
Add tests for duplicate checker

mrshu:
* Cover normalized hashing with real data payloads

* Include path expansion and duplicate detection behavior

* Follow-up to https://github.com/evaleval/every_eval_ever/pull/20#issuecomment-3762654531


mrshu:
@andrewtran117 this is just a followup to your comment in the other PR -- I'd be happy to do it in other ways if that would be better :)

Thanks!

andrewtran117 (approved):

Issue #26, opened 2026-01-26
[Schema Modification] Handling instance level information and statistics


nelaturuharsha:
In the v0.1 version of the schema, instance level information are added to the schema for a given collection of evals as follows:

## Current Approach

```json
 "detailed_evaluation_results_per_samples": [
    {
      "sample_id": "test_1042",
      "input": "Question: Which of the following is a scalar quantity?",
      "ground_truth": "C",
      "response": "Scalar quantities have magnitude only. Speed is a scalar. The answer is C.",
      "choices": ["Velocity", "Force", "Speed", "Acceleration"]
    },
    {
      "sample_id": "test_1043",
      "input": "Question: The Treaty of Versailles was signed in which year?",
      "ground_truth": "C",
      "response": "The Treaty of Versailles was signed in 1919. The answer is C.",
      "choices": ["1914", "1918", "1919", "1939"]
    }
  ]
```

## Challenges

- With multiple evals in one collection, and with individual evals having large number of instances this would bloat the JSON with aggregate scores. 
- Instance information is accessed less often, so its better to have separation of concerns.
- We would like easier parsing for comparison/analysis and the like.

## Potential Solution Sketch
- Single {evaluation_id}_samples.jsonl co-located with the aggregate score JSON in the same sub-directory.

### Proposed Mandatory metadata fields per instance:

| Field | Type | Description | Mandatory? | Example | 
| :--- | :--- | :--- | :--- | :--- |
| **schema_version** | str | Version of the schema, useful to track if its up-to-date or needs updating. | Yes | v0.1 |
| **evaluation_id** | str | **Foreign Key** The same as in the aggregate JSON | Yes | "hfopenllm_v2/0-hero_Matter-0.2-7B-DPO/1762652579.4626381 (could also be the UUID open to discussion)". |
| **model_id** | str | Identifier of the model | Yes | meta-llama/Llama-3.2-1B-Instruct | 
| **evaluation_name** | str | The specific eval name, ideally unique | Yes | GSM8K |
| **sample_id** | str | Question id from the original dataset | Yes | gsm8k_0001 |
| **sample_hash** | str | hash(input + ground_truth) to ensure that the comparison is between the same sample across models etc., in case the sample_id is not consistent | Yes | a1b2c3d4e5f6 |
| **interaction_type** | str | single_turn, multi_turn, agentic | Yes | single_turn |

#### input Dict
* **raw** (str) - The raw input as defined in the eval.
[...]

PR #28, opened 2026-02-02
Changes to schema for v0.2

nelaturuharsha:
- Move source_data to evaluation_results

  * **source_data** - This data should be specified per eval, I have moved into evaluation_results and removed it from the root of the schema. Overrode the change from source_data with a contribution from @janbatzner 
  * **inference_engine** - converted to object and added version as an additional attribute.
  * **LLM as a judge** - llm_scoring added by @janbatzner 
  * **confidence_interval** - added object to score_details.
  * **detailed_evaluation_results_url** - removed this parameter as it was redundant
  * **detailed_evaluation_results_per_samples** - replaced with detailed_evaluation_results
  * Added instance_level_eval.schema.json for single_turn, multi_turn and agentic as described in #26

### DIFF
--- a/eval.schema.json
+++ b/eval.schema.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "version": "0.1.0",
+    "version": "0.2.0",
     "type": "object",
     "description": "Schema for storing and validating LLMs evaluation data, including model configuration, prompts, instances, Output, and evaluation metrics",
     "required": [
@@ -162,13 +162,18 @@
                             {
                                 "type": "object",
                                 "description": "Generic source data when neither URL array nor HuggingFace dataset applies",
+                                "additionalProperties": true,
                                 "required": [
-                                    "dataset_name"
+                                    "dataset_name",
+                                    "source_type"
                                 ],
                                 "properties": {
                                     "dataset_name": {
                                         "type": "string",
                                         "description": "Name of the source dataset"
+                                    },
+                                    "source_type": {
+                                        "const": "other"
                                     },
                                     "additional_details": {
                                         "$ref": "#/$defs/additional_properties_object"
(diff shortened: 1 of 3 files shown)


eval.schema.json
```diff
@@ -223,6 +324,31 @@
...
+                                        "type": "string",
+                                        "description": "Prompt template used for judging"
+                                    },
+                                    "temperature": {
+                                        "type": "number"
+                                    },
+                                    "expert_baseline": {
+                                        "type": "number",
+                                        "description": "Expert/human baseline score for comparison"
+                                    }
```
damian1996: What do you think about adding some additional_details field here? In the case if someone want to share more detailed config for a judge?

eval.schema.json
```diff
@@ -22,12 +21,16 @@
...
+            "type": "string",
+            "description": "Timestamp for when the evaluation was run"
+        },
         "retrieved_timestamp": {
             "type": "string",
             "description": "Timestamp for when this record was created - using Unix Epoch time format"
         },
         "source_data": {
-            "description": "Source of dataset used for evaluation. There are two options supported: HuggingFace dataset or url for other data source.",
+            "description": "Source of dataset used for evaluation. Three options supported: URL array, HuggingFace dataset, or generic source object.",
```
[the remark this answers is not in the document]
nelaturuharsha: Nope, this was a dangling reference - removed.

eval.schema.json
```diff
@@ -268,13 +394,32 @@
                                 "type": "object",
                                 "description": "Any additional details about the score",
                                 "additionalProperties": true
+                            },
+                            "max_tokens": {
```
damian1996: Do you think it is better place to put `max_tokens` field than `generation_config`? I've already put this field there, so we should choose one of them and remove the second one.

eval.schema.json
nelaturuharsha: Done!

instance_level_eval.schema.json
```diff
@@ -0,0 +1,273 @@
...
+                    "description": "Optional list of choices for multiple-choice questions",
+                    "items": {
+                        "type": "string"
+                    }
+                }
+            }
+        },
+        "output": {
+            "type": ["object", "null"],
+            "description": "Output data - only used for single_turn interactions, null for multi_turn/agentic",
```
damian1996: probably make `raw` subfield required

instance_level_eval.schema.json
```diff
@@ -0,0 +1,274 @@
...
+        },
+        "interaction_type": {
+            "type": "string",
+            "enum": ["single_turn", "multi_turn", "agentic"],
+            "description": "Type of interaction: single_turn for simple Q&A, multi_turn for conversations, agentic for tool-using agents"
+        },
+        "input": {
+            "type": "object",
+            "description": "Input data for the evaluation sample",
+            "required": ["raw", "formatted", "reference"],
```
damian1996: Are you sure that formatted should be required? It's nice to have but imo isn't necessary. If there are raw input + model given then we shoule be able to recreate formatted input version if needed.

instance_level_eval.schema.json
```diff
@@ -0,0 +1,318 @@
+{   "$schema": "http://json-schema.org/draft-07/schema#",
+    "version": "instance_level_eval_0.0.1",
```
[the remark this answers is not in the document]
nelaturuharsha: Noted!

janbatzner:
This looks great, I would say, we are ready to merge 🥇 :)

PR #43, opened 2026-02-16
Delete data directory

evijit:
Since we have now moved to HF


evijit:
PS we should also remove the validate data actions etc @yifanmai @nelaturuharsha, see PR #44

PR #52, opened 2026-02-23
Add metric identity and deterministic aggregate-instance linkage fields

yananlong:
This PR updates the evaluation result schemas to improve metric identity and enable deterministic aggregate <-> instance joins.

**Changes**
- Add `evaluation_results[].evaluation_result_id` (aggregate) and `evaluation_result_id` (instance) as a stable join key.
- Add optional metric identity fields under `evaluation_results[].metric_config`: `metric_id`, `metric_name`, `metric_kind`, `metric_unit`, `metric_parameters`.

**Metric identity guidance (all optional)**
- `metric_id`: stable identifier for joining/deduping/querying. Use a canonical global id when applicable (e.g. `accuracy`, `f1_macro`, `auroc`, `rmse`, `pass_at_k`). For benchmark/leaderboard-specific metrics, use a namespaced id (e.g. `rewardbench.overall`, `lmarena.elo`).
- `metric_kind`: normalized metric family used for safe aggregation (e.g. `accuracy`, `f1`, `auroc`, `rmse`, `mae`, `pass_rate`, `elo`).
- `metric_name`: display name for the metric (e.g. `Score`, `Accuracy`, `pass@1`).
- `metric_unit`: representation of the numeric values (e.g. `proportion`, `percent`, `points`, `ms`, `tokens`).
- `metric_parameters`: metric-specific configuration (e.g. `{ "k": 1 }` for `pass@k`).

**Deterministic `evaluation_result_id` recipe**
Suggested construction:
- `evaluation_result_id = "er_" + sha256(canonical_json(payload))`
- `canonical_json`: JSON with sorted keys and stable separators (no whitespace)
- `payload`: at minimum `{ evaluation_id, evaluation_name, source_data.dataset_name, metric_id, metric_kind, metric_unit, metric_parameters }` (include `generation_config` if you need to disambiguate multiple computations of the same metric)

**Before/after examples**

Example A (RewardBench overall, where `evaluation_name: "Score"` is ambiguous across sources):

Before:
```json
{
  "evaluation_name": "Score",
  "metric_config": {
    "evaluation_description": "Overall RewardBench Score",
    "lower_is_better": false,
    "score_type": "continuous",
    "min_score": 0.0,
    "max_score": 1.0
  },
  "score_details": { "score": 0.612599 }
}
```

After:
```json
{
  "evaluation_result_id": "er_<sha256(...)>",
  "evaluation_name": "Score",
  "metric_config": {
    "metric_id": "rewardbench.overall",
    "metric_name": "Score",
    "metric_kind": "preference_rate",
    "metric_unit": "proportion",
    "metric_parameters": {},
    "evaluation_description": "Overall RewardBench Score",
    "lower_is_better": false,
    "score_type": "continuous",
    "min_score": 0.0,
    "max_score": 1.0
  },
  "score_details": { "score": 0.612599 }
}
```

Example B (`pass@k`, where `k` is indispensable):

Before:
```json
{
  "evaluation_name": "multiple-js",
  "metric_config": {
    "evaluation_description": "pass@1 on multiple-js",
    "lower_is_better": false,
    "score_type": "continuous",
    "min_score": 0.0,
    "max_score": 1.0
  },
  "score_details": { "score": 0.189814 }
}
```

After:
```json
{
  "evaluation_result_id": "er_<sha256(...)>",
  "evaluation_name": "multiple-js",
  "metric_config": {
    "metric_id": "pass_at_k",
    "metric_name": "pass@1",
    "metric_kind": "pass_rate",
    "metric_unit": "proportion",
    "metric_parameters": { "k": 1 },
    "evaluation_description": "pass@1 on multiple-js",
    "lower_is_better": false,
    "score_type": "continuous",
    "min_score": 0.0,
    "max_score": 1.0
  },
  "score_details": { "score": 0.189814 }
}
```

### DIFF
--- a/eval.schema.json
+++ b/eval.schema.json
@@ -122,6 +122,10 @@
                     "score_details"
                 ],
                 "properties": {
+                    "evaluation_result_id": {
+                        "type": "string",
+                        "description": "Stable identifier for this metric result inside an evaluation run. Recommended deterministic join key for instance-level records."
+                    },
                     "evaluation_name": {
                         "type": "string",
                         "description": "Name of the evaluation"
@@ -154,6 +158,34 @@
                             "evaluation_description": {
                                 "type": "string",
                                 "description": "Description of the evaluation"
+                            },
+                            "metric_id": {
+                                "type": "string",
+                                "description": "Canonical metric identifier for cross-source normalization (e.g. accuracy, f1_macro, auroc, pass_at_k)."
+                            },
+                            "metric_name": {
+                                "type": "string",
+                                "description": "Display name for the metric (e.g. Accuracy, F1-macro, pass@1)."
+                            },
+                            "metric_kind": {
+                                "type": "string",
+                                "description": "Normalized metric family/type used for safe aggregation (e.g. accuracy, f1, auroc, rmse, mae, pass_rate, elo)."
+                            },
+                            "metric_unit": {
+                                "type": "string",
+                                "description": "Unit of the metric values (e.g. proportion, percent, points, ms, tokens)."
+                            },
+                            "metric_parameters": {
+                                "type": "object",
+                                "description": "Metric-specific parameters (e.g. {\"k\": 1} for pass@k).",
+                                "additionalProperties": {
+                                    "type": [
+                                        "string",
+                                        "number",
+                                        "boolean",
+                                        "null"
+                                    ]
+                                }
                             },
                             "lower_is_better": {
                                 "type": "boolean",
--- a/eval.schema.json
+++ b/eval.schema.json
@@ -161,7 +161,7 @@
                             },
                             "metric_id": {
                                 "type": "string",
-                                "description": "Canonical metric identifier for cross-source normalization (e.g. accuracy, f1_macro, auroc, pass_at_k)."
+                                "description": "Stable metric identifier for joining/deduping/querying. Use a canonical global id when applicable (e.g. accuracy, f1_macro, auroc, rmse, pass_at_k). For benchmark/leaderboard-specific metrics, use a namespaced id (e.g. rewardbench.overall, lmarena.elo)."
                             },
                             "metric_name": {
                                 "type": "string",
--- a/instance_level_eval.schema.json
+++ b/instance_level_eval.schema.json
@@ -29,11 +29,11 @@
         },
         "evaluation_name": {
             "type": "string",
-            "description": "The specific eval name, ideally unique (e.g. GSM8K, mmlu_physics). Primarily for display/filtering when evaluation_result_id is unavailable."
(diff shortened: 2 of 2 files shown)


borgr:
The matching keys make a lot of sense

"cross-source normalization" is that a known term?

There's quite a lot of fields in the metric, it is unclear to me which of those are necessary and how would I get about to fill them if I was contributing a new evaluation.
Why do we need a "kind" as opposed to name and id?
Also unit, is that something that we need to write ourselves? won't it always be deducible from the rest?
parameters and "more" sounds good

yananlong:
Thanks for the detailed feedback @nelaturuharsha @borgr @janbatzner @mrshu @dmjoy.

I pushed a couple updates based on this thread:
- PR description now includes: (1) a concrete `metric_id` convention (global ids when applicable, otherwise namespaced ids like `rewardbench.overall` / `lmarena.elo`), (2) a deterministic `evaluation_result_id` recipe, and (3) 2 before/after JSON examples.
- Schema text: `metric_id` description explicitly mentions namespacing (`eval.schema.json`), and instance-to-aggregate linkage semantics are clarified (`instance_level_eval.schema.json`) so one instance record links to one aggregate result; emit multiple instance rows when a sample contributes to multiple aggregate results.

If you want the `evaluation_result_id` recipe to *always* include/exclude `generation_config` (vs only when needed to disambiguate), I can tighten that wording.

dmjoy:
@yananlong I'm not about the approach to handling instances that are associated with more than one evaluation result (comment quote):
> if a single underlying interaction/sample contributes to multiple aggregate results, emit multiple instance records with different evaluation_result_id values.

At least for some of the HELM outputs & conversions I've played with, there can be 10 or so evaluation result records (for different metrics) across the same set of instances.  It seems excessive to just duplicate each instance 10 times with a different evaluation result ID.  I understand there are constraints here trying to fit everything into two JSON files, but in a relation database for example you might have a join table since this is a many-to-many relationship (between instances and evaluation results).  Maybe a compromise here would be to have the `evaluation_result_id` foreign key in the instances be a list instead that can reference all of the `EvaluationResults` that they're associated with.  (There are of course other ways to handle this but the list of `evaluation_result_id`s seems like the shortest path from where this PR is currently) 

EDIT: It may also make sense to have this relationship to be implicit if every instance is associated with every evaluation result (and it's only necessary to include the `evaluation_result_id` foreign key(s) in the results where this isn't true)

yananlong:
@dmjoy Great point, and I agree this is the main tradeoff.

I kept `evaluation_result_id` as a single FK because each instance row currently has a single `evaluation.score` / `evaluation.is_correct`. If we make `evaluation_result_id` a list, the row no longer says which score belongs to which aggregate metric result (unless we also redesign `evaluation` into a per-metric map/list).

So in this PR the model is intentionally: one instance-level row = one metric result for one sample, with explicit duplication when the same sample contributes to multiple aggregate results.

I agree this is not storage-optimal for HELM-like runs with many metrics. A cleaner many-to-many shape would be a follow-up normalization (e.g., stable `instance_id` + separate instance↔metric-result table/object), but that is a larger schema change than this PR.

I can open a follow-up issue/PR sketch for that normalized design right after this lands if that sounds good.

PR #53, opened 2026-02-23
fix: handle missing metadata in Inspect adapter

mrshu:
Previously Inspect conversion assumed metadata_args was always provided; this commit makes missing metadata safe and keeps outputs valid. It also avoids writing instance-level files when output metadata is absent.

- Normalize `metadata_args` to `{}` in Inspect adapter entry points
- Default source metadata to `unknown` org and `third_party` relationship
- Skip detailed sample file emission without `parent_eval_output_dir`
- Add regression test for `transform_from_file(..., metadata_args=None)`

### DIFF
--- a/eval_converters/inspect/adapter.py
+++ b/eval_converters/inspect/adapter.py
@@ -466,7 +466,7 @@ class InspectAIAdapter(BaseEvaluationAdapter):
 
         evaluation_name = eval_spec.dataset.name or eval_spec.task
 
-        parent_eval_output_dir = metadata_args.get('parent_eval_output_dir')
+        parent_eval_output_dir = metadata_args.get("parent_eval_output_dir", "data")
         if raw_eval_log.samples and parent_eval_output_dir:
             if "/" in model_info.id:
                 model_dev, model_name = model_info.id.split("/", 1)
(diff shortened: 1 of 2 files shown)


eval_converters/inspect/adapter.py
```diff
@@ -448,15 +463,21 @@ def _transform_single(
...
-        parent_eval_output_dir = metadata_args.get('parent_eval_output_dir')
-        evaluation_dir = f'{parent_eval_output_dir}/{source_data.dataset_name}/{model_dev}/{model_name}'
-        detailed_results_id = f'{metadata_args.get('file_uuid')}_samples'
-        
+
         evaluation_name = eval_spec.dataset.name or eval_spec.task
 
-        if raw_eval_log.samples:
+        parent_eval_output_dir = metadata_args.get('parent_eval_output_dir')
+        if raw_eval_log.samples and parent_eval_output_dir:
```
damian1996: Hmm, I think we still should save instance-level logs if `samples` are available and `parent_eval_output_dir` isn't provided. We can default for set `parent_eval_output_dir` as `"data"`.

mrshu: Thanks @damian1996, that makes a lot of sense -- let me know if this last commit does it in the way you'd expect.

PR #56, opened 2026-02-25
Update HELM leaderboards parser for schema v0.2.1

damian1996:


utils/helm/adapter.py
```diff
@@ -94,7 +94,9 @@ def extract_generation_config(run_specs: List[str]) -> Dict[str, Any]:
     # Collapse values if they are identical
     for key, values in list(generation_config.items()):
         if len(set(values)) == 1:
-            generation_config[key] = values[0]
+            values = values[0]
+
+        generation_config[key] = str(values)
```
mrshu: Just a quick note: converting lists via str(...) gives Python reprs (e.g. "['a', 'b']"), which can be fragile for downstream parsers. Would json.dumps(values) be safer here for a stable serialized format?

PR #57, opened 2026-03-02
Refactor code into every_eval_ever namespace and add modal CLI

Erotemic:
Here's an initial pass at a refactor to make this module easier to deploy to pypi and use as a standalone tool.

* All library code and json schemas are now under the `every_eval_ever` namespace. 
* I did keep symlinks to the json schemas as I know external links reference them. They can be removed in the future. 
* Added an CLI entrypoint in pyproject.toml so that `every_eval_ever` becomes a command line tool. 
* Added modal CLIs using basic argparse for conversions and schema validation.
* Made specific backend dependencies optional (the converter CLI will give you a helpful message if you don't have them installed and try to use them)
* The validate and duplicate checks are moved out of utils and into the main library, but the oneoff scripts I kept in utils. It probably would be best to rename that to "scripts", "dev", or "maintain" because right now it looks like its something that should belong in the library even though it is just repo maintenance.
* Removed a lot of unnecessary boilerplate from tests now that the library can just be imported by name.

I didn't do github actions stuff, but you can at least `uv pip install -e .` the repo now and get a nice CLI. 

On slack there was mention of github actions existing, but I didn't see a .github directory and lack of CI tests for the PR makes me nervous. Are there CI tests somewhere? I do see that actions are running in the actions tab, but I've never set them up without a .github folder. 

Note: the script in utils/helm/parse_helm_leaderboards.sh seems to have an error which I don't think is introduced by this PR. I've left it alone for now, but in general `uv run` commands would need to be updated with `uv run --extra <extra>` for whichever extra they needed.

### DIFF
--- a/tests/test_helm_instance_level_adapter.py
+++ b/tests/test_helm_instance_level_adapter.py
@@ -17,9 +17,9 @@ def _load_instance_level_data(adapter, filepath, metadata_args):
         output_path=str(Path(metadata_args['parent_eval_output_dir']) / 'helm_output'),
         metadata_args=metadata_args
     )
-    
+
     converted_eval = converted_eval_list[0]
-    
+
     instance_level_path = Path(converted_eval.detailed_evaluation_results.file_path)
     instance_logs = []
     with instance_level_path.open('r', encoding='utf-8') as f:
@@ -27,13 +27,13 @@ def _load_instance_level_data(adapter, filepath, metadata_args):
             if line.strip():
                 data = json.loads(line)
                 instance_logs.append(InstanceLevelEvaluationLog.model_validate(data))
-    
+
     return converted_eval, instance_logs
 
 
 def test_mmlu_instance_level():
     adapter = HELMAdapter()
-    
+
     with tempfile.TemporaryDirectory() as tmpdir:
         metadata_args = {
             'source_organization_name': 'TestOrg',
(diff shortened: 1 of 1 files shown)


Erotemic:
Also note the github diff reports that the schema files were deleted and re-added. I think this is a consequence of adding the symlinks to keep existing links alive as I did `git mv` them. Regardless, you can validate that there is no difference in those files with: 

```
# show original hashes from the main branch
sha256sum <(git show main:instance_level_eval.schema.json) <(git show main:eval.schema.json)
# show updated hashes on this branch 
sha256sum every_eval_ever/schemas/instance_level_eval.schema.json every_eval_ever/schemas/eval.schema.json
```

Which on my machine results in:


```
ab656c857373393c76eade428fa4b7b346438e90aaedafd6add9f1b0b6954639  /dev/fd/63
a8c534a87358c9f38e5a059c921f0c4ad51a6653f90c349e84e82c6a5bbc9719  /dev/fd/62
ab656c857373393c76eade428fa4b7b346438e90aaedafd6add9f1b0b6954639  every_eval_ever/schemas/instance_level_eval.schema.json
a8c534a87358c9f38e5a059c921f0c4ad51a6653f90c349e84e82c6a5bbc9719  every_eval_ever/schemas/eval.schema.json
```

PR #63, opened 2026-03-10
[Submission] Add Terminal-Bench 2.0 adapter script

StevenDillmann:
Adapter script to convert Terminal-Bench 2.0 leaderboard data (115 agent+model entries) into EEE schema-conformant JSON files. Data submitted to HF datastore via PR #28: 
https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/28


StevenDillmann:
@evijit TB doesn't currently have a public API to get the live leaderboard - I could do website scrapping but that may be less robust

PR #64, opened 2026-03-12
added testing dataset

joeda:
The evals are shortened to 5 samples each and are sourced from a full evaluation on Inspect AI

No idea if that many evals are actually useful since many might be redundant, but since the conversion of some of them trigger errors the current testing coverage seems to be insufficient

also not sure if just checking that an eval loads and is not empty is the best way


damian1996:
@joeda Thanks for the update on your changes! It looks solid now. Could you please review these Copilot mentions (or have you already done so?) and resolve any conflicts? Then we would like to merge it.

Erotemic:
An immediate hesitation I have is that checking data into a repo can cause a lot of bloat. That said, it is not without precedent. There are basically three ways to handle this:

* check all data into the main repo
* maintain a separate companion dataset repo or package
* use a hybrid strategy, with small canonical data in-tree and larger data fetched separately

Example projects that follow each pattern include:

* all data in the main repo: Fiona and Rasterio
* hybrid: scikit-learn, scikit-image, statsmodels
* separate companion dataset repo/package: GeoPandas / geodatasets, seaborn / seaborn-data, statsmodels / smdatasets

Personally, I would prefer a separate companion package for test data. I have run into this problem enough times that I think it is worth investing the effort to set it up cleanly at the start of the project.

PR #90, opened 2026-03-24
Fix different uuids for json and jsonl files for the same log in the CLI

damian1996:


every_eval_ever/cli.py
```diff
@@ -57,6 +57,23 @@ def _write_log(
     return out_file
 
 
+def _extract_file_uuid_from_detailed_results(log: Any) -> str | None:
```
nelaturuharsha: It might be better to use regex of the form:
```
import re
_UUID_FILE_RE = re.compile(
--
r'^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}'
r'(?:_samples)?\.jsonl?$',
re.IGNORECASE,
)
```

 rather than expect the stem maybe

PR #100, opened 2026-04-06
fix: correct save_to_file creating directory at file path

mrshu:
Previously save_to_file() called mkdir() on the target file path itself instead of its parent directory, causing an IsADirectoryError on every call. Also fix the return type annotation from bool to None since the function never returned a value.

- Change obj_path.mkdir() to obj_path.parent.mkdir()
- Fix return type annotation from bool to None
- Add tests covering nested paths, existing dirs, and exclude_none behavior

### DIFF
--- a/every_eval_ever/converters/inspect/utils.py
+++ b/every_eval_ever/converters/inspect/utils.py
@@ -334,16 +334,6 @@ def extract_model_info_from_model_path(model_path: str) -> ModelInfo:
     )
 
 
-def save_to_file(path: str, obj: BaseModel) -> bool:
-    json_str = obj.model_dump_json(indent=4, exclude_none=True)
-
-    obj_path = Path(path)
-    obj_path.mkdir(parents=True, exist_ok=True)
-
-    with open(obj_path, 'w') as json_file:
-        json_file.write(json_str)
-
-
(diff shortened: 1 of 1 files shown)


mrshu:
@nelaturuharsha @borgr I just noticed this while going through the code and thought it might make sense to fix.

Let me know if things should be done in some other way -- happy to update it! :)

damian1996 (approved):

damian1996:
Thanks for your contribution @mrshu . We can merge it now. However I've seen this function wasn't used in the repository. How did you manage to find out about this problem? I'm wondering if we need to keep it or remove it as stale code.

mrshu:
> However I've seen this function wasn't used in the repository

@damian1996 you are indeed correct, it's basically dead code -- it just jumped up on me :)

It would probably be best to remove it though.

damian1996:
Yeah, please update this PR to remove it instead of the proposed fix and we will merge it. :)

mrshu:
> Yeah, please update this PR to remove it instead of the proposed fix and we will merge it. :)

Will do -- thanks!

mrshu:
@damian1996 thanks again for the ping, should be done now :)

PR #102, opened 2026-04-06
fix: preserve falsy generation args in HELM adapter

mrshu:
Previously _extract_generation_args() used `or` to fall back to adapter defaults. This treated legitimate values like temperature=0 (deterministic sampling) as missing and silently replaced them with the adapter spec defaults. Converted HELM metadata could therefore lie about how a run was actually configured.

- Replace `or` with `if ... is not None` for temperature, max_tokens, top_p, and top_k
- Add tests verifying zero values are preserved and None falls back correctly


mrshu:
@nelaturuharsha @borgr same here, a small problem I noticed (happy to just close this)

damian1996 (approved):

PR #105, opened 2026-04-14
Tests for Inspect adapter edge cases (empty choices, local paths, misleading dataset names)

MattFisher:
## Summary

Three issues discovered while converting ~100 real-world Inspect evaluation logs (from [CVE-Bench](https://github.com/bento-cyber/cvebench) and the [inspect_evals dashboard](https://github.com/UKGovernmentBEIS/inspect_evals)):

### 1. Empty `output.choices` → IndexError
Some agentic samples (e.g. sandbox failures) have `output.choices == []`. The instance-level adapter unconditionally indexes `choices[0]`, raising `IndexError` (wrapped in `TransformationError`).

**Possible fix:** Guard the access:
```python
if sample.output.choices:
    message = sample.output.choices[0].message
    content = message.content
else:
    content = ''
```
Also needs a similar guard for `stop_reason` in the metadata dict.

### 2. Local filesystem paths in `hf_repo`
`_extract_source_data` always returns `SourceDataHf` with `hf_repo=dataset.location`. When the dataset comes from a local file (not HuggingFace), this produces paths like `/root/.cache/inspect_evals/...` or `/opt/venv/lib/.../challenges.json`. This is particularly an issue for cyber evals that don't use an external dataset.

### 3. Misleading `dataset_name`
`dataset_name` is derived from `dataset.name` which for many inspect_evals is the internal filename rather than the benchmark name:
- `cyse2_vulnerability_exploit` → `'challenges'`
- `gdm_intercode_ctf` → `'ic_ctf'`

## Tests
5 tests in `tests/test_inspect_edge_cases.py`, all marked `xfail(strict=True)` to document current behavior. Fixtures are stripped-down real logs (~10–27KB each, 1 sample, events removed).

## Fixtures
- `data_cvebench_empty_choices.json` — CVE-Bench sample with empty choices (sandbox failed to start)
- `data_intercode_ctf_local_path.json` — gdm_intercode_ctf with local cache path
- `data_cyse2_vuln_exploit_challenges.json` — cyberseceval_2 with misleading dataset name


damian1996:
Hi @MattFisher ! Thank you for your contribution :) Would you like to include fixes for these issues in the Inspect adapter? We could add these changes to this PR for simplicity.

- The fix for Issue 1 looks good to me. Regarding stop_reason, we currently use:
`str(sample.output.stop_reason) if sample.output.stop_reason else ''`
Do you think this is insufficiently guarded? Given that the documentation indicates `stop_reason` may depend on the choices list, we could change it to:
`str(sample.output.stop_reason) if sample.output.choices and sample.output.stop_reason else ''`
Alternatively, we could modify `stop_reason` to include stop reasons for all messages in the metadata, rather than just the first one. What are your thoughts?

- For the second issue, good catch! What do you think about using a heuristic to detect local paths and then crafting a `SourceDataPrivate` object? Does that seem like a robust enough solution?

- Regarding the last issue: if `dataset.name` is consistently misleading, do you see any other valid options for what we should use as the `dataset_name`?

MattFisher:
Thanks for the review @damian1996!

> Would you like to include fixes for these issues in the Inspect adapter? We could add these changes to this PR for simplicity.

I've added the fixes and updated the tests so they're not xfail.

> Given that the documentation indicates stop_reason may depend on the choices list, we could change it to:
`str(sample.output.stop_reason) if sample.output.choices and sample.output.stop_reason else ''`

This seems fine to me, it's what I went with.

> What do you think about using a heuristic to detect local paths and then crafting a SourceDataPrivate object?

I've added logic so that if the dataset location looks like a huggingface path ("{owner}/{dataset_name}") then to treat it as such, otherwise assume it's a local path and output a `SourceDataPrivate`.

> if dataset.name is consistently misleading, do you see any other valid options for what we should use as the dataset_name?

It wasn't consistently wrong, but it was unpredictable.

| Task | `dataset.name` | What it actually is |
|---|---|---|
| `inspect_evals/gaia` | `GAIA` | benchmark name (correct) |
| `inspect_evals/pubmedqa` | `bigbio/pubmed_qa` | HF repo |
| `inspect_evals/arc_easy` | `allenai/ai2_arc` | HF repo (ambiguous - same for `arc_challenge`) |
| `inspect_evals/humaneval` | `openai/openai_humaneval` | HF repo |
| `inspect_evals/cyse2_vulnerability_exploit` | `challenges` | internal filename (incorrect) |
| `inspect_evals/gdm_intercode_ctf` | `ic_ctf` | internal filename (incorrect) |

Using the same logic as above, the problems are when the dataset isn't a HF dataset, so in that case we fall back to the task name, like `gdm_intercode_ctf`.

I'e also updated the instance-level adapter to store the appropriate value in `evaluation_id` as identified in https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/57#69df8fadca94a2540cd9770a

every_eval_ever/converters/inspect/adapter.py
```diff
@@ -228,9 +231,75 @@ def _extract_evaluation_results(
...
         self, dataset: EvalDataset, task_name: str
-    ) -> SourceDataHf:
+    ) -> SourceDataHf | SourceDataPrivate:
+        sample_ids = (
+            [str(sid) for sid in dataset.sample_ids]
+            if dataset.sample_ids is not None
+            else None
+        )
+
+        if self._looks_like_local_path(dataset.location):
```
damian1996, line 264: What do you think about one of examples in the repo?

```
"dataset": {
      "name": "GAIA",
      "location": "inspect_evals/gaia_dataset/GAIA",
}
```
SourceDataPrivate and SourceDataHf aren't viable options in this case (for any inspect_evals) :c Do you think it's possible to check if it is a benchmark from inspect_evals and save it as SourceDataUrl?

MattFisher: Do you know where this log came from or what version of `inspect_ai` or `inspect_evals` it was generated from? I guess it doesn't really matter - we'd want to be able to handle logs of any vintage.

It won't be hard to identify this as an inspect_evals dataset; I'm just not sure what other formats this string can be in. I'll check the logs that I have access to and see what the distribution looks like.

damian1996: Fair enough 👍

damian1996 (approved):

PR #70, opened 2026-03-17
Draft Proposal: Agent Session Result Layer

elronbandel:
## Draft Proposal: Agent Session Result Layer

This PR proposes extending Every Eval Ever with session-level reporting for agentic evaluations. It is a starting point for discussion, not a finished specification.

Full context and motivation: **[What Agent Evaluation Teams Don’t Tell You](https://www.linkedin.com/posts/elron_aiagents-agentevaluation-llm-ugcPost-7439715804885159936-3nEw)**

### The gap

EEE already captures:
- aggregate score outcomes
- agentic setup (`agentic_eval_config`, `eval_limits`, `sandbox`)
- instance-level traces (`interaction_type`, `messages`, `tool_calls`)

What is missing is standardized session-level semantics: how the run ended, which side failed, how much interaction occurred, and what system was actually evaluated. These are part of the evaluation result, not just diagnostics.

### Proposed extensions (all optional under `evaluation_results[]`)

`session_result`
- `status`: `success | unsuccessful | unfinished | error | cancelled | limit_reached`
- `is_finished`: `boolean`
- `finish_accepted`: `boolean`
- `stop_reason`: `agent_done | timeout | max_steps | error | cancelled | benchmark_policy`
- `error_attribution`: `agent | benchmark | external | unknown`
- `error_detail`: `string`

`session_accounting`
- `step_count`, `action_count`, `invalid_action_count`, `parallel_action_max`: `integer`
- `time_to_first_action`, `wall_clock_seconds`, `agent_cost`, `benchmark_cost`: `number`

`agent_system` / `benchmark_system`
- White-box description of the evaluated agent (models, tools, subagents, memory) and benchmark-side runtime/grader/protocol

`eval_conditions`
- `internet_access`, `memory_exposure`, `reset_policy`, `permissions`, `repeated_runs`, `seed`

`robustness` (optional / emerging)
- `method`, `num_variants`, `variance_metric`, `variance_value`

### Backward compatibility

- No existing fields are removed or changed
- All new fields are optional
- Existing records remain valid
- Model-only evaluations are unchanged

### Suggested rollout

- Add fields as optional in schema
- Encourage early adoption of `session_result` and `session_accounting`
- Keep composition/conditions/robustness optional while conventions converge

### Discussion welcome

Feedback is especially welcome on:
- field naming
- schema placement/hierarchy
- priority order for adoption

[...]

PR #108, opened 2026-04-15
Add AIR-Bench leaderboard to HELM adapter

yifanmai:


tommasocerruti:
The current `convert()` logic is not compatible with AIR-Bench’s header format.

This PR adds `HELM_AIR_Bench` and routes it through the existing generic parsing path, but for AIR-Bench, the live HELM file selected by the adapter is `safety_scenarios.json`, and its headers all look like:

```text
AIRBench 2024 - Refusal Rate
AIRBench 2024 - Security Risks
AIRBench 2024 - Operational Misuses
...
```

The problem is that the current keying logic effectively does this:

```python
short_name = full_eval_name.split()[0] if '-' in full_eval_name else full_eval_name
```

i.e., because the header contains `-`, it returns the first whitespace splitted token, which for AIR-Bench is always `AIRBench`.

As a result:
- only the first AIR-Bench column creates an `EvaluationResult`
- the remaining 379 AIR-Bench scores are folded into `score_details.details`
- AIR-Bench is therefore stored as one giant aggregated result per model instead of many distinct results

So while this PR correctly wires AIR-Bench into the adapter entrypoints, it does not parse AIR-Bench correctly.

I checked the live HELM AIR-Bench source from my shell, and the actual headers are the following:

```bash
curl -sS https://storage.googleapis.com/crfm-helm-public/air-bench/benchmark_output/releases/v1.19.0/groups/safety_scenarios.json \
| jq -r '.[].header[1:][]?.value' \
| sed -n '1,20p'

AIRBench 2024 - Refusal Rate
AIRBench 2024 - Security Risks
AIRBench 2024 - Operational Misuses
AIRBench 2024 - Violence & Extremism
AIRBench 2024 - Hate/Toxicity
AIRBench 2024 - Sexual Content
AIRBench 2024 - Child Harm
AIRBench 2024 - Self-harm
AIRBench 2024 - Political Usage
AIRBench 2024 - Economic Harm
AIRBench 2024 - Deception
AIRBench 2024 - Manipulation
AIRBench 2024 - Defamation
AIRBench 2024 - Fundamental Rights
AIRBench 2024 - Discrimination/Bias
AIRBench 2024 - Privacy
AIRBench 2024 - Criminal Activities
AIRBench 2024 - #1: Confidentiality
AIRBench 2024 - #2: Integrity
AIRBench 2024 - #3: Availability
```

The current keying logic collapses those headers to the same key `AIRBench` 380 times:

```bash
curl -sS https://storage.googleapis.com/crfm-helm-public/air-bench/benchmark_output/releases/v1.19.0/groups/safety_scenarios.json \
| jq -r '.[].header[1:][]?.value' \
| awk '{print $1}' \
| sort | uniq -c

 380 AIRBench
```

The data uploaded in your HF PR appears to confirm exactly this behavior:

- HF PR: [evaleval/EEE_datastore discussion #70](https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/70)
- Sample raw file from that PR: [example JSON](https://huggingface.co/datasets/evaleval/EEE_datastore/resolve/49669cc33bef9536f6407e9cdb0e04cbdce186d6/data/helm_air_bench/01-ai/yi-34b-chat/6f282a52-0a35-47af-8c55-6777ed6711e4.json)

That file contains:
- only 1 entry in `evaluation_results`
- `evaluation_name = "AIRBench 2024"`
- hundreds of AIR-Bench category scores stored under `score_details.details`

(NIT: There is also a smaller metadata inconsistency in the PR title that says `v1.16.0`, but the sampled JSON points to `v1.19.0` in `source_data.url`).

This needs to be fixed here before merge, Codex suggested these as potential action items:

1. Handle AIR-Bench separately in `convert()`
   - AIR-Bench should not use the current `short_name = full_eval_name.split()[0]` logic.
   - The per-result key should come from something unique per column, likely the part after `' - '`.

2. Emit separate `evaluation_results` for AIR-Bench metrics
   - `Refusal Rate`, `Security Risks`, `Operational Misuses`, etc. should become distinct results, not entries inside one `details` blob.

3. Add a regression test
   - A minimal AIR-Bench fixture with two headers would be enough to catch this:
     - `AIRBench 2024 - Refusal Rate`
     - `AIRBench 2024 - Security Risks`
   - The test should assert that two `EvaluationResult`s are produced.

4. Regenerate the HF AIR-Bench data after the adapter fix
   - The currently uploaded JSONs in [HF PR #70](https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/70) already reflect the collapsed structure, so they would need to be regenerated after fixing the adapter.

utils/helm/adapter.py
```diff
@@ -198,23 +199,42 @@ def convert(
...
+                # The column headers look like "AIRBench 2024 - Security Risks".
+                # For for this example, the dataset name should be "AIRBench 2024 - Security Risks"
+                # and the metric name should be "Refusal Rate".
+                # This differs from other HELM tables, in which the column headers
+                # are in the format "dataset_name - metric_name" (e.g. "MMLU - EM")
+                # This boolean indicates whether the special handling is needed.
+                is_helm_air_bench_category_table = (
+                    leaderboard_name == "helm_air_bench"
+                    and tab_name.startswith("AIR")
+                    and tab_name.endswith("categories"
```
damian1996: Missing )

yifanmai: Fixed.

tommasocerruti:
Looks good to me, thanks!

PR #109, opened 2026-04-16
feat: add base script for displaying statistics and creating visuals

gbemike:
- reads from files in `viewer_parquets/` in hugging face repo
- creates in memory database instance with duckdb
- creates a table from all parquet files in `viewer_parquet/`
- prints statistics from created table

Future commits will include visualizations.

### DIFF
--- a/every_eval_ever/helpers/eee_stats.py
+++ b/every_eval_ever/helpers/eee_stats.py
@@ -12,6 +12,11 @@ from tqdm import tqdm
 SEP = '=' * 60
 SUB = '-' * 60
 
+REPO_ID = 'evaleval/EEE_datastore'
+FOLDER_PATH = 'viewer_parquets'
+
+HUGGING_FACE_DATASTORE = f'datasets/{REPO_ID}/{FOLDER_PATH}/**/*.parquet'
+
 
 def execute_query(con, sql):
     return con.execute(sql).fetchall()
@@ -179,10 +184,6 @@ def analyze_data(con, schema_table, instance_table) -> None:
 
 
 def main():
-    repo_id = 'evaleval/EEE_datastore'
-    folder_path = 'viewer_parquets'
-
-    HUGGING_FACE_DATASTORE = f'datasets/{repo_id}/{folder_path}/**/*.parquet'
     print(f'\n{SEP}')
     print('EVERY EVAL EVER STATS')
     print(f'DATASTORE: {HUGGING_FACE_DATASTORE}')
--- a/every_eval_ever/helpers/eee_stats.py
+++ b/every_eval_ever/helpers/eee_stats.py
@@ -36,7 +36,7 @@ def print_startup_summary(
     instance_urls: List[str],
 ) -> None:
     print(f'\n{SEP}')
-    print('  EVERY EVAL EVER STATS')
+    print('EVERY EVAL EVER STATS')
     print(SUB)
     print(f'datastore: {datastore}')
     print(f'tables:    {table_name}_schema, {table_name}_instances')
@@ -77,15 +77,15 @@ def build_instance_select_sql(available_columns: set[str]) -> str:
         scalar_column('sample_id', 'VARCHAR'),
         scalar_column('sample_hash', 'VARCHAR'),
         scalar_column('interaction_type', 'VARCHAR'),
-        json_column('input', 'input_json'),
-        json_column('output', 'output_json'),
-        json_column('messages', 'messages_json'),
-        json_column('answer_attribution', 'answer_attribution_json'),
-        json_column('evaluation', 'evaluation_json'),
-        json_column('token_usage', 'token_usage_json'),
-        json_column('performance', 'performance_json'),
+        json_column('input', 'input'),
+        json_column('output', 'output'),
+        json_column('messages', 'messages'),
+        json_column('answer_attribution', 'answer_attribution'),
+        json_column('evaluation', 'evaluation'),
+        json_column('token_usage', 'token_usage'),
+        json_column('performance', 'performance'),
         scalar_column('error', 'VARCHAR'),
-        json_column('metadata', 'metadata_json'),
+        json_column('metadata', 'metadata'),
         scalar_column('filename', 'VARCHAR'),
     ]
 
@@ -267,15 +267,15 @@ def main():
                     sample_id VARCHAR NOT NULL,
                     sample_hash VARCHAR,
                     interaction_type VARCHAR NOT NULL,
-                    input_json JSON NOT NULL,
-                    output_json JSON,
-                    messages_json JSON,
-                    answer_attribution_json JSON NOT NULL,
-                    evaluation_json JSON NOT NULL,
-                    token_usage_json JSON,
-                    performance_json JSON,
+                    input JSON NOT NULL,
+                    output JSON,
+                    messages JSON,
+                    answer_attribution JSON NOT NULL,
+                    evaluation JSON NOT NULL,
+                    token_usage JSON,
+                    performance JSON,
                     error VARCHAR,
-                    metadata_json JSON,
+                    metadata JSON,
                     filename VARCHAR NOT NULL
                 )
                 """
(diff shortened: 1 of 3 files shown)


nelaturuharsha:
@evijit can you please run co-pilot?

every_eval_ever/helpers/eee_stats.py
```diff
@@ -0,0 +1,315 @@
...
+from typing import List, Optional, Tuple
+
+import duckdb
+from huggingface_hub import HfFileSystem
+from tqdm import tqdm
+
+SEP = '=' * 60
+SUB = '-' * 60
+
+hffs = HfFileSystem()
```
evijit: Module-level side effect: `HfFileSystem()` runs on import, so any test or tooling that imports this module touches HF auth/network. Move this into `main()` (or lazy-init inside `read_data`).

every_eval_ever/helpers/eee_stats.py
```diff
@@ -0,0 +1,315 @@
...
+
+    section('overview')
+    print(f'  {"total files":<32} {n_schema_files + n_instance_files:>10,}')
+    print(f'  {"total rows":<32} {n_schema_rows + n_instance_rows:>10,}')
+    print(f'  {"schema files":<32} {n_schema_files:>10,}')
+    print(f'  {"instance files":<32} {n_instance_files:>10,}')
+    print(f'  {"schema rows":<32} {n_schema_rows:>10,}')
+    print(f'  {"instance rows":<32} {n_instance_rows:>10,}')
+
+    # scehma vs instance columns
```
evijit: Typo: `scehma` → `schema`.

every_eval_ever/helpers/eee_stats.py
```diff
@@ -0,0 +1,315 @@
...
+    cols = execute_query(
+        con,
+        f"""
+        SELECT column_name FROM information_schema.columns
+        WHERE table_name = '{schema_table}';
+    """,
+    )
+    section(f'columns  ({len(cols)} total)')
```
evijit: This header and the identical one at line 158 both print `columns  (N total)`, so you can't tell which table the column list belongs to. Suggest `section(f'{schema_table} columns  ({len(cols)} total)')` and the same pattern for the instance block.

every_eval_ever/helpers/eee_stats.py
```diff
@@ -0,0 +1,315 @@
...
+        FROM {schema_table}
+        WHERE eval_library IS NOT NULL
+        GROUP BY 1 ORDER BY 2 DESC;
+    """,
+    )
+    section('eval library usage  (schema level)')
+    for lib, n in lib_count:
+        print(f'  {str(lib):<42} {n:>8,}')
+
+    # scheme source metadata
```
evijit: Typo: `scheme` → `schema`.

every_eval_ever/helpers/eee_stats.py
```diff
@@ -0,0 +1,315 @@
...
+    else:
+        print('  no params_billions data found')
+
+    print(f'\n{SEP}\n')
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--table', default='eee', help='Table name for database'
```
evijit, line 193: **Blocking-ish: SQL injection surface.** `args.table` flows unvalidated into ~10 f-string-interpolated SQL statements (`CREATE OR REPLACE TABLE {table_name}_schema`, `SELECT ... FROM {schema_table}`, `WHERE table_name = '{schema_table}'`, etc.). In-memory DB limits the blast radius, but it's still worth rejecting bad input here:

```python
import re
if not re.fullmatch(r'[A-Za-z_][A-Za-z0-9_]*', args.table):
    parser.error('--table must be a valid SQL identifier')
```

every_eval_ever/helpers/eee_stats.py
```diff
@@ -0,0 +1,315 @@
...
+    )
+
+    args = parser.parse_args()
+    table_name = args.table
+
+    datastore = HUGGING_FACE_DATASTORE
+
+    with duckdb.connect(':memory:') as con:
+        con.execute('INSTALL httpfs; LOAD httpfs;')
```
evijit: `INSTALL httpfs` hits the network on every run; after the first invocation only `LOAD httpfs;` is needed. Minor, but worth simplifying unless you're specifically guarding against fresh environments.

every_eval_ever/helpers/eee_stats.py
```diff
@@ -0,0 +1,315 @@
...
+
+        schema_urls, instance_urls = data_urls
+        print_startup_summary(
+            datastore=datastore,
+            table_name=table_name,
+            schema_urls=schema_urls,
+            instance_urls=instance_urls,
+        )
+
+        if not schema_urls and not instance_urls:
```
evijit, line 215: Effectively unreachable: `read_data` returns `None` when `files` is empty (caught at line 227) and otherwise always returns a 2-tuple of lists built from the same non-empty `files`. Either make `read_data` return `([], [])` uniformly and drop the `None` path, or drop this second guard.

every_eval_ever/helpers/eee_stats.py
```diff
@@ -0,0 +1,315 @@
...
+                    evaluation JSON NOT NULL,
+                    token_usage JSON,
+                    performance JSON,
+                    error VARCHAR,
+                    metadata JSON,
+                    filename VARCHAR NOT NULL
+                )
+                """
+            )
+            for url in tqdm(
```
evijit, line 259: **N+1 loading.** For each URL you run a `DESCRIBE` then an `INSERT ... SELECT`. DuckDB's `read_parquet(urls, union_by_name=true, filename=true)` handles missing columns across a heterogeneous list in one call — you already rely on this a few lines up for schemas. If the strict `NOT NULL` typing on `{table}_instances` is the reason you can't use the fast path, a one-line comment explaining that would help future readers. Otherwise replace the loop with a single `INSERT ... SELECT ... FROM read_parquet(?, union_by_name=true, filename=true)` passing the full `instance_urls` list.

evijit (commented):
_Review from @claude (posted by @evijit — the Claude GitHub App isn't installed here, so I'm just tagging rather than impersonating.)_

## Overview
Adds `every_eval_ever/helpers/eee_stats.py`, a standalone script that reads `viewer_parquets/**/*.parquet` from the `evaleval/EEE_datastore` HF dataset into an in-memory DuckDB (`{table}_schema` + `{table}_instances`), then prints summary stats (row/file counts, columns, eval-library usage, source-type breakdown, parameter range). Adds `duckdb>=1.5.2` to core deps.

## Blocking
**`--table` is f-string-interpolated into ~10 SQL statements** with no validation. In-memory DB limits blast radius, but it's still worth rejecting bad input early — see inline on `main()`.

## Correctness / style
- Two typos (`scehma`, `scheme`) — inline.
- Duplicate `section(f'columns  ({len(cols)} total)')` for schema and instance blocks — reader can't tell which table the column list belongs to. Inline.
- `analyze_data` prints a hand-rolled banner instead of going through `section()` — inconsistent.
- Module-level `hffs = HfFileSystem()` runs at import time. Inline.
- No entry point in `[project.scripts]` and no README note on how to invoke it — users have to know to run `python -m every_eval_ever.helpers.eee_stats`.

## Performance
- **N+1 queries** when loading instances: per-file `DESCRIBE` + `INSERT ... SELECT`. DuckDB can union-by-name across a list in one `read_parquet(urls, union_by_name=true)` call (you already do this for schemas). Inline on the loop.
- `INSTALL httpfs` hits the network every run; `LOAD httpfs;` alone is usually enough after the first install. Inline.

## Tests
No tests added. `build_instance_select_sql` is a pure function over a `set[str]` — a tiny unit test would catch the common failure of dropping/renaming a required column.

## Summary
Core logic looks right and the output is readable. Main asks: validate `--table`, collapse the per-file DuckDB loop into a single `union_by_name` read, fix the typos and duplicate headers.

gbemike:
done @evijit

tommasocerruti:
LGTM

PR #114, opened 2026-04-23
Adding support for HAL

AsafYehudai:
Adding support for HAL, the Holistic Agent Leaderboard. Nine agentic benchmarks.

### DIFF
--- a/utils/hal/adapter.py
+++ b/utils/hal/adapter.py
@@ -12,7 +12,7 @@ Each benchmark page embeds the leaderboard as an HTML table. This adapter:
   3. Produces one EEE JSON file per (benchmark, agent, model) entry
 
 Output structure:
-    data/hal-{benchmark_slug}/{developer}/{model_slug}/{uuid}.json
+    data/{benchmark_output_name}/{developer}/{model_slug}/{uuid}.json
 
 Usage:
     cd every_eval_ever
@@ -250,20 +250,10 @@ MODEL_DEVELOPER_MAP: dict[str, str] = {
     # Mistral
     "mistral": "mistralai",
     "mixtral": "mistralai",
-    # Alibaba
-    "qwen": "alibaba",
+    # Qwen / Alibaba — use "qwen" to match all other adapters in this repo
+    "qwen": "qwen",
     # Microsoft
     "phi": "microsoft",
-}
-
-# Explicit full-name overrides for models that are ambiguous
-FULL_NAME_DEVELOPER_OVERRIDES: dict[str, str] = {
-    "o3": "openai",
-    "o4-mini": "openai",
-    "gpt-5": "openai",
-    "gpt-4.1": "openai",
-    "gpt-4o": "openai",
-    "gpt-4": "openai",
 }
 
 # Maps cleaned model name → canonical EEE model ID
@@ -339,10 +329,18 @@ def get_model_id(raw_model: str) -> str:
     """
     cleaned = _normalize_model_name(raw_model)
 
-    # Check overrides
-    for key, model_id in MODEL_ID_OVERRIDES.items():
-        if cleaned == key or cleaned.startswith(key):
+    # Check overrides — longest key first to avoid shorter keys shadowing more
+    # specific ones (e.g. "gpt-4.1" must not match "gpt-4.1-mini").
+    # A prefix match is only accepted when the remaining suffix is solely a
+    # known inference-effort token (low / medium / high).
+    _EFFORT_SUFFIX = re.compile(r'^(low|medium|high)$', re.IGNORECASE)
+    for key, model_id in sorted(MODEL_ID_OVERRIDES.items(), key=lambda kv: len(kv[0]), reverse=True):
+        if cleaned == key:
             return model_id
+        if cleaned.startswith(f"{key} "):
+            suffix = cleaned[len(key):].strip()
+            if _EFFORT_SUFFIX.fullmatch(suffix):
+                return model_id
 
     # Auto-derive: developer/slugified-name
     developer = get_developer_for_model(raw_model)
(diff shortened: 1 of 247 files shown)


utils/hal/adapter.py
```diff
@@ -0,0 +1,792 @@
...
+    Returns: (record_dict, developer_slug, model_slug)
+    """
+    model_id = get_model_id(row.model_raw)
+    developer = model_id.split("/")[0] if "/" in model_id else "unknown"
+    model_slug_clean = model_id.split("/", 1)[-1] if "/" in model_id else model_id
+
+    effort = _effort_level(row.model_raw)
+    agent_slug = slugify(row.agent_name)
+
+    eval_id = (
```
tommasocerruti, line 635: This generated `evaluation_id` is based on benchmark + normalized model id + agent scaffold + timestamp, but HAL has separate leaderboard rows for things like “High” vs non-High, or different dated versions of the same model. Because those variants can normalize to the same `model_id`, two different HAL rows can end up with the same `evaluation_id`, which I saw it happened in the HF PR you opened that has 70 duplicated `evaluation_id`s.

Could we include something that uniquely identifies the HAL row in this ID, such as a slug of the raw HAL model name or the effort/date variant? After that, the HF data should be regenerated to avoid thid data duplication issue.

utils/hal/adapter.py
```diff
@@ -0,0 +1,797 @@
...
+
+    try:
+        rows = parse_table(html, benchmark)
+    except ValueError as e:
+        print(f"  ERROR parsing table: {e}")
+        return 0, 1
+
+    print(f"  Found {len(rows)} leaderboard entries")
+
+    benchmark_out_root = out_root / benchmark.output_name
```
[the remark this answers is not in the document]
tommasocerruti: The docstring is now consistent with the code, but I think the directory issue itself is still worth fixing. Right now, HAL outputs go into general benchmark folders like `data/gaia/`, `data/usaco/`, etc. Since HAL is its own leaderboard source, I think these should probably be namespaced as HAL outputs, for example, `data/hal-gaia/` or `data/hal-usaco/`.

There is also the additional conflict of the datastore already having `data/swe-bench-verified-mini/`, but this adapter writes `data/swebench-verified-mini/`. That would split the same benchmark family across two different directory spellings.

Could we settle the output paths before regenerating the HF PR?

utils/hal/adapter.py
```diff
@@ -0,0 +1,792 @@
...
+        slug="assistantbench",
+        name="AssistantBench",
+        output_name="assistantbench",
+        category="Web Assistance",
+        dataset_url="https://assistantbench.github.io",
+        description=(
+            "AssistantBench evaluates AI agents on 214 realistic, time-consuming, "
+            "and automatically verifiable tasks based on real human needs."
+        ),
+        metric_description="Accuracy on 214 real-world web assistance tasks (0.0–1.0)",
```
tommasocerruti: Nit: this describes the full AssistantBench benchmark, but HAL’s AssistantBench leaderboard appears to evaluate a 33-task split ([source](https://hal.cs.princeton.edu/assistantbench)). AssistantBench consists of 214 tasks, but the leaderboard stats show 33 total tasks, and, since these generated rows are HAL leaderboard results, the metric description “Accuracy on 214 real-world web assistance tasks” is misleading.

Could we change this to something like “Accuracy on HAL’s 33-task AssistantBench evaluation split”, and optionally record the full benchmark size / evaluated task count in `source_data.additional_details`?

tommasocerruti (commented):
Thanks for the updates :) I left a few remaining comments on issues that affect the generated data. Once those are addressed, could you rerun the adapter and update the HF dataset PR so the data matches the latest fixes from the adapter?

utils/hal/adapter.py
AsafYehudai: Good points @tommasocerruti, more broadly, I think it relates to the best practices in EEE around:
1. How to track results for the same model, with different hyperparameters? tem. reasoning -yes/no, effort level, etc.
2. How to write results for composite benchmarks? a. falt as look like from HF 2. add "hal-" prefix c. add "data/hal/benchmark"?
3. RE AssistantBench , a. is the best way to note it somewhere, b. remove it. Write it as a separate benchmark related to point 2.

What do you think @nelaturuharsha @damian1996 @janbatzner?

tommasocerruti: @nelaturuharsha @damian1996 @janbatzner tagging you again in case you missed this.

1. I think we should keep the standard `model_info.id` where possible and make `evaluation_id` include enough HAL config info to distinguish metadata such as effort, date, etc... If we don't distinguish them, different HAL rows for the same base model would collapse into the same EEE record identity;
2. I think your second proposed solution (adding `hal-` prefix) is the cleanest and resolvs the issue of mixing HAL leaderboard results with other records for the same underlying benchmark;
3. I think keeping AssistantBench makes sense, as long as the metadata distinguishes the 33 tasks evaluated by HAL from the full 214-task benchmark;

damian1996: 1. Storing hyperparams in `additional_details` is probably the best option the current schema has to offer... Yeah, it's definitely need better handling in new versions. Regarding packing the hyperparams for model into`evaluation_id` @yananlong do you think is acceptable option for backend? From backend perspective you only want it to be unique, right?
2. for me `hal-` prefix is the best option as well
3. Yeah, the option with keeping AssistantBench seems be okay

AsafYehudai: 1. Not sure if it is common, but in HAL, we have the same model with different hyperparams. This means the results json will have two different sets of results. See the first two lines here: https://hal.cs.princeton.edu/gaia 
Is this supported?
2-3. V

damian1996: Initially we haven't supported it in `evaluation_id`. But we will definitely change `evaluation_id` format in the future versions so to do not harm a backend operations the best option will be to follow Tommaso idea:

> I think we should keep the standard model_info.id where possible and make evaluation_id include enough HAL config info to distinguish metadata such as effort, date, etc...

tommasocerruti:
Great job!

PR #116, opened 2026-04-24
Canonical identity and schema upgrade tooling

yananlong:
## Summary
- add canonical identity backfill and validation tooling
- add strict schema upgrade support for 0.2.2 records
- extend converter output validation and tests
- remove the tracked temp datastore gitlink from the fork branch

## Validation
- .venv/bin/python -m pytest


every_eval_ever/eval_types.py
```diff
@@ -489,7 +489,7 @@ class EvaluationLog(BaseModel):
     model_config = ConfigDict(
         extra='forbid',
     )
-    schema_version: str = Field(
+    schema_version: Literal['0.2.2'] = Field(
```
damian1996: Do you think is it good idea to hardcore this value? It would be ideal to change the version in only one place after release of a new version to make it as painless as possible. In this case it would be ideal to have these literals fetched when eval_types and instance_level_types are rebuilded from json schemas.

every_eval_ever/instance_level_types.py
```diff
@@ -154,7 +154,7 @@ class InstanceLevelEvaluationLog(BaseModel):
     model_config = ConfigDict(
         extra='forbid',
     )
-    schema_version: str = Field(
+    schema_version: Literal['instance_level_eval_0.2.2'] = Field(
```
damian1996: the same case as eval_types.py

every_eval_ever/upgrade_schema_version.py
```diff
@@ -0,0 +1,192 @@
...
+from __future__ import annotations
+
+import argparse
+import json
+from dataclasses import dataclass
+from pathlib import Path
+
+from every_eval_ever.validate import expand_paths
+
+AGGREGATE_SCHEMA_VERSION = '0.2.2'
```
damian1996: There should be some global state of these variables with versions (like in every_eval_ever/converters/__init__.py) to avoid hardcoding them in all places.

I'm flagging it because this script could be somewhat useful in fixing data in the future.

every_eval_ever/instance_level_types.py
evijit: Codex: making this field a strict literal exposes a missed producer outside `every_eval_ever/converters`. The full suite now fails in `tests/test_openeval_adapter.py::test_include_instances_writes_valid_jsonl_sidecar` and `test_non_binary_instance_scores_do_not_claim_correctness` because `utils/openeval/adapter.py::make_instance_log` still passes the aggregate `SCHEMA_VERSION` value (`0.2.2`) into `InstanceLevelEvaluationLog`. Please update that adapter to use `instance_level_eval_0.2.2` as well, otherwise any OpenEval instance export path is broken by this change.

every_eval_ever/augment_canonical_identity.py
```diff
@@ -0,0 +1,1249 @@
...
+        counts[base] = counts.get(base, 0) + 1
+        evaluation_result_id = base
+        if counts[base] > 1:
+            evaluation_result_id = f'{base}__{counts[base]}'
+
+        if result.get('evaluation_result_id') != evaluation_result_id:
+            result['evaluation_result_id'] = evaluation_result_id
+            changed = True
+
+        sample_updates[raw_evaluation_name] = {
```
evijit: Codex: this mapping is keyed only by raw `evaluation_name`, so files with multiple aggregate results for the same eval slice but different metrics overwrite earlier entries. For example, HELM-style results `DatasetA` with `Acc on DatasetA` and `F1 on DatasetA` produce two distinct `evaluation_result_id`s, but `sample_updates['DatasetA']` keeps only the last one, and every sample row gets linked to that one metric. That silently corrupts the aggregate-to-sample foreign key for multi-metric evals. The updater needs a disambiguator, or it should duplicate/emit one instance row per aggregate result when a sample contributes to multiple metrics.

every_eval_ever/cli.py
```diff
@@ -412,6 +489,44 @@ def main(argv: list[str] | None = None) -> int:
...
+        if args.fail_on_issues:
+            forwarded_args.append('--fail-on-issues')
+        return check_canonical_identity_main(forwarded_args)
+
+    if args.command == 'augment-canonical-identity':
+        from every_eval_ever.augment_canonical_identity import (
+            main as augment_canonical_identity_main,
+        )
+
+        forwarded_args = [*args.paths]
```
evijit: Codex: `augment_canonical_identity.build_parser()` added `--skip-path-normalization`, but the top-level `every_eval_ever` CLI does not expose or forward it here. As a result `python -m every_eval_ever augment-canonical-identity --skip-path-normalization ...` exits with `unrecognized arguments`, even though the underlying command supports the option. Please add the flag to this parser and forward it alongside `--write` / `--skip-samples`.

evijit (commented):
Codex review: I left a few inline comments. I ran the new focused tests successfully, then ran the full suite and found one regression tied to the instance schema-version strictness change.

every_eval_ever/instance_level_types.py
yananlong: We shouldn't hard code the version, right?

every_eval_ever/augment_canonical_identity.py
```diff
@@ -0,0 +1,1249 @@
...
+        return _generic_inspect_eval_patch(
+            benchmark_family=benchmark_family,
+            raw_evaluation_name=raw_evaluation_name,
+            metric_config=metric_config,
+        )
+
+    if benchmark_family == 'theory_of_mind':
+        return _theory_of_mind_patch(raw_evaluation_name, metric_config)
+
+    return None
```
mrshu: This leaves the repo's checked-in HAL data uncovered by the backfill. On a copy of the current PR checkout, `augment-canonical-identity data/ --write` updated all 246 HAL aggregate files, but `check-canonical-identity data/ --fail-on-issues` still failed with 342 rows missing `metric_id`, `metric_name`, `metric_kind`, and `metric_unit`. The HAL adapter emits single-score accuracy-style metrics in `utils/hal/adapter.py`, so either the backfill needs a `hal-*` rule or the HAL producer should be updated; otherwise the new audit cannot pass on the repo's own `data/` directory.

every_eval_ever/upgrade_schema_version.py
```diff
@@ -0,0 +1,192 @@
...
+        action='store_true',
+        help='Write changes in place. Without this flag, run as a dry run.',
+    )
+    return parser
+
+
+def main(argv: list[str] | None = None) -> int:
+    args = build_parser().parse_args(argv)
+    file_paths = expand_paths(args.paths)
+    reports = [
```
mrshu: This command assumes every path returned by `expand_paths()` is an EEE datastore file. That is true for the checked-in `data/` directory, but not for common repo-scoped invocations: `upgrade-schema-version .` walks schema JSON and fixture JSON under `tests/data`, then crashes on list-shaped HELM fixture JSON; with `--write`, dict-shaped non-aggregate JSON can also be rewritten with a datastore schema version. Please filter JSON files by aggregate shape, or skip/report non-aggregate JSON instead of passing every `.json` to `upgrade_aggregate_file`.

every_eval_ever/augment_canonical_identity.py
```diff
@@ -0,0 +1,1249 @@
...
+    path_changed = target_file_path != file_path
+
+    reported_file_path = target_file_path if path_changed else file_path
+    reported_sample_file_path = (
+        target_sample_file_path if sample_exists else None
+    )
+
+    if write_changes and path_changed:
+        if sample_exists and target_sample_file_path is not None:
+            _move_file(sample_file_path, target_sample_file_path)
```
mrshu: Moving the sample sidecar before the aggregate can leave the datastore split if the aggregate move fails. I reproduced this with an over-nested aggregate plus sidecar where the normalized aggregate target already existed: the sidecar was moved to the normalized directory, then `_move_file(file_path, target_file_path)` raised `FileExistsError`, leaving the aggregate in the old path and its samples in the new path. Preflight both target paths before moving either file, or move through a rollback-safe sequence.

every_eval_ever/upgrade_schema_version.py
nelaturuharsha: there is a global state of this I think, was it over-written?

PR #117, opened 2026-04-24
Helm stress test fixes

Erotemic:
I'm converting all of the existing HELM results, and I ran into several errors. 

Summary of issues / fixes:

(1) Some scenarios have no reference answers, so correct_refs[0] could raise IndexError; the converter now uses an empty fallback when no references are present.

(2) Some completions contain thinking objects with text=None; since reasoning_trace is typed as list[str], the converter now drops None entries and returns None if no reasoning text remains.

(3) HELM multiple_choice_separate_* runs can set max_tokens=0 because they score by log-prob rather than generation; since EEE requires max_tokens >= 1, the converter now maps 0 to None to represent “not applicable.”

This PR is based on an older one, so the diff is muddled. It can likely be separated out as needed. I don't have tests in yet.

Benchmarks that caused the issues (has my root disk prefix, but they can be mapped to public URLs): 

Bug 1 — IndexError (correct_refs empty)
Benchmark: ifeval and wildbench, capabilities suite v1.12.0. Minimal reproducer:

/data/crfm-helm-public/capabilities/benchmark_output/runs/v1.12.0/ifeval:model=amazon_nova-premier-v1:0
Bug 2 — ValidationError (reasoning_trace None entry)
Benchmark: gpqa with a chain-of-thought qwen3 model, capabilities v1.12.0:

/data/crfm-helm-public/capabilities/benchmark_output/runs/v1.12.0/gpqa:subset=gpqa_main,use_chain_of_thought=true,use_few_shot=false,model=qwen_qwen3-235b-a22b-instruct-2507-fp8
Bug 3 — ValidationError (max_tokens=0)
Benchmark: bbq and blimp with multiple_choice_separate_* methods, classic v0.2.2:

/data/crfm-helm-public/classic/benchmark_output/runs/v0.2.2/bbq:subject=all,method=multiple_choice_separate_calibrated,model=anthropic_stanford-online-all-v4-s3,groups=ablation_multiple_choice

### DIFF
--- a/every_eval_ever/converters/helm/adapter.py
+++ b/every_eval_ever/converters/helm/adapter.py
@@ -7,6 +7,7 @@ from typing import Any, Dict, List, Tuple
 
 _HELM_IMPORT_ERROR: Exception | None = None
 try:
+    from dacite import Config as DaciteConfig
     from dacite import from_dict
     from helm.benchmark.adaptation.scenario_state import (
         AdapterSpec,
@@ -28,7 +29,7 @@ except (
     Exception
 ) as ex:  # pragma: no cover - exercised only when optional deps missing
     _HELM_IMPORT_ERROR = ex
-    from_dict = None  # type: ignore[assignment]
+    DaciteConfig = from_dict = None  # type: ignore[assignment]
     PerInstanceStats = AdapterSpec = RequestState = ScenarioState = Stat = (
         RunSpec
     ) = Any  # type: ignore[assignment]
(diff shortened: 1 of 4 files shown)


Erotemic:
Two more issues popped up (and I'm guessing there will be more). Bug 5 is left unfixed for now. Building the HELM objects requires constructing a MediaObject and that might not be necessary to do the actual conversion, so the change for bug 5 is just a better error message. 

## Bug 4 — `WrongTypeError` from `instance.id` type drift across HELM versions

**Triggering benchmark:** `infinite_bench_en_mc` in HELM Long-Context v1.0.0

**Root cause:** newer HELM runs can serialize `instance.id` as a JSON integer, while the HELM dataclass expects `Optional[str]`. This caused `dacite.from_dict` to fail during `ScenarioState` deserialization even though the underlying identifier is valid.

**Fix:** deserialize `ScenarioState` with `DaciteConfig(cast=[str])` so integer ids are coerced to strings during load.

**Rational:** `instance.id` is an identifier, not a numeric quantity. Converting `295` to `"295"` preserves the intended value and restores compatibility across HELM result versions rather than failing on valid official runs.

---

## Bug 5 — opaque `AssertionError` when local media assets are missing

**Triggering benchmark:** speech/audio runs such as `air_bench_chat` in HELM Speech v1.0.0

**Root cause:** HELM `MediaObject.__post_init__` asserts that referenced local media files exist. Some public speech runs include media paths in `scenario_state.json`, but the actual media assets are not present in the standard run JSON download. This caused deserialization to fail with a bare `AssertionError`, which was not actionable.

**Fix:** wrap `ScenarioState` deserialization and re-raise this case as a `FileNotFoundError` with a clear message explaining that the benchmark media assets must be downloaded alongside the run directory.

**Rational:** this does not hide a real data problem or misclassify the run as successfully converted. It preserves the failure, but turns an opaque internal assertion into a diagnosable and actionable error message so these runs can be triaged correctly during corpus sweeps.

damian1996 (approved):

PR #121, opened 2026-04-27
feat: add visualisation function

gbemike:
### Summary
This PR introduces a new function that creates the following visualisations and saves them to a path:
- Score distributions for key benchmarks (GPQA, IFEval)
- Inference platform usage
- Organization run counts
- Top benchmarks by number of models
- Top models by number of benchmarks
- All plots are saved to a configurable output directory.

### Usage
```bash
python every_eval_ever/helpers/eee_stats.py --table eee --viz-dir data/outputs/viz --top-n 10
```


every_eval_ever/helpers/eee_stats.py
```diff
@@ -225,6 +552,7 @@ def main():
                 [schema_urls],
             )
             schema_loaded = True
+            print(f'Schema Table Successfully Loaded')
```
mrshu: @gbemike this will likely have `ruff` scream that it uses f-strings without placeholders :)

every_eval_ever/helpers/eee_stats.py
```diff
@@ -275,12 +603,14 @@ def main():
                     [url],
                 )
             instance_loaded = True
+            print(f'Instance Table Successfully Loaded')
```
mrshu: @gbemike same as above :)

.gitignore
```diff
@@ -213,3 +213,4 @@ __marimo__/
 *.DS_Store*
 None/
 global-mmlu-lite/
+data/
```
mrshu: Note that this will match all directories named `data`, such as for instance `tests/data` and so on. If you only want to ignore the data folder on root (`/`), it would be best to say `/data/` here :)

gbemike: thanks

gbemike:
done @mrshu

every_eval_ever/helpers/eee_stats.py
```diff
@@ -180,9 +189,312 @@ def analyze_data(con, schema_table, instance_table) -> None:
...
+        y='benchmark',
+        palette='viridis',
+        ax=ax,
+    )
+    for container in ax.containers:
+        ax.bar_label(container, fmt='%d', padding=4, fontsize=9)
+
+    ax.set_xlabel('Number of Models', fontsize=11)
+    ax.set_ylabel('Benchmark', fontsize=11)
+    ax.set_title(
```
tommasocerruti: nit: could we build this title and filename (line 448) from `top_n`, e.g. `f'Top {top_n} Benchmarks by Number of Models'` and `f'top{top_n}_benchmarks_by_models.png'`? That way the plot stays accurate when someone runs the CLI with `--top-n 3` or another value, instead of hardcoding `Top 10`.

every_eval_ever/helpers/eee_stats.py
```diff
@@ -180,9 +189,312 @@ def analyze_data(con, schema_table, instance_table) -> None:
...
+def create_visualisations(con, schema_table, instance_table) -> None:
+    # for a benchmark, how do scores for that benchmark vary across
+    # eval_harnesses for the same evaluation / in general (get mean, range, get min, max)
+    try:
+        import matplotlib.pyplot as plt
+        import seaborn as sns
+    except ModuleNotFoundError:
+        raise ImportError('seaborn or matplotlib not installed')
+
+    os.makedirs(create_visualisations.outdir, exist_ok=True)
```
tommasocerruti, line 357: Is this file meant to be run only as a CLI script? If yes, ignore this comment. If not, we should pass `viz_dir` and `top_n` directly into `create_visualisations()`, because the function currently depends on `main()` setting those values first and would crash if someone imports and calls it directly.

tommasocerruti (commented):
Overall this looks good, I left two small comments to address before the merge. Thanks!

tommasocerruti:
Great job! @gbemike

PR #127, opened 2026-04-30
Update HELM adapter to use schema v0.2.2

yifanmai:


tommasocerruti:
The data pushed to the datastore looks correct, so the adapter appears to be working well. Thanks!

PR #133, opened 2026-05-08
[DRAFT] Transparent Compression

Erotemic:
This PR adds transparent compression support for aggregate and sample result files. Compressed variants of `<uuid>.json` and `<uuid>_samples.jsonl` are now treated as equivalents of the plain files for supported codecs: `.gz`, `.zst`, `.bz2`, `.xz`, and `.lz4`.

Changes:
- Add `every_eval_ever/io.py` with compression suffix detection, transparent openers, result discovery, and duplicate-variant resolution.
- Route validation and duplicate-entry checks through the new IO helpers.
- Add a `duplicate_variant` validation error when both plain and compressed variants of the same logical file are present.
- Add writer-side `--compress`, `--compress-aggregate`, and `--compress-samples` flags.
- Compute recorded checksums over the actual on-disk bytes.
- Document compression behavior and add optional `[zst]` and `[lz4]` codec extras.

Tests:
- Validate compressed files through `expand_paths` and `validate_file`.
- Ensure compressed files with invalid payloads still fail validation.
- Cover duplicate compressed/plain variants as CI-failing errors.
- Confirm aggregate and sample files do not false-positive as duplicates.
- Ensure missing optional codecs produce typed errors instead of crashes.
- Verify writer-side compression flags and override behavior.

Defaults remain unchanged: outputs are still uncompressed unless explicitly requested.

Out of scope:
The manifest-generation and viewer-Parquet scripts also need this resolver, but they live in `EEE_datastore`. This PR keeps the package-level compression support isolated and leaves datastore integration for a companion PR.

NOTE:

I would not be comfortable merging this until we have CI running. I have not tested this robustly yet.

### DIFF
--- a/every_eval_ever/check_duplicate_entries.py
+++ b/every_eval_ever/check_duplicate_entries.py
@@ -2,22 +2,36 @@ import argparse
 import hashlib
 import json
 import os
+from pathlib import Path
 from typing import Any, Dict, List
+
+from every_eval_ever import io as eee_io
 
 IGNORE_KEYS = {'retrieved_timestamp', 'evaluation_id'}
 
 
 def expand_paths(paths: List[str]) -> List[str]:
-    """Expand folders to file paths."""
+    """Expand folders to aggregate-EEE file paths.
+
+    Recognizes both plain ``<uuid>.json`` and compressed forms
+    (``.gz``, ``.zst``, ``.bz2``, ``.xz``, ``.lz4``).
+    ``<uuid>_samples.jsonl`` files are intentionally excluded — duplicate
+    detection runs over aggregate metadata, not per-instance samples.
+    """
     file_paths: List[str] = []
     for path in paths:
-        if os.path.isfile(path) and path.endswith('.json'):
-            file_paths.append(path)
-        elif os.path.isdir(path):
-            for root, _, file_names in os.walk(path):
-                for file_name in file_names:
-                    if file_name.endswith('.json'):
-                        file_paths.append(os.path.join(root, file_name))
+        p = Path(path)
+        if p.is_file():
+            if eee_io.is_eee_result(p) == 'aggregate':
+                file_paths.append(str(p))
+            else:
+                # Preserve historical behaviour: be explicit when a
+                # passed file is not an aggregate JSON.
+                continue
+        elif p.is_dir():
+            for found in eee_io.iter_eee_results([p]):
+                if eee_io.is_eee_result(found) == 'aggregate':
+                    file_paths.append(str(found))
         else:
             raise Exception(f'Could not find file or directory at path: {path}')
     return file_paths
@@ -84,7 +98,7 @@ def main(argv: List[str] | None = None) -> int:
     groups: Dict[str, List[Dict[str, Any]]] = {}
     for file_path in file_paths:
         try:
-            with open(file_path, 'r') as f:
+            with eee_io.open_eee_text(file_path, 'r') as f:
                 payload = json.load(f)
         except json.JSONDecodeError as e:
             message = f'JSONDecodeError: {str(e)}'
--- a/every_eval_ever/io.py
+++ b/every_eval_ever/io.py
@@ -79,16 +79,17 @@ _SUFFIX_BY_COMPRESSION: dict[str, str] = {
     name: suf for suf, name in _COMPRESSION_BY_SUFFIX.items()
 }
 
-# Canonical kind -> base suffix (the part *before* any compression suffix).
-_BASE_SUFFIX_BY_KIND: dict[str, str] = {
-    'aggregate': '.json',
-    'samples': '_samples.jsonl',
+# Canonical filename suffixes for each kind. The kind detector matches on
+# the *first* listed suffix that ends the filename (after stripping any
+# compression suffix). The ordering matters: ``.jsonl`` must be checked
+# before ``.json`` because ``.jsonl`` does not end in ``.json`` — but for
+# kinds that share a final extension (none currently do) this would matter.
+_KIND_SUFFIXES: dict[str, tuple[str, ...]] = {
+    'samples': ('_samples.jsonl', '.jsonl'),
+    'aggregate': ('.json',),
 }
 
-# Order matters: 'samples' must be checked before 'aggregate' because
-# ``_samples.jsonl`` ends in ``.jsonl`` not ``.json`` (different base suffix
-# overall) but the kind detector should not be confused by future shape
-# changes.
+# Order in which kinds are tried by the detector.
 _KIND_ORDER: tuple[str, ...] = ('samples', 'aggregate')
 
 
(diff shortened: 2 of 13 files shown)


Erotemic:
GPT Review (5m 12s):

## Review: **Request changes**

### High — Inspect sample outputs ignore the compression flags

The CLI advertises `--compress` as controlling both aggregate and per-instance outputs and exposes `--compress-samples` for every converter. However, `_cmd_convert_inspect` applies compression only when writing the aggregate, after the Inspect adapter has already generated its sample file. The Inspect instance writer still constructs a plain `.jsonl` path and uses `Path.open()` directly.

Consequently, `convert inspect --compress gz` produces a compressed aggregate but an uncompressed sample file, while `--compress-samples` is silently ignored. Thread the samples codec through `metadata` into `InspectInstanceLevelDataAdapter`, use `add_compression_suffix()` and `open_eee_text()`, and add an Inspect end-to-end compression test.

### High — Corrupt compressed streams can abort the entire validator

Aggregate validation catches only `OSError` and `ImportError` around decompression. Errors from some compression decoders are not necessarily `OSError`, so malformed compressed content can escape instead of becoming a `ValidationReport`.

Instance validation is more broadly exposed: its `try` block covers only `open_eee_text()`. Actual decompression happens while iterating later, outside the exception handler. A corrupt sample stream can therefore terminate the complete list comprehension in `main()`, preventing reports for every remaining file.

Wrap the actual reads and iteration in the error boundary, normalize decompression failures into a typed validation error, and test deliberately corrupt bytes—not merely valid compressed streams containing invalid JSON—for every supported codec.

### Medium — `check-duplicates` silently succeeds when an explicit input is unsupported

When an existing file is not recognized as an aggregate result, `expand_paths()` silently skips it. If every supplied path is skipped, the command checks zero files, prints “No duplicates found,” and exits successfully. That can green-light a mistyped filename or a CI invocation pointed at the wrong artifact.

Reject explicit non-aggregate files with a nonzero result rather than continuing silently, and add a regression test for an existing `.txt`, `.jsonl`, or otherwise unsupported path.

### Low — Sample duplicate reports use an inconsistent `file_type`

Normal sample validation reports use `file_type='instance'`, which is also the documented dataclass contract. Duplicate detection instead copies the IO kind directly, producing `file_type='samples'`. JSON consumers therefore receive two names for the same logical file type.

Map the IO kind to the validation-report vocabulary when constructing synthetic duplicate reports.

## Verification status

No GitHub Actions runs are attached to either the current head or the synthetic merge commit, so the reported local test result has not been independently exercised against the merge result.

every_eval_ever/io.py
```diff
@@ -0,0 +1,306 @@
...
+    if cs == COMPRESSION_XZ:
+        import lzma
+        return lzma.open(p, text_mode, encoding='utf-8')
+    if cs == COMPRESSION_ZST:
+        try:
+            import zstandard  # type: ignore[import-not-found]
+        except ImportError as exc:  # pragma: no cover - exercised in env w/o dep
+            raise ImportError(_missing_codec_msg('zst', 'zstandard', 'zst')) from exc
+        if text_mode == 'rt':
+            reader = zstandard.ZstdDecompressor().stream_reader(open(p, 'rb'))
```
mrshu, line 223: **[high] Zstandard reading** — The shared reader stops after the first frame

`stream_reader()` omits `read_across_frames=True`. Valid concatenated Zstandard frames can contain later JSONL records, so validation can finish without seeing them.

The shared reader can include every frame. A two-frame test can assert the later row and complete line count.

every_eval_ever/validate.py
```diff
@@ -375,7 +431,12 @@ def main(argv: list[str] | None = None) -> int:
         print('No files found to validate.', file=sys.stderr)
         return 1
 
-    reports = [
+    # Detect duplicate variants up front. Producing a fail-closed
+    # report per collision means the CI/PR-bot path cannot silently
+    # green-light a folder containing both `abc.json` and `abc.json.gz`.
+    duplicate_reports = _duplicate_variant_reports(file_paths)
```
mrshu, line 437: **[medium] duplicate variants** — Explicit-file validation cannot see compressed siblings

`expand_paths()` returns only the named file, and this call checks that limited list. `abc.json.gz` can therefore escape the one-variant rule when the caller names only `abc.json`.

Collision discovery can inspect same-parent, same-stem variants while payload validation stays limited to requested files.

every_eval_ever/check_duplicate_entries.py
```diff
@@ -84,7 +98,7 @@ def main(argv: List[str] | None = None) -> int:
     groups: Dict[str, List[Dict[str, Any]]] = {}
     for file_path in file_paths:
         try:
-            with open(file_path, 'r') as f:
+            with eee_io.open_eee_text(file_path, 'r') as f:
```
mrshu, line 101: **[medium] duplicate CLI errors** — Expected compressed-read failures bypass the file annotation

This handler catches only `JSONDecodeError`. A missing codec raises `ImportError`, and corrupt gzip data raises `BadGzipFile`, so the command exits early with a traceback.

Known codec and read failures can use the existing file-specific error path. Other programming errors can remain visible.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This is a comment-only review, and one confirmed high-priority issue remains.
**Result:** The compression design is broad and well tested. A few focused edge cases would make it safer.
**Scope:** This PR adds transparent compression across CLI conversion, I/O, validation, adapters, tests, and user documentation.

**Checks:** 3 concerns checked; 2 confirmed and 1 narrowed.
**Second check:** 2 reviewers checked 4 findings; 2 kept, 2 clarified, and none removed.

## Earlier review comments

The earlier review remains relevant at this head:

- **[high] Inspect output** — Inspect sample writers still ignore `--compress` and `--compress-samples`. [Earlier comment](https://github.com/evaleval/every_eval_ever/pull/133#issuecomment-5121789012)
- **[high] validation** — Corrupt compressed streams can still escape the validator error boundary. [Earlier comment](https://github.com/evaleval/every_eval_ever/pull/133#issuecomment-5121789012)
- **[medium] duplicate CLI** — An explicit unsupported file can still produce a successful zero-file check. [Earlier comment](https://github.com/evaleval/every_eval_ever/pull/133#issuecomment-5121789012)
- **[low] report schema** — Sample duplicate reports still use `samples` instead of `instance`. [Earlier comment](https://github.com/evaleval/every_eval_ever/pull/133#issuecomment-5121789012)

## What I noticed

|---|---|---|---|---|
| RAVF001 | high | Zstandard reading | `every_eval_ever/io.py:223` | The shared reader stops after the first frame in a valid multi-frame Zstandard stream. Later JSONL records can escape validation. |
| RAVF002 | medium | duplicate variants | `every_eval_ever/validate.py:437` | Explicit-file validation checks only the named path. A compressed sibling can therefore escape the one-variant rule. |
| RAVF004 | medium | duplicate CLI errors | `every_eval_ever/check_duplicate_entries.py:101` | Expected codec and decompression failures bypass the file-specific error path. The command exits early with a traceback. |

<details>

- **[low] UUID parsing** `every_eval_ever/converters/common/utils.py:9` — The shared UUID parser does not accept compressed result paths. No in-repository production caller was found, so this looks like a compatibility suggestion. (`RAVF003`)

</details>

## Things to try

- **[high] Zstandard reading** — The shared reader can read every concatenated frame. A two-frame JSONL test can assert the later row and full line count. (`RAVF001`)
- **[medium] duplicate variants** — Collision discovery can inspect only same-parent, same-stem variants. Payload validation can stay limited to requested files. (`RAVF002`)
- **[medium] duplicate CLI errors** — Known codec and read failures can use the existing file annotation path. Other programming errors can remain visible. (`RAVF004`)
- **[low] UUID parsing** — One central compression suffix can be removed before UUID matching. Parameterized utility tests can cover every supported codec. (`RAVF003`)

<details>

- Target: PR #133 (`dev/compression`, 13 files, +1270/-48)
- Mix: 3 codex-exec
- Earlier review comments: 4 still present
- Finding counts: 0 critical, 1 high, 2 medium, 1 low, 0 nit
- Checks: concerns=3; confirmed=2, ruled-out=0, set-aside=0, lowered=0, narrowed=1
- Second check: targeted; reviewers=2; kept=2, clarified=2, set-aside=0, removed=0; approval changed=no

</details>

---

evijit:
@Erotemic -- did you want to resume work on this? The datastore is increasing in size so this is a great idea to finish

Erotemic:
I can probably (90% chance) find time in the next week or two make this unstale. I'll comment if I start working on it. If anyone else wants to bring this to the finish line before that, then feel free.

PR #137, opened 2026-05-18
[Feature] Add text-to-image modality support

felifri:
## Summary

Three minimal, additive changes let the schema represent text-to-image generation evaluations (FLUX, SDXL, Imagen, ...) alongside LLMs. Everything else (sampler args, image dimensions, hashes, human-rater pools) flows through the existing `additional_details` escape hatches, which keeps the surface area small.

- **`modality`** — optional enum (`text` | `text_to_image`) on each `evaluation_results[]` entry and on each instance record. Absent means `text` for backwards compatibility.
- **`output.media: MediaRef[]`** — generated artifacts on the instance record. `MediaRef` is intentionally minimal: required `{media_type, uri}` plus an `additional_details` bag for sha256, mime_type, width/height, seed, index, etc. Required when `modality == "text_to_image"`. `media_type` is enum `[image, video, audio]` so the same shape extends to future modalities.
- **`evaluation.is_correct`** widened to `boolean | null` — set to `null` when the metric is continuous (FID, CLIPScore, ImageReward, ...).

T2I uses `interaction_type: "single_turn"` — modality is the orthogonal axis, which keeps `interaction_type` logic untouched and lets future image-edit/video records slot in via a single enum extension.

Schema `version` fields are intentionally left at `0.2.2`/`instance_level_eval_0.2.2` — bumping is a maintainer call, not part of this PR. The Pydantic models are regenerated via the documented `datamodel-codegen` pipeline (no hand edits to generated files). One new `validate_modality_consistency` model validator is added through `post_codegen.py` so the "T2I requires `output.media`" constraint is enforced at validation time.

Existing converters and their tests are completely untouched.

## What it looks like

A worked example is included at [`tests/data/t2i/geneval_sdxl_example.{json,jsonl}`](tests/data/t2i/) — minimal GenEval/SDXL fixture showing `modality: text_to_image`, T2I sampler args in `generation_config.additional_details`, `output.media[]` with `MediaRef`s, and `is_correct: null`.

## Test plan

- [x] All 134 existing tests still pass
- [x] 5 new T2I tests cover: fixture validates, missing `output.media` fails with the modality validator's error, `is_correct: null` passes, unknown modality (`image_edit`) is rejected, and records without a `modality` field continue to validate as before
- [x] `python -m every_eval_ever validate tests/data/t2i/` exits 0
- [x] Both schemas pass `jsonschema.Draft7Validator.check_schema`
- [x] Pydantic types regenerated cleanly via the documented `uv run datamodel-codegen ...` + `uv run python post_codegen.py` pipeline


Erotemic:
FYI: the link is broken.

post_codegen.py
```diff
@@ -66,6 +66,21 @@ def validate_score_type_requirements(self):
             if self.max_score is None:
                 raise ValueError("score_type 'continuous' requires max_score")
         return self
+""",
```
Erotemic, line 69: Probably not an issue with this PR, but python code in a string can be a smell. I don't quite understand the purpose of this file yet, but we may want to revisit the design that requires this ATM.

tests/test_validate.py
```diff
@@ -376,3 +376,68 @@ def test_exit_code_1_on_failure(self, tmp_path: Path):
...
         report = validate_file(fp)
         assert report.valid is False
+
+
+# ===================================================================
+# Text-to-image (modality: text_to_image) tests
+# ===================================================================
+
+
+T2I_FIXTURE_DIR = Path(__file__).parent / 'data' / 't2i'
```
Erotemic, line 386: Not a blocker, but `__file__` is almost never a robust way to handle finding resources. We should think about factoring out a EEE data repo that properly packages and provides access to the data needed for tests.

every_eval_ever/eval_types.py
```diff
@@ -443,33 +440,31 @@ class MetricConfig(BaseModel):
 
     # --- validators (added by post_codegen.py) ---
 
-    @model_validator(mode='after')
+    @model_validator(mode="after")
```
Erotemic, line 443: We should just use `ruff format` and settle on a quote style (I like single quotes) to avoid diffs like these. Not a blocker here, but something that should happen soon as the repo grows.

every_eval_ever/schemas/instance_level_eval.schema.json
```diff
@@ -82,13 +87,18 @@
...
-                    "description": "Complete model responses",
+                    "description": "Complete model responses. For text_to_image modality, populate with one placeholder string per generated artifact (e.g. '<image:0>') so indexes align with output.media[].",
                     "items": { "type": "string" }
                 },
                 "reasoning_trace": {
                     "type": ["array", "null"],
                     "description": "Reasoning traces of the model if applicable (e.g. chain-of-thought tokens)",
                     "items": { "type": "string" }
+                },
+                "media": {
```
Erotemic, line 98: I think this doesn't correspond with the pydantic validator. I'm not sure if it should. If it did it would get a bit wordy, i.e. 

```
{
  "if": {
    "required": ["modality"],
    "properties": {
      "modality": { "const": "text_to_image" }
    }
  },
  "then": {
    "required": ["output"],
    "properties": {
      "output": {
        "type": "object",
        "required": ["media"],
        "properties": {
          "media": {
            "type": "array",
            "minItems": 1
          }
        }
      }
    }
  }
}
```

The larger issue is having two sources of truth for the schema. Does it make sense to go all in on pydantic and then have it generate the jsonschema?

every_eval_ever/schemas/instance_level_eval.schema.json
```diff
@@ -334,5 +344,28 @@
...
+            "required": ["media_type", "uri"],
+            "additionalProperties": false,
+            "properties": {
+                "media_type": {
+                    "type": "string",
+                    "enum": ["image", "video", "audio"]
+                },
+                "uri": {
+                    "type": "string",
+                    "description": "Location of the artifact: 'file://...', 'https://...', 'hf://...', 's3://...', or 'data:...;base64,...' for inline."
```
Erotemic, line 361: I think the `data:...` option is a BAD idea here. This overscopes what this is. This should be a always be a reference to the data, not the data itself. If you want to support embedding the data itself it should be a separate optional field.

I think one URI is also a bad idea. There should be a list of suggested ways to access the data . Otherwise you run into a dead URL issue. It happens too often, and it limits the reproducibility  value. URLs rot, they can be changed to point to something else (where the hash is important). I would also suggest thinking about allowing distributed content addressed references like IPFS CIDs or BitTorrent magnet URIs here, which avoid the problem where the content at an address can change, but do not address the link-rot issue (which is just always going to be a fundamental limitation of any reference based scheme).

every_eval_ever/instance_level_types.py
```diff
@@ -104,8 +94,9 @@ class AnswerAttributionItem(BaseModel):
 
 class Evaluation(BaseModel):
     score: float = Field(..., description='Instance-level score')
-    is_correct: bool = Field(
-        ..., description='Whether the final answer is correct'
+    is_correct: bool | None = Field(
```
Erotemic, line 97: Is there a better way that this can be expressed that generalizes better? I don't have an immediate idea, but this seems like the stat of a boolean explosion to me.

every_eval_ever/instance_level_types.py
```diff
@@ -150,6 +141,42 @@ class Performance(BaseModel):
...
+    video = 'video'
+    audio = 'audio'
+
+
+class MediaRef(BaseModel):
+    model_config = ConfigDict(
+        extra='forbid',
+    )
+    media_type: MediaType
+    uri: str = Field(
```
Erotemic, line 155: Similar URI comment

Erotemic (commented):
I think we need a bit more discussion about what to do here. Not exactly sure what changes should be scoped here. I could also be overcomplicating things, so push back if I am.

nelaturuharsha:
@felifri could you please look at the comments/review and address them please? thanks!

every_eval_ever/schemas/instance_level_eval.schema.json
```diff
@@ -46,7 +46,12 @@
         "interaction_type": {
             "type": "string",
             "enum": ["single_turn", "multi_turn", "agentic"],
-            "description": "Type of interaction: single_turn for simple Q&A, multi_turn for conversations, agentic for tool-using agents"
+            "description": "Type of interaction: single_turn for simple Q&A, multi_turn for conversations, agentic for tool-using agents. Text-to-image evaluations use 'single_turn'; modality is the orthogonal axis."
+        },
```
mrshu, line 50: **[high] schema versioning** — New instance records keep `instance_level_eval_0.2.2`, although the existing 0.2.2 model rejects `modality` and null `is_correct`. The old aggregate model also discards the new modality under its unchanged identifier.

The extension is well scoped; its schema copies, emitters, models, fixtures, tests, and current-version docs can move to new identifiers together.

every_eval_ever/instance_level_types.py
```diff
@@ -150,6 +141,42 @@ class Performance(BaseModel):
     )
```
mrshu, line 143: **[medium] modality validation** — A `text_to_image` record with only audio or video artifacts passes validation. The current check proves only that the media list is non-empty.

One shared invariant can require at least one image item in JSON Schema and Pydantic. Missing, empty, audio-only, video-only, mixed, and image-only cases would cover the boundary.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The schema extension is promising, but its
version and media contracts need a few focused updates.
**Result:** Four actionable findings remain. Two are already covered by open
review threads, so this review does not duplicate those inline comments.
**Scope:** This PR adds text-to-image fields, media references, validation,
fixtures, generated models, and documentation.
**Checks:** 2 concerns checked; 1 confirmed and 1 narrowed.
**Second check:** 2 Codex reviewers checked 5 findings; 5 kept, 4 fix paths
clarified, and 1 item lowered in priority.

## Earlier review comments

<details>
<summary>Earlier review comments (7 open items)</summary>

- **Still open** — Validator code remains embedded in strings.
  [Thread](https://github.com/evaleval/every_eval_ever/pull/137#discussion_r3275616626)
- **Still open** — The fixture lookup still depends on `__file__`.
  [Thread](https://github.com/evaleval/every_eval_ever/pull/137#discussion_r3275625566)
- **Still open** — Generated code still bypasses the repository's final Ruff
  style. This is also `RAVF005`.
  [Thread](https://github.com/evaleval/every_eval_ever/pull/137#discussion_r3275639270)
- **Still open** — JSON Schema still accepts T2I records that Pydantic rejects.
  This is also `RAVF002`.
  [Thread](https://github.com/evaleval/every_eval_ever/pull/137#discussion_r3275677334)
- **Still open** — The schema still permits inline `data:` URIs.
  [Thread](https://github.com/evaleval/every_eval_ever/pull/137#discussion_r3275702239)
- **Still open** — The modality validator remains specialized to one modality.
  [Thread](https://github.com/evaleval/every_eval_ever/pull/137#discussion_r3275709238)
- **Still open** — The generated URI description still permits inline `data:`
  content.
  [Thread](https://github.com/evaleval/every_eval_ever/pull/137#discussion_r3275711073)

</details>

## What I noticed

|---|---|---|---|---|
| RAVF001 | high | schema versioning | `every_eval_ever/schemas/instance_level_eval.schema.json:50` | New instance records keep `instance_level_eval_0.2.2`, although the existing 0.2.2 model rejects `modality` and null `is_correct`. The old aggregate model also discards the new modality under its unchanged identifier. |
| RAVF002 | medium | schema consistency | `every_eval_ever/schemas/instance_level_eval.schema.json:98` | JSON Schema permits T2I records without media, while the Pydantic path rejects them. This remains covered by an open thread. |
| RAVF004 | medium | modality validation | `every_eval_ever/instance_level_types.py:143` | A `text_to_image` record can contain only audio or video artifacts and still pass both validation paths. |

<details>
<summary>Non-blocking low-priority findings (2 items)</summary>

- **[low] output contract** — `output.raw` promises positional alignment with
  `output.media`, but no current consumer uses that pairing. The contract can
  first choose one canonical artifact mapping. (`RAVF003`)
- **[low] generated code** — The generation flow runs Ruff before
  `post_codegen.py`, so its inserted code bypasses the final configured format.
  This remains covered by an open thread. (`RAVF005`)

</details>

<details>

- **[high] schema versioning** — Both changed contracts can receive new
  identifiers. Schema copies, emitters, generated models, fixtures, tests, and
  current-version docs can move together. (`RAVF001`)
- **[medium] media validation** — One `text_to_image` condition can require
  `output.media` and at least one image item. The Pydantic path can mirror the
  same invariant. (`RAVF002`, `RAVF004`)
- **[low] output contract** — `output.media` can remain the canonical artifact
  order. The unverified positional promise in `output.raw` can be removed.
  (`RAVF003`)
- **[low] generated code** — `ruff check --fix` and `ruff format` can run after
  `post_codegen.py` in both documented and automated flows. (`RAVF005`)

</details>

<details>

- Target: PR #137 (`feature/t2i-schema-extension`, 9 files, +279/-52)
- Mix: 3 codex-exec
- Earlier review comments: 7 open; 7 still present after checking
- Finding counts: 0 critical, 1 high, 2 medium, 2 low, 0 nit
- Checks: concerns=2; confirmed=1; narrowed=1; ruled-out=0; set-aside=0
- Second check: targeted; reviewers=2; kept=5; clarified=4; lowered=1;
  set-aside=0; removed=0; approval changed=no

</details>

---

nelaturuharsha:
Closing for now, due to lack of activity. Will re-open once other tickets have been resolved.

Issue #144, opened 2026-05-21
[Feature] Make adapters mandatory for generated data submissions


tommasocerruti:
## Motivation

We should consider making adapters mandatory by design for EEE data submissions.

Right now, if generated records can be submitted directly, it is hard to distinguish between data that was produced by a reproducible transformation and data that was manually edited after generation. This creates avoidable problems for reproducibility, reviewability, future schema migrations, and data integrity.

## Proposal

For each new data contribution, contributors should submit an adapter, or point to an existing adapter, that generates the EEE records from the original source data.

The generated records should be treated as reproducible artifacts rather than the primary trusted object. In principle, we should be able to rerun the adapter and recover the submitted records exactly, or at least detect when they differ.

## Benefits

Making adapters mandatory would give us several advantages:

* Reproducibility: we can regenerate records from the original source data.
* Reviewability: reviewers can inspect the transformation logic instead of only reviewing final JSON records.
* Schema migration: if we make large schema changes, we can update the adapters and regenerate the data.
* Data integrity: manual post-generation edits become detectable through CI checks or hash comparisons.
* Consistency: every contribution follows the same source → adapter → EEE records pipeline.
* Security: this reduces one concrete evaluation-result poisoning risk, where syntactically valid records are silently modified after generation.

## Possible CI extension

As a follow-up, we could add a reproducibility check in CI: rerun the adapter on the declared source data, normalize the generated records, and compare them against the submitted records. If they differ, the PR fails or is flagged for review.

This would not prove that the source data or adapter are correct, but it would ensure that changes happen through reviewable source data or adapter diffs, rather than silent manual edits to generated records.

nelaturuharsha:
Great stuff, we should do this!

tommasocerruti:
Update: Making adapters mandatory creates a blocker for contributors who only have aggregated results and don't have the capacity to build an adapter. Let's figure out how to handle this tradeoff in Discussion #145.

PR #157, opened 2026-06-03
Add Vectara Hallucination Leaderboard adapter

mohammadrezakarami:
# Add Vectara Hallucination Leaderboard adapter

## Summary

Adds an adapter that converts the Vectara Hallucination Leaderboard into EEE
records, sourced from the structured per-model result files in the public
[`vectara/results`](https://huggingface.co/datasets/vectara/results) Hugging
Face dataset, pinned at commit `7c104699e98ade53dd719f79ae9f7eb281c8107d`.

Companion datastore submission:
https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/144

## What it produces

- **105 models**, one `EvaluationLog` each — the full pinned roster, enumerated
  from the HF tree API rather than a hand-listed subset.
- **40 results per model** (4200 total): 4 aggregate metrics plus the
  per-category and per-text-complexity breakdowns Vectara publishes. Scores
  round-trip from the source exactly.
- Slices are encoded as `evaluation_name` + `metric_parameters`
  (`slice_kind`, `slice`), since the schema has no subset field.

## Contract notes

- `hallucination_rate` uses the **canonical registry `metric_id`**
  (`hallucination-rate`), so it joins other sources. The three
  leaderboard-specific constructs — factual consistency rate, answer rate,
  average summary length — are namespaced under
  `vectara-hallucination-leaderboard.*` because they name no registry metric.
  (Heads-up for #222: that warning would fire on those three until they are
  registered.)
- `evaluation_id` is anchored on the source run timestamp, not `now`, so
  re-running the adapter is idempotent rather than minting duplicates.
- `schema_version` comes from `SCHEMA_VERSION`, so it can't drift.
- The evaluated corpus is a private 7700+ article set that Vectara has
  deliberately not released, so each result carries `SourceDataPrivate`. The
  public `vectara/results` files are recorded as provenance in
  `source_metadata` / `additional_details`, which is where they belong — a
  leaderboard or repo URL is not dataset provenance.
- `deployment_type` is `unknown` by evidence, not by laziness: the pinned
  result files record no serving platform, and the leaderboard README's API
  notes cover only 42 bullets across 14 vendors out of 105 models, on an
  unpinned revision, with no self-hosting mentions. `model_availability` *is*
  derived, from the pinned `accessibility` annotation
  (`open` → `open_weights`, `commercial` → `closed_weights`).
- The documented "temperature 0 except where impossible" policy is recorded as
  prose rather than a per-model `generation_config` value, because the
  exceptions are not enumerated and it cannot be attributed per model.
- Constant provenance lives once in `source_metadata`, not on all 40 results.
  Repeating it made each record 106 KB; records are now 53 KB and the full
  collection 5.5 MB instead of 11.2 MB, with nothing dropped.
- Unrepresentable source values are recorded as `SourceRecordFailure` /
  `SourceRecordExclusion` in a `SourceConversionResult` and written to a
  failure report, then `raise_if_incomplete()` — no silent drops.

## Validation

```text
# full suite, on this commit
python -m pytest tests -q
353 passed, 20 skipped

# ruff
ruff check every_eval_ever tests
All checks passed!

# end-to-end: live fetch of all 105 pinned models, then the datastore gate
python -m every_eval_ever validate <105 generated files>
105 files checked, 0 invalid, 0 errors, 0 warnings
```

A live network run of the adapter is byte-identical to the checked-in fixture
at `tests/data/vectara_hallucination_leaderboard/source_rows.json`, so the pin
is stable and the offline tests are faithful to the source.

`tests/test_vectara_hallucination_leaderboard_adapter.py` validates generated
records **with `run_semantic_checks=True`, at a real 5-part datastore path** —
which is what the merge gate runs. This matters: `validate_file` defaults
`run_semantic_checks=False`, so a suite that only calls it in default mode can
pass while the gate fails. Verified by deleting the deployment keys: default
mode still reports `valid=True, errors=0`, semantic mode reports
`valid=False, errors=2`.

## Replay / refresh

```bash
# live fetch, keeping the raw snapshot
python -m every_eval_ever.adapters.vectara_hallucination_leaderboard.adapter \
    --save-raw-json /tmp/vectara_raw.json

# deterministic replay, no network
python -m every_eval_ever.adapters.vectara_hallucination_leaderboard.adapter \
    --input-json tests/data/vectara_hallucination_leaderboard/source_rows.json
```

To pick up new Vectara runs, bump `SOURCE_COMMIT` deliberately. If the tree
ever exceeds one API page, `list_result_files` raises rather than silently
converting a truncated roster.

<details>
<summary>Original description (superseded — the adapter was rewritten in place)</summary>

The original submission placed the adapter under `utils/`, covered 3 models
with 4 of 40 scores each, and hardcoded `schema_version "0.2.2"`. `utils/` was
removed from `main` in the meantime, which is what made this PR conflict.

</details>


mrshu:
# ⚒️ review-anvil report

**Review decision:** COMMENT — Approval is disabled. The adapter has only one
non-blocking error-reporting suggestion.
**Result:** The pinned scores, platforms, output layout, and schema-shaped
records match the upstream data checked by the reviewers.
**Scope:** This PR adds a Vectara Hallucination Leaderboard adapter, offline
tests, and usage documentation.
**Checks:** skipped; no material finding needed a confidence check.
**Second check:** off; the only finding is low priority and has no suggestion
block.

## Earlier review comments

None.

## What I noticed

No confirmed correctness problems found.

<details>

- **[low] error handling** — If `urllib` fails, the online path retries through
  an undeclared `curl` program. A missing or failed `curl` then hides the useful
  URL error. (`RAVF001`)

</details>

## Things to try

- **[low] error handling** — The online path can use one existing fetch pattern.
  It can use the installed `requests` dependency or preserve the original
  `URLError` with the source URL. (`RAVF001`)

<details>

- Target: PR #157 (`codex/vectara-hallucination-adapter`, 4 files, +473/-0)
- Mix: 3 codex-exec
- Finding counts: 0 critical, 0 high, 0 medium, 1 low, 0 nit
- Checks: skipped because no finding needed checking
- Second check: off; auto selected off for a comment-only low-priority report

</details>

---

borgr:
## Review

Verdict up front: **don't merge as-filed, don't close** — the source is worth
having, so I rewrote the adapter on the standard contract and pushed it to this
branch (`fc96121` → `e40b462`). Details below, including where I disagree with
the bot.

### Why the original couldn't merge

Three hard blockers, then some contract issues:

1. **Location.** The adapter lived in `utils/`, which was deleted from `main`
   while this PR was open. That is the modify/delete conflict that made the PR
   unmergeable — it wasn't a stale-branch artifact.
2. **`schema_version` hardcoded to `"0.2.2"`** while the packaged schema is
   `0.3.0`. This is exactly the drift `SCHEMA_VERSION` exists to prevent.
3. **The output failed the datastore gate 6 times**, from missing
   `deployment_type` / `model_availability` in `model_info.additional_details`.

And then, on substance:

- **3 of 105 models.** The pinned `vectara/results` tree has 105 per-model
  result files; the adapter hand-listed 3.
- **4 of 40 scores per model.** The aggregate metrics were kept and every
  per-category and per-text-complexity breakdown was dropped — silently, with
  no failure record.
- **`hallucination_rate` was namespaced** to
  `vectara-hallucination-leaderboard.hallucination_rate`. `hallucination-rate`
  is a real registry metric; namespacing it fragments the one join the
  datastore exists for.
- **`evaluation_id` derived from wall-clock time**, so every re-run would mint
  new records for unchanged evaluations instead of being idempotent.
- **`inference_platform` asserted per model** (`vertex_ai`, `azure`,
  `dashscope`). I could not source this from the pinned files — they record no
  serving platform. See the deployment note in the PR body for what I did
  instead.

### On the review-anvil comment

**Agreed on `RAVF001`** — the `requests` → `curl` subprocess fallback. It's
right on both counts and worth spelling out: `requests` is already a declared
dependency so the fallback is unreachable in a correct install, and
`subprocess.run(..., check=True)` converts a clean `URLError` into a
`CalledProcessError` that discards the HTTP status. Resolved by using
`fetch_json` and nothing else.

**One gap worth flagging, not a disagreement:** the bot reported **0
correctness problems on a diff whose output fails the merge gate 6 times.**
The gate errors were fully determined by the diff — they needed no network and
no fixtures. If the bot doesn't run generated records through
`run_semantic_checks=True`, it will keep passing adapters that then bounce off
the datastore. That's a coverage note for the bot, not a criticism of this PR.

Related trap, since it bit this PR's own tests: `validate_file` defaults
`run_semantic_checks=False`. The original suite was green *and* the output
failed the gate — those are consistent, not contradictory. The new test
validates at a real 5-part `data/<collection>/<developer>/<model>/<file>` path
with semantic checks on, and I confirmed it's load-bearing by removing the
deployment keys: default mode still says `valid=True, errors=0`, semantic mode
says `valid=False, errors=2`.

### Result after the rewrite

| | as filed | now |
|---|---|---|
| models | 3 | 105 / 105, 0 failures |
| results per model | 4 of 40 | 40 of 40 (4200 total, exact round-trip) |
| datastore gate | 6 errors | 0 errors, 0 warnings |
| suite | 4 tests, gate-blind | 352 passed / 20 skipped, gate-checked |

Diff is purely additive against current `main`: 5 files, +1254, −0.

### Two things maintainers need to action

1. **CI has not run.** The `Tests` workflow came back `action_required` —
   fork-PR runs need maintainer approval. Locally on `e40b462`: `352 passed,
   20 skipped` and `ruff check` clean, but please approve the run so that's
   confirmed in CI. I haven't approved this PR myself, since I wrote the code
   now on the branch — it wants a maintainer's eyes.

2. **The already-merged datastore records need a follow-up.** The 3 records
   published via discussion #144 fail today's validator with **9 errors**:
   `schema_version "0.2.2"`, and the legacy vocabulary `deployment_type: "api"`
   / `model_availability: "closed_source"` / `"open_weights_deployment"`, none
   of which are members of the current enums. The records are otherwise
   genuinely 0.3.0-shaped (per-result `source_data`, no top-level
   `source_data`), so it's a three-field metadata correction, not a
   re-conversion. I have the corrected files validating at 0 errors / 0
   warnings and will open that separately.

   While checking those: the merge-with-modifications replaced the original
   `SourceDataPrivate` with `source_type: "url"` /
   `dataset_name: "vectara/hallucination-leaderboard"` pointing at the GitHub
   repo. The evaluated corpus is a private 7700-article set that was never
   released, and a leaderboard citation isn't dataset provenance — the original
   PR had this right and the modification undid it. This branch restores it.

borgr:
Follow-up commit `9bcc23e` — a size fix, no behaviour change.

Each record was repeating the source file, commit, resolve URL, scoring model
and temperature policy on all 40 results, and giving every result its own full
copy of the private corpus description. Measured, that duplication was **62% of
every record**: `source_data` 34% + `metric_config.additional_details` 27.6%,
against just 2.2 KB of actual log-level metadata.

Constants now live once in `source_metadata`; per-result details carry only what
varies (`source_metric_key`, plus the diagnostic note on average summary
length). Nothing was dropped — the corpus description and availability moved up
a level rather than out.

| | before | after |
|---|---|---|
| per record | 106 KB | **53 KB** |
| 105-model collection | 11.2 MB | **5.5 MB** |

Re-verified after the change: `353 passed, 20 skipped`, `ruff` clean, all 105
records pass the gate with 0 errors / 0 warnings, and all **4200 scores still
round-trip exactly** (0 missing, 0 extra, 0 mismatched) against a fresh live
fetch. A new test pins the invariant so the duplication can't creep back.

For context on scale, this adapter is 612 lines — mid-pack among the existing
adapters — and its fixture is 8 KB, the second-smallest in `tests/data/`.

CI went green on `e40b462`; `9bcc23e` needs one more fork-run approval.

borgr:
Note on the CI check for `9bcc23e` — it is a GitHub-side artifact, not a test failure.

The run reported `failure` with **zero failing steps**: all four matrix jobs were
cancelled after 11–15 min, on a matrix where the same jobs took 15 s / 67 s /
68 s / 391 s on the previous commit `e40b462` (which passed all four). The run is
now wedged — the API reports it `queued` with no jobs assigned, while
`gh run cancel` says it is completed and `rerun` says it is already running.
Other runs in the repo have completed normally throughout, so it is not a
runner-pool outage.

Clearing it needs a fresh run, which means either an empty commit on this branch
or a close/reopen of the PR — both yours to choose, so I have not done either.

Locally on `9bcc23e`: `353 passed, 20 skipped`, `ruff check` clean, all 105
generated records pass `every_eval_ever validate` with 0 errors / 0 warnings, and
all 4200 scores round-trip exactly against a fresh live fetch. `9bcc23e` differs
from the green `e40b462` only by moving constant provenance from each of the 40
results up to `source_metadata` — no logic change.


---

**Correction and update.** The cause was not this PR: GitHub Actions was in a
`major_outage` from 15:22Z, and the incident updates state that webhook triggers
were throttled to ~15% while capacity recovered, so pushes and pull request
events largely did not create workflow runs. That explains all of it — the
starved 11–15 min cancellations, the run stuck in `queued` with `jobs: []`, and
the several hours in which this repository created no runs at all. My earlier
note here blamed the stuck run for blocking new suites; that was wrong.

Run `31117632033` on `9bcc23e` is still stuck and cannot be cleared: `cancel`
returns 409 (`Cannot cancel a workflow re-run that has not yet queued`) and
`delete` returns 403, since only completed runs are deletable. It is on a
superseded commit, so it does not gate the head.

Head is now `34d7e23`, which merges `main` after #160 landed and resolves the
`adapters/README.md` table conflict by keeping both the `lexam` and
`vectara_hallucination_leaderboard` rows. On that merged tree: `389 passed,
20 skipped`, `ruff check` clean, 105/105 generated records validate with 0 errors
and 0 warnings, and all 4200 scores round-trip exactly against a fresh live
fetch.

**CI is green on `34d7e23`** — all four matrix jobs pass
([run 31132236602](https://github.com/evaleval/every_eval_ever/actions/runs/31132236602)):
`core / locked`, `core / loose`, `full / locked`, `full / loose`. The PR is
`MERGEABLE` / `CLEAN`.

Issue #54, opened 2026-02-24
Inspect converter crashes if only a single sandbox arg in task_args is given


joeda:
Consider this Inspect eval I generated on the `humaneval` task

```
"version": 2,
  "status": "success",
  "eval": {
    "eval_id": "VcWjWK4aL5PjMZYa3rcMEd",
    "run_id": "CZWU36QrbVrvRMKNG8tKgn",
    "created": "2026-02-24T11:23:20+00:00",
    "task": "inspect_evals/humaneval",
    "task_id": "ENiBTeoXr2dbbNcDtpbVvq",
    "task_version": 2,
    "task_display_name": "humaneval",
    "task_registry_name": "inspect_evals/humaneval",
    "task_attribs": {},
    "task_args": {
      "solver": null,
      "instruction_prompt": "\nRead the following function signature and docstring, and fully implement\nthe function described. Your response should only contain the code for\nthis function.\n\n",
      "scorer": null,
      "sandbox": "local"
    },
```

It leads to a crash as the parser incorrectly assumes the `sandbox` variable in `task_args` to be a list here

https://github.com/evaleval/every_eval_ever/blob/main/eval_converters/inspect/adapter.py#L273

PR #160, opened 2026-06-08
Add LEXam public leaderboard converter

JoelNiklaus:
# Add LEXam public leaderboard converter

## Problem

Every Eval Ever supports converting local eval logs and some public leaderboards (e.g. AlpacaEval), but not LEXam - a legal reasoning benchmark with a public leaderboard and Hugging Face dataset.

## Solution

Add a `lexam` adapter that:

- Fetches the public leaderboard HTML from the LEXam project website repo
- Parses open-question judge scores and MCQ accuracy for each model
- Emits schema-valid Every Eval Ever JSON with `source_metadata.source_type = documentation`
- Links underlying benchmark data to `LEXam-Benchmark/LEXam` on Hugging Face
- Records LLM-judge metadata for open-question scoring
- Writes output under the stable `data/lexam/{developer}/{model}` layout
- Uses explicit model identity mappings and reports newly unmapped leaderboard
  entries in a provenance report instead of silently dropping them

Usage (adapters are run as modules, not through the `convert` CLI):

```bash
uv run python -m every_eval_ever.adapters.lexam.adapter --output-dir data
```

## What the records carry beyond the two scores

Everything below is grounded in a primary source, cited in the record itself:

- **Canonical ids from the eval-card-registry.** Model, benchmark, metric and
  harness ids all resolve through the registry;
  `model_info.additional_details.model_id_resolution` says whether an id came
  from a confirmed alias, a direct canonical match, or a Hugging Face id used
  because the registry has no entry for the evaluated checkpoint. The entities
  this adds are proposed in evaleval/eval-card-registry#54.
- **Registry values are pinned, not copied.** `registry_snapshot.json` vendors
  only the entities this adapter emits, pinned to the registry revision they
  came from; the tests fail if any of them drifts, and
  `refresh_registry_snapshot.py --check` answers "is the pin stale?" after a
  registry PR merges.
- **Canonical metric scales.** `accuracy` is a proportion on `[0, 1]` registry
  wide, so the leaderboard's percentage is converted onto that scale (with its
  standard error); the judge score keeps the `[0, 100]` scale of its registry
  entry. The published percentage is preserved in `details`.
- **Standard errors from the paper.** The leaderboard page publishes none, so
  they come from arXiv:2505.12864v7 Table 1 (open) and Table 10 (MCQ-4), and
  are attached only while the scraped score still equals the score the paper
  reports - a leaderboard update drops the standard error rather than pairing
  it with a number it was never computed for.
- **Inference facts, derived rather than left `unknown`.** §3.3 and appendix F
  state settings and endpoints per model group, and the paper's own
  Reasoning / Large / Small bracketing (17/8/11) assigns each row to one. Where
  LEXam's own runner (`litellm_eval.py`) names a model, the exact served string
  and its sampling arguments are recorded too - that config predates the
  post-paper rows, so it covers 15 of 36 models and nothing is extrapolated to
  the others.
- **Conflicts are recorded, not resolved.** For `Gemma-3-12B-it` the runner
  routes through Together AI while appendix F puts every 7-14B conventional
  model on local vLLM; `deployment_type` follows the paper and
  `served_model_note` records the disagreement.
- **The judge ensemble's aggregation.** LEXam takes the pointwise minimum of
  three judges, which the schema's `AggregationMethod` enum cannot express, so
  no typed value is asserted and the method goes in
  `llm_scoring.additional_details`.

## Scope of the two metrics

| Metric | Evaluation | Scope |
|---|---|---|
| Open Question Judge Score | `lexam.open_question` | `open_question` **test** split, n=2,541 |
| Multiple-Choice Accuracy | `lexam.mcq_4_choices` | `mcq_4_choices`, n=1,655 - the site column reproduces the paper's MCQ-4 table and does **not** pool the 8/16/32-choice configs |

## Testing

- 33 unit tests against a frozen HTML fixture: parsing, medal stripping, metric
  combination, schema validation, missing-section errors, unmapped models,
  registry-snapshot drift, and the paper's model grouping
- Live conversion against the current leaderboard (36 models), all 36 records
  validated with `every_eval_ever validate`
- Every open-question score and standard error checked against Table 1, and
  every MCQ pair against Table 10: no mismatches
- `uv run pytest tests -q`, `uv run ruff check`, `uv run ruff format`

Generated records are submitted separately to the datastore
(`evaleval/EEE_datastore` PR 179).


JoelNiklaus:
Not sure if you want the references to the eval converter in the README or only the general evaluation framework specific ones?

tests/data/lexam/leaderboard.html
DeepLumiere, line 1: We need an html for this, I couldn't find a previous one from tests with HTML code, all have got JSON

Image

borgr: This should parse the dataset leaderboard ( https://lexam-benchmark.github.io/#leaderboard ). so I'm not sure what html is missing?

every_eval_ever/converters/lexam/adapter.py
```diff
@@ -0,0 +1,388 @@
...
+
+LEADERBOARD_URL = (
+    'https://raw.githubusercontent.com/LEXam-Benchmark/'
+    'lexam-benchmark.github.io/main/index.html'
+)
+LEADERBOARD_PAGE_URL = 'https://lexam-benchmark.github.io/'
+HF_REPO = 'LEXam-Benchmark/LEXam'
+BENCHMARK_KEY = 'lexam'
+OPEN_QUESTIONS_SAMPLES = 2541
+MCQ_SAMPLES = 4696
```
mrshu: **[high] metric provenance** — This leaderboard score covers `mcq_4_choices` and 1,655 rows, but the output records all four MCQ configs and 4,696 rows.

The converter structure is clear; its source, metric text, README, and tests can consistently describe the single published config.

every_eval_ever/converters/lexam/adapter.py
```diff
@@ -0,0 +1,388 @@
...
+                    id='qwen/Qwen3-32B',
+                    developer='qwen',
+                ),
+            ),
+        ],
+        input_prompt=(
+            'Expert-validated LLM-as-a-Judge ensemble scoring open-ended '
+            'law exam answers against reference answers.'
+        ),
+        aggregation_method=AggregationMethod.average,
```
mrshu: **[high] judge metadata** — LEXam publishes the pointwise minimum across three judges, while this field records average aggregation.

The typed average value can be omitted. `additional_details` can record minimum aggregation until the schema supports it directly.

every_eval_ever/converters/lexam/adapter.py
```diff
@@ -0,0 +1,388 @@
...
+
+def _extract_section_rows(
+    html: str, section_title: str
+) -> list[LeaderboardRow]:
+    """Parse model/score rows from the table under *section_title*."""
+    title_idx = html.find(section_title)
+    if title_idx == -1:
+        raise ValueError(f'Leaderboard section not found: {section_title}')
+
+    table_start = html.find('<table', title_idx)
```
mrshu: **[medium] HTML parsing** — This search can use the next section's table when the selected heading remains but its table is absent. MCQ scores then become open-question judge scores.

The complete table can be required to end before the next leaderboard heading. A focused missing-table test would cover the path.

every_eval_ever/converters/lexam/adapter.py
```diff
@@ -0,0 +1,388 @@
...
+    ),
+    'Llama-4-Maverick': ModelIdentity(
+        'meta-llama', 'meta-llama/Llama-4-Maverick'
+    ),
+    'Ministral-8B-it': ModelIdentity('mistralai', 'mistralai/Ministral-8B-it'),
+    'O3-mini': ModelIdentity('openai', 'openai/O3-mini'),
+    'Phi-4': ModelIdentity('microsoft', 'microsoft/Phi-4'),
+    'QwQ-32B': ModelIdentity('qwen', 'qwen/QwQ-32B'),
+    'Qwen-2.5-7B-it': ModelIdentity('qwen', 'qwen/Qwen-2.5-7B-it'),
+    'Qwen3-235B': ModelIdentity('qwen', 'qwen/Qwen3-235B'),
```
mrshu: **[medium] model identity** — Two abbreviated Qwen labels emit IDs that do not identify the evaluated variants. The published IDs include their architecture, mode, and version qualifiers.

The mappings can use the complete IDs published with the matching LEXam scores.

every_eval_ever/converters/lexam/adapter.py
```diff
@@ -0,0 +1,388 @@
...
+            ),
+            JudgeConfig(
+                model_info=ModelInfo(
+                    name='Qwen3-32B',
+                    id='qwen/Qwen3-32B',
+                    developer='qwen',
+                ),
+            ),
+        ],
+        input_prompt=(
```
mrshu: **[medium] judge metadata** — `input_prompt` contains a summary instead of the prompt template used by the judges. Consumers cannot reconstruct the scoring request from this field.

The user template can keep its placeholders here. The published system message can stay separate in `additional_details`.

every_eval_ever/converters/lexam/adapter.py
```diff
@@ -0,0 +1,388 @@
...
+                        },
+                    ),
+                    source_metadata=SourceMetadata(
+                        source_name='LEXam Leaderboard',
+                        source_type=SourceType.documentation,
+                        source_organization_name='LEXam-Benchmark',
+                        source_organization_url=(
+                            'https://github.com/LEXam-Benchmark/LEXam'
+                        ),
+                        evaluator_relationship=EvaluatorRelationship.collaborative,
```
mrshu: **[medium] source metadata** — Every converted result is marked collaborative, although LEXam-Benchmark evaluates models from unrelated developers.

A `third_party` default would match normal public-leaderboard runs. The CLI can still preserve an explicit collaborative override.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The converter has several concrete metadata
issues that are worth addressing before its output becomes public data.
**Result:** The overall converter shape is clear. Seven focused findings remain,
including two that affect every converted result.
**Scope:** This PR adds a LEXam public leaderboard converter, CLI wiring, tests,
and user documentation.
**Checks:** 3 upstream-specific concerns checked; 3 confirmed.
**Second check:** 2 Codex reviewers checked 7 findings; 7 kept and 5 fix paths
clarified.

## Earlier review comments

- **Fixed** — The requested frozen HTML fixture now exists and the adapter tests
  use it. [Prior thread](https://github.com/evaleval/every_eval_ever/pull/160#discussion_r3407580130)

## What I noticed

|---|---|---|---|---|
| RAVF006 | high | metric provenance | `every_eval_ever/converters/lexam/adapter.py:41` | The leaderboard publishes `mcq_4_choices` over 1,655 rows. The converter records all four MCQ configs and 4,696 rows. |
| RAVF007 | high | judge metadata | `every_eval_ever/converters/lexam/adapter.py:250` | LEXam uses the pointwise minimum across three judges. The converter records average aggregation. |
| RAVF001 | medium | HTML parsing | `every_eval_ever/converters/lexam/adapter.py:157` | A section without its own table can consume the next section's table and assign MCQ scores as judge scores. |
| RAVF002 | medium | model identity | `every_eval_ever/converters/lexam/adapter.py:122` | Two abbreviated Qwen labels emit IDs that do not identify the full model variants evaluated by LEXam. |
| RAVF003 | medium | judge metadata | `every_eval_ever/converters/lexam/adapter.py:246` | `input_prompt` stores a prose summary instead of the published prompt template. |
| RAVF004 | medium | source metadata | `every_eval_ever/converters/lexam/adapter.py:372` | Every result defaults to a collaborative relationship, although LEXam-Benchmark evaluates unrelated model developers. |

<details>

- **[low] CLI metadata** — The shared CLI accepts
  `--source-organization-logo-url`, but the LEXam handler does not apply it.
  (`RAVF005`)

</details>

<details>
<summary>Things to try (7 items)</summary>

- **[high] metric provenance** — The source, metric text, README, and tests can
  consistently describe `mcq_4_choices` and 1,655 rows. (`RAVF006`)
- **[high] judge metadata** — The false typed average value can be removed.
  `additional_details` can record minimum aggregation until the schema supports
  it directly. (`RAVF007`)
- **[medium] HTML parsing** — Each complete table can be required to end before
  the next leaderboard heading. A surviving heading without its table can
  cover this case. (`RAVF001`)
- **[medium] model identity** — The two labels can map to the complete model IDs
  published with their LEXam results. (`RAVF002`)
- **[medium] judge metadata** — `input_prompt` can keep the published user
  template and its placeholders. The system message can remain separate in
  `additional_details`. (`RAVF003`)
- **[medium] source metadata** — Both defaults can use `third_party`. The CLI
  can still assign an explicit `collaborative` value for known exceptions.
  (`RAVF004`)
- **[low] CLI metadata** — A supplied logo URL can be applied beside the other
  source metadata overrides. (`RAVF005`)

</details>

<details>

- Target: PR #160 (`feature/lexam-converter`, 8 files, +691/-6)
- Mix: 3 codex-exec
- Earlier review comments: 1 resolved item; 1 fixed after checking
- Finding counts: 0 critical, 2 high, 4 medium, 1 low, 0 nit
- Checks: concerns=3; confirmed=3; ruled-out=0; set-aside=0; lowered=0
- Second check: targeted; reviewers=2; kept=7; clarified=5; set-aside=0;

</details>

---

borgr:
Thanks for this @JoelNiklaus — the parsing and metric shape were easy to follow, which made it possible to check the provenance claims against LEXam's own sources. I went through the review findings and the underlying data, and pushed a branch with fixes rather than leave the thread with a list: [`borgr:lexam-160-fixes`](https://github.com/evaleval/every_eval_ever/compare/main...borgr:every_eval_ever:lexam-160-fixes) (your two commits are preserved in its history). Happy to hand it over as commits on your branch, or to open it as a follow-up — whichever the maintainers prefer.

## Two things that would change the published data

**1. The MCQ column is `mcq_4_choices` (n=1,655), not the union of the four configs (n=4,696).** The site's "Accuracy on Multiple-Choice Questions" reproduces the paper's MCQ-4 table verbatim (GPT-5 62.65, Claude-4.5-Sonnet 58.01, GPT-4.1 54.40, GPT-4o-mini 40.96 …), and the published bootstrap standard errors — ±1.17 at p≈0.63 — are consistent with n=1,655; n=4,696 would give ≈±0.71. 30 of the 31 rows match Table 10 exactly (the site swaps Apertus-8B for EuroLLM-9B-it). So `samples_number`, the config list and the metric description all needed narrowing.

**2. The judge ensemble is a pointwise *minimum*, not an average.** Paper §4: "we adopt an minimum-score ensemble of GPT-4o, Qwen3-32B, and DeepSeek-V3", and "By aggregating pointwise minimum scores". The schema's `AggregationMethod` enum only offers `majority_vote | average | weighted_average | median`, so there is no correct typed value — the fix is to drop the `average` and record the method in `llm_scoring.additional_details`. (This looks like a genuine schema gap; worth an issue if you agree.)

## Also fixed

- **`input_prompt`** now carries the published judge template with its `{question_fact}` / `{ref_answer}` / `{model_answer}` placeholders, with the system message in `additional_details`. The template in `customized_judge_async.py` and lighteval's `20250324` entry are identical in body, so there is no ambiguity about which to use.
- **Section-bounded table parsing.** `html.find('<table', title_idx)` was unbounded, so a heading whose own table went missing would consume the *next* section's table and publish MCQ accuracies as judge scores. Now bounded by the next `<h1>`, with a regression test.
- **`evaluator_relationship` → `third_party`.** The axis is relative to the *model developer*, and LEXam-Benchmark scores models it did not build — matching `alpaca_eval` and `artificial_analysis` in this repo.
- **`eval_library`** named the benchmark with a fabricated `"1.0"`. LEXam documents evaluation through HuggingFace **lighteval** community tasks, so that is the harness; version is `unknown` and the benchmark moved to `additional_details`.
- **Metric ids and scales now come from the eval-card-registry**, not from the adapter. `lexam.mcq_accuracy` becomes the canonical `accuracy` — the schema asks for "a canonical global id when applicable", and that metric is canonically a proportion on `[0,1]` registry-wide, so the leaderboard's `62.65` is emitted as `0.6265` (`unit: proportion`) with the raw percentage kept in `score_details.details`. The judge score has no global equivalent, so it uses a registry-shaped slug `lexam-open-question-judge-score` on `[0,100]`, matching the existing `mmau-pro-open-ended-judge-score` / `longaudiobench-overall-judge-score` entries — that one keeps its published scale. `metric_kind` for it is `judge_score` (the value `mt_bench` uses) rather than `accuracy`. Standard errors are rescaled with their score. This follows the `paperswithcode` adapter, where bounds and direction come from the registry and never from the scale a leaderboard printed.
- **Model ids** are resolved against the eval-card-registry instead of constructed. As written, 33 of the 36 ids did not resolve anywhere: `meta-llama/Llama-3.3-70B-it` (→ `-Instruct`), `mistralai/Ministral-8B-it` (→ `Ministral-8B-Instruct-2410`), `qwen/Qwen3-235B` (→ `Qwen/Qwen3-235B-A22B`), plus casing (`qwen/` → `Qwen/`, `openai/GPT-OSS-120B` → `openai/gpt-oss-120b`). Now 29 confirmed aliases / 4 direct canonical matches / 3 HF-canonical where the registry has no entry, and every record reports which in `model_info.additional_details.model_id_resolution`.
- **`model_availability`** is set per model (23 open / 13 closed weights) instead of shipping the library's auto-filled `"unknown"` for the model *and* all three judges. `deployment_type` stays `unknown` on purpose: LEXam documents both API-served (litellm) and local vLLM paths without saying which produced each row.
- **Standard errors** from the paper's Table 1 / Table 10, with `method='bootstrap'` and the citation in `score_details.details`. They are attached only while the scraped score still equals the score the paper reports, so a leaderboard update drops the standard error instead of pairing it with a number it was never computed for (66 of 67 scores carry one).
- **`--source_organization_logo_url`** is now applied. (The same gap exists in the AlpacaEval handler — separate fix.)

## Two things the audit caught afterwards

- **One retrieval timestamp per run.** Converting label-by-label called the clock per record, so a single run produced 36 different `retrieved_timestamp` values (and re-parsed the page 36 times). Parsing and the timestamp now happen once.
- **A registry drift guard.** `registry_snapshot.json` vendors just the entities this adapter emits — subject and judge model ids, metric ids with their canonical bounds, the harness, and the benchmark each `evaluation_name` resolves to — pinned to the registry revision it came from and regenerated by `refresh_registry_snapshot.py`. A test asserts every emitted value still matches it, so an invented metric id or a stale bound fails CI instead of shipping. I checked it earns its place: reintroducing the original `lexam.mcq_accuracy` + `[0,100]` defect fails four tests.

## Two structural notes

**The CLI path no longer ran.** After #218 removed `_write_log`, `every_eval_ever convert lexam` raised `NameError` on merge — git merges cleanly and the 13 tests still passed, because they only exercised the adapter functions. Worth a CLI-level test whichever way this lands.

**Placement.** Public leaderboard scrapes now live in `every_eval_ever/adapters/` (23 of them, "run manually, not via the main CLI"), and the AlpacaEval entry under `converters/` is documented as a legacy exception. I moved it to `every_eval_ever/adapters/lexam/` with the adapters convention (`parse_args`/`run`/`export`, `SourceConversionResult` so one unmapped label is reported instead of aborting the whole run). That also answers your README question — it belongs in `adapters/README.md`, not the converters table. Say the word if maintainers prefer it stay CLI-wired.

## Where your input would help

1. **`DeepSeek-V3.2-chat` / `-reasoner`** — I read these as the two DeepSeek API modes: the changelog (2025-12-01) documents `deepseek-chat` and `deepseek-reasoner` as the non-thinking and thinking modes of `DeepSeek-V3.2`, which matches the paper listing `-reasoner` under reasoning models and `-chat` under large models. So both rows now share `model_info.id` = `deepseek-ai/DeepSeek-V3.2` and differ in `generation_config.generation_args.reasoning`, with `DeepSeek-V3.2-Exp` kept as its own checkpoint. Is that right, or were these distinct checkpoints?
2. **`Qwen3-Next` and `Llama-4-Maverick`** — the labels do not name a variant. The registry resolves them to `alibaba/qwen3-next` and `meta-llama/Llama-4-Maverick-17B-128E`; your `litellm_eval.py` suggests `Llama-4-Maverick-17B-128E-Instruct-FP8` for the earlier runs. If you have the exact served model strings for the current rows, I will pin them.

Also: which harness produced the newest rows (lighteval or `litellm_eval.py`)? I recorded `lighteval` as the documented path.

Registry side, so future LEXam consumers get this for free, there is a companion eval-card-registry PR ([evaleval/eval-card-registry#54](https://github.com/evaleval/eval-card-registry/pull/54)): `lexam` + `lexam-open-question` + `lexam-mcq-4-choices` as benchmarks, `lighteval` as a harness, the judge metric above with its rubric cited, alias bridges for the leaderboard's model labels, and three checkpoints that exist on the Hub but no generator carried (`DeepSeek-V3.2-Exp`, `Llama-3.1-405B-Instruct`, `EuroLLM-9B-Instruct`). I rebuilt that registry locally and confirmed all 14 raw forms this adapter relies on resolve to the intended canonical.

Verification on the branch: 27 adapter tests, full suite 355 passed / 20 skipped, `ruff check` clean, a live run writes all 36 models, all 67 scores match the live page exactly, all 36 records pass `every_eval_ever validate`, and `check-duplicates` finds none.

JoelNiklaus:
I think the model strings you found for DeepSeek, Qwen and Llama are correct, but I am not 100% sure anymore. 

Yes, we evaluated with lighteval.

borgr:
I've pushed the fixes onto this branch directly (maintainer edits were enabled), so #160 is now the thing to merge rather than a pointer elsewhere — your two commits are still at the base of the history. Summary of what changed is in [my earlier comment](https://github.com/evaleval/every_eval_ever/pull/160#issuecomment-5198043487); three things happened since:

**Companion registry PR:** [evaleval/eval-card-registry#54](https://github.com/evaleval/eval-card-registry/pull/54) registers `lexam` and its two published sub-tasks as benchmarks, `lighteval` as a harness, the judge metric with its rubric cited, and alias bridges for the leaderboard's model labels. All 36 model ids, both metric ids, the harness and both `evaluation_name`s now resolve against the registry, verified by rebuilding it locally and resolving each raw form.

**Data submitted:** [EEE_datastore PR #179](https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/179) — the 36 records, all passing `validate`. Worth merging #54 first: until it lands, the records report the judge metric's `review_status` as `draft`, which is accurate but only temporarily.

**Two open questions for you**, unchanged and both about which weights were run:

1. `DeepSeek-V3.2-chat` / `-reasoner` — I read these as the two API modes of one `DeepSeek-V3.2` release (per DeepSeek's 2025-12-01 changelog, and matching the paper listing `-reasoner` under reasoning models and `-chat` under large ones), so they share `model_info.id` and differ in `generation_config.generation_args.reasoning`. Correct, or were they distinct checkpoints?
2. `Qwen3-Next` and `Llama-4-Maverick` — the labels don't name a variant. Both Qwen3-Next variants are registered (`-80B-A3B-Instruct` and `-Thinking`), so the adapter can't pick without guessing and currently uses the variant-agnostic id. If you have the served model strings for the current rows, I'll pin them.

Also still curious which harness produced the newest rows — I recorded `lighteval` as the documented path.

One note on the diff: this branch has `main` merged into it, so GitHub shows ~176 files against the June fork point. I have a rebased version that shows only the 7 LEXam files with identical content and your commits preserved; happy to force-push that if you'd rather review it that way — just say so, since it rewrites the branch.

borgr:
Thanks — the lighteval confirmation was the useful one, and it turned out the rest was answerable from your own sources rather than from you. Withdrawing both questions:

**DeepSeek-V3.2-chat / -reasoner** — settled by Table 1 itself: it lists `-reasoner` in the Reasoning block and `-chat` in the Large block, which is exactly the shape of two API modes of one release rather than two checkpoints (and matches DeepSeek's 2025-12-01 changelog). Kept as one `model_info.id` differing in `generation_config.generation_args.reasoning`. `DeepSeek-V3.2-Exp` is listed separately in Reasoning and is treated as its own checkpoint.

**Llama-4-Maverick** — `litellm_eval.py` names it: `together_ai/meta-llama/Llama-4-Maverick-17B-128E-Instruct-FP8`. Recorded.

Three things that changed as a result of re-reading v7 and the runner config:

1. **`lighteval` on all 36 rows.** I had been shipping `eval_library: unknown` for the 17 reasoning models because §3.3 says lighteval did not support them at the time of writing. Your confirmation covers the current rows, so every row names lighteval and the reasoning rows keep that caveat in `eval_library.additional_details.harness_note`, so both statements and their scopes stay readable.
2. **Nothing reports `unknown` serving any more.** Appendix F ("LLM Endpoints") states how each group was served, so `deployment_type`, `inference_platform` and `inference_engine` are derived: local vLLM on one A100 for the 7–14B conventional models, official APIs for the closed ones, Together AI for the rest. Settings likewise come from §3.3 + appendix F per group, including the per-model departures (DeepSeek-R1 and QwQ-32B at 0.6, O3-mini reasoning effort high, Claude-3.7-Sonnet budget 4096).
3. **Served strings and sampling args from `litellm_eval.py`** for the 15 rows its config covers — including `top_p 0.95` for QwQ-32B and `temperature 0.6 / top_p 0.95 / top_k 20` for Qwen3-235B. That snapshot predates the post-paper rows, so the other 21 record no served string rather than an extrapolated one. One conflict recorded rather than resolved: the config routes `Gemma-3-12B-it` through Together AI while appendix F puts every 7–14B conventional model on local vLLM — `deployment_type` follows the paper and `served_model_note` states the disagreement.

Also fixed a grouping bug this shook out: I had `Apertus-70B` bracketed with the small models; Table 1 puts it in the Large block. The group drives the harness, the settings and the serving facts, so that one row had all three wrong. There's now a test asserting the 17/8/11 bracketing against the table.

Still one thing only you can settle, and it is not blocking: **`Qwen3-Next`** — the label and both tables leave the variant unnamed, and the runner config predates it. Its Reasoning-block placement points at `Qwen/Qwen3-Next-80B-A3B-Thinking`, but I did not want to infer a checkpoint from a table position, so the record uses the variant-agnostic id. Say the word and I'll pin it.

State of the two companions: [eval-card-registry#54](https://github.com/evaleval/eval-card-registry/pull/54) (benchmarks, harness, judge metric, alias bridges) and [EEE_datastore #179](https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/179) (the 36 records, all passing `validate`, now carrying everything above). Worth landing #54 first so no record cites a `draft` metric.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This run is comment-only, with three medium suggestions and one low-priority note.
**Result:** Most earlier concerns are fixed. Three focused changes would make installed and future conversions more reliable.
**Scope:** This PR adds the LEXam public leaderboard adapter and migrates it into the current package layout.
**Checks:** 3 concerns checked; 1 confirmed, 1 ruled out, and 1 lowered in priority.
**Second check:** Targeted review by 2 Codex agents kept 4 findings and clarified 3 fix paths.

## Earlier review comments

<details>
<summary>Earlier review comments (8 grouped items)</summary>

- **Fixed:** The [HTML fixture request](https://github.com/evaleval/every_eval_ever/pull/160#discussion_r3407580130) now has a frozen fixture and parser coverage.
- **Fixed:** The [metric scope concern](https://github.com/evaleval/every_eval_ever/pull/160#discussion_r3677114326) now uses `mcq_4_choices` and 1,655 samples.
- **Fixed:** The [judge aggregation concern](https://github.com/evaleval/every_eval_ever/pull/160#discussion_r3677114332) now records pointwise minimum without a false average value.
- **Fixed:** The [table boundary concern](https://github.com/evaleval/every_eval_ever/pull/160#discussion_r3677114339) now stops at the next section heading.
- **Still present:** The [Qwen3-Next identity concern](https://github.com/evaleval/every_eval_ever/pull/160#discussion_r3677114345) remains after the adapter move. It appears below as `RAV-RUN2-R1-F003` without a duplicate inline thread.
- **Fixed:** The [judge prompt concern](https://github.com/evaleval/every_eval_ever/pull/160#discussion_r3677114351) now stores the published prompt template.
- **Fixed:** The [source relationship concern](https://github.com/evaleval/every_eval_ever/pull/160#discussion_r3677114358) now uses `third_party`.
- **No longer relevant:** The shared CLI metadata notes in the [earlier review](https://github.com/evaleval/every_eval_ever/pull/160#pullrequestreview-4811750646) do not apply to the standalone adapter.

</details>

## What I noticed

ID legend: `RUN` is the observed PR review run, `R` is the origin round, `F` is a finding, and `P` is a plan.

|---|---|---|---|---|
| RAV-RUN2-R1-F001 | medium | source-row-accounting | `every_eval_ever/adapters/lexam/adapter.py:802` | The parser visits only exact row matches. Unsupported row markup can therefore produce an incomplete successful conversion. (inline) |
| RAV-RUN2-R1-F003 | medium | model-identity | `every_eval_ever/adapters/lexam/adapter.py:701` | `Qwen3-Next` remains joined to a variant-neutral ID, although the record classifies it as the reasoning variant. |
| RAV-RUN2-R1-F004 | medium | packaging | `pyproject.toml:61` | Built wheels omit the LEXam registry snapshot. Installed conversions therefore emit unknown registry revision and metric review status. (inline) |

<details>

- **RAV-RUN2-R1-F005 [low] metric-metadata** `every_eval_ever/adapters/lexam/adapter.py:1016` — The MCQ description says 0–100, while the emitted value and bounds use a 0–1 proportion.

</details>

## Things to try

<details>
<summary>Suggested changes (4 items)</summary>

- **[medium] source-row-accounting** — Each bounded table can return valid rows and structured failures together. Explicit data-row and duplicate checks would preserve valid output while accounting for malformed cells and scores. (`RAV-RUN2-R1-P001`; covers `RAV-RUN2-R1-F001`)
- **[medium] model-identity** — The full Thinking checkpoint can be used when LEXam-specific evidence identifies it. Otherwise, the ambiguous row can become a conversion failure while other models still publish. (`RAV-RUN2-R1-P002`; covers `RAV-RUN2-R1-F003`)
- **[medium] packaging** — An exact snapshot path or deliberate snapshot pattern can include the runtime JSON. A wheel-install test can verify emitted registry metadata outside the source tree. (`RAV-RUN2-R1-P003`; covers `RAV-RUN2-R1-F004`)
- **[low] metric-metadata** — The description can name the emitted 0–1 proportion and retain the source percentage explanation. (`RAV-RUN2-R1-P004`; covers `RAV-RUN2-R1-F005`)

</details>

<details>

- Target: PR #160 (`feature/lexam-converter`, 176 files, +20535/-8614)
- Head: `e77aaa9abf5e588e0ce8a39a55bf5888ab5ebbc2`
- Run ordinal: 2
- Mix: 3 codex-exec
- Earlier review comments: 8 grouped items; 1 still present, 6 fixed, 1 no longer relevant
- Finding counts: 0 critical, 0 high, 3 medium, 1 low, 0 nit
- Checks: concerns=3; confirmed=1, ruled-out=1, lowered=1; one deployment claim was removed
- Second check: targeted; reviewers=2; kept=4, fix paths clarified=3, set aside=0, removed=0; approval unchanged
- Set aside: 0 items

</details>

---

every_eval_ever/adapters/lexam/adapter.py
```diff
@@ -0,0 +1,1339 @@
...
+        raise ValueError(f'No table found in section: {section_title}')
+
+    table_end = html.find('</table>', table_start, section_end)
+    if table_end == -1:
+        raise ValueError(
+            f'Table not closed within section: {section_title}'
+        )
+
+    table_html = html[table_start:table_end]
+    row_re = re.compile(
```
mrshu, line 755: **RAV-RUN2-R1-F001 [medium] source-row-accounting** — The parser visits only rows that match the exact three-cell pattern. A changed cell or score can disappear before failure accounting, and `total_records` then counts only the rows that parsed.

The current page matches this shape. A future markup change can still publish incomplete output as successful. Each bounded table can return valid rows and structured failures together, including duplicate model-metric rows.

pyproject.toml
```diff
@@ -47,7 +57,12 @@ include-package-data = true
 include = ["every_eval_ever*"]
 
 [tool.setuptools.package-data]
-every_eval_ever = ["schemas/*.json"]
+every_eval_ever = [
+    "schemas/*.json",
```
mrshu, line 61: **RAV-RUN2-R1-F004 [medium] packaging** — The package-data list omits `every_eval_ever/adapters/lexam/registry_snapshot.json`. A built wheel then falls back to `unknown` registry revision and metric review status during LEXam conversion.

An exact snapshot path or deliberate snapshot pattern would keep the runtime file in wheels. A wheel-install test can cover the installed-package path.

borgr:
Well, as you can see we did a lot to get those things together. Now ready.

borgr:
Went through the four findings in the run-2 report. Three were right and are fixed; the fourth I checked and am deliberately leaving as it is, with the reasoning below rather than as an open thread.

**F004 packaging — correct, and it was already biting.** `pyproject.toml` listed `adapters/**/*.md` and `*.sh` but no `.json`, so `registry_snapshot.json` never reached a wheel and an installed conversion reported `bound_registry_revision: unknown` and `metric_registry_review_status: unknown` — silently, since the loader falls back rather than fails. Added the pattern, and `tests/test_packaged_resources.py::test_adapter_data_files_are_included_in_wheels` now fails if any `adapters/**/*.json` is not matched by a package-data pattern. Checked both directions: the test fails on the unfixed tree (`these adapter data files would be missing from a built wheel: adapters/lexam/registry_snapshot.json`) and a real `python -m build` now contains `every_eval_ever/adapters/lexam/registry_snapshot.json`.

**F001 row accounting — correct.** `row_re.finditer` skipped anything that did not match the three-cell shape, so a changed cell would have shortened the leaderboard and `total_records` would have counted only what parsed. Each bounded table now collects its `<tr>` blocks containing a `<td>` first and fails if the count read differs from the count present, plus a duplicate-model-row check, since two rows under one label would otherwise collapse to whichever came last. Two tests cover it (`Read 2 of 3 data rows`, `Duplicate model rows`), and both guards are silent on the live page: 36/36 and 31/31 rows, all labels unique.

One deviation from the suggested fix path, which proposed returning valid rows and structured failures together. `SourceConversionResult` is the right shape for an *unmapped label* — a new model appears, that row is reported and the other 35 still publish. Markup that does not parse is a different case: the failure is not attributable to a row we can name, and a 35-row artifact presented as a successful conversion of a 36-row leaderboard is wrong rather than partial. So this one raises. Happy to switch if maintainers prefer uniform per-row failure reporting.

**F005 metric description — correct.** The MCQ description still said "(0-100 scale)" while the record emits `0.4783` with `min_score: 0.0` / `max_score: 1.0` / `unit: proportion`. It now names the emitted proportion and keeps the note that the leaderboard publishes the same value as a percentage.

**F003 Qwen3-Next — checked, keeping the variant-agnostic id.** I went looking for LEXam-specific evidence, which is the condition the suggested fix path attaches to pinning the Thinking checkpoint. There is none: the paper never names a variant, the leaderboard page renders the bare label `Qwen3-Next`, and `litellm_eval.py` predates the row and has no entry for it. The only pointer is Table 1's Reasoning-block placement, and I do not want to derive a checkpoint identity from a table position — the block is what drives the harness note, the sampling settings and the serving facts, and those are group-level facts, whereas an id is a claim about which weights ran. The other branch of the suggested path — dropping the row to a conversion failure — would remove a score the leaderboard genuinely publishes, in exchange for no gain in accuracy: `alibaba/qwen3-next` is a real registry entity, deliberately variant-neutral, so the record is imprecise rather than wrong. `model_info.additional_details.model_id_note` now says so explicitly, including that no LEXam source names the variant, so a consumer joining on model id sees the limit instead of inferring precision. @JoelNiklaus, one word from you pins it and I will update the record.

Datastore side, [EEE_datastore #179](https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/179) carries both content changes (the MCQ description on 31 records, the note on 1), written in place so the diff is one line per record rather than 36 new uuid-named files — the update recipe in `adapters/README.md` documents that form, since the `upload_folder` it previously showed would have duplicated every record.

Verification on the new head: `ruff check` clean, `ruff format` clean, full suite 374 passed / 20 skipped, all 36 records still pass `every_eval_ever validate`, and a live run reproduces all 67 scores. The suite also passes on a clean `uv sync --locked` core-deps environment. Fork workflow approval is per-push, so the last two commits will need a maintainer to approve CI again.

Remaining sequencing, unchanged: land [eval-card-registry#54](https://github.com/evaleval/eval-card-registry/pull/54) first, then I refresh the pinned snapshot so the judge metric's `review_status` flips from `draft` to `reviewed` and re-push the records.

borgr:
CI closed the loop from my last comment: all four matrix legs pass on the current head `54c9bddd` (`core`/`full` × `locked`/`loose`). The three legs that showed red earlier were never executed — they sat queued without a runner and were cancelled with no step output, alongside another run in this repo at the same time; a re-run passed unchanged.

So from my side #160 is merge-ready: 9 files, `mergeable`/`CLEAN`, review findings either fixed with tests or answered on the thread.

One sequencing request, not a blocker: landing [eval-card-registry#54](https://github.com/evaleval/eval-card-registry/pull/54) first means no record ships citing a `draft` metric. Once it merges I refresh the pinned snapshot so the judge metric's `review_status` flips to `reviewed`, push that one-line change here, and update the 36 records on [EEE_datastore #179](https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/179) to match.

borgr:
Readability pass on `b39e01f`, no behaviour change: −241/+108 lines, all of it comments and prose.

The comments and docstrings had been carrying the reasoning behind each choice alongside the choice itself, which is the wrong place for it — a maintainer reading `_score_details` needs to know that the MCQ column is divided onto `[0, 1]` and the judge column is not, not why that was decided. Everything is now cut to what a reader or editor needs: the two leaderboard columns and their scope, where each field's value comes from, and what each guard protects. The `adapters/README.md` section drops the provenance essay and two inlined Python snippets, keeping the run command, a metric/scale/scope table, the snapshot refresh commands and the one submission caveat specific to this adapter; general datastore mechanics stay in the conversion skill where the other adapters find them.

Verified: `ruff check` clean, 374 passed / 20 skipped, and a live run reproduces records byte-identical to the ones on [EEE_datastore #179](https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/179) apart from the per-run uuid and timestamp.

Fork CI approval is per-push, so this head needs a maintainer to approve workflows once more. After that the only open sequencing item is [eval-card-registry#54](https://github.com/evaleval/eval-card-registry/pull/54): once it lands I refresh the pinned snapshot so the judge metric's `review_status` flips from `draft` to `reviewed`, and re-push the records.

PR #188, opened 2026-06-15
Convert BountyBench run logs through an adapter, with per-configuration record identity

borgr:
Converts BountyBench cybersecurity agent run logs into EEE aggregate records plus per-bounty instance-level sidecars.

The converter arrived as a standalone script under `utils/`. It is now an adapter under `every_eval_ever/adapters/bountybench/`, run like every other adapter and publishing through `publish_evaluation_logs`. Against `main` the branch is purely additive — it no longer introduces `utils/` at all.

```bash
uv run python -m every_eval_ever.adapters.bountybench.adapter \
  --logs-dir bountybench/logs/2026-03-26 \
  --output-dir /tmp/eee-bountybench/data/bountybench \
  --source-org 'Your Organization' \
  --bountybench-version 0.3.0
```

## What changed beyond the move

**One record per (model, workflow, configuration).** A BountyBench log carries no run id, so the configuration — the sha256 of `resources_used.model.config`, published as `config_fingerprint` — is the finest run boundary the source supports. Grouping per model let one aggregate report a temperature or token budget that only some of its bounties ran under. The three workflows are now separate evaluations (`bountybench.detect`, `.exploit`, `.patch`) and `evaluation_id` carries the workflow and the fingerprint, so two workflows of one model can no longer collide.

**Repeated attempts are rejected, not silently collapsed.** Two logs for one bounty under one configuration may be a retry or a second run, and they disagree on the score. `--attempt-policy` defaults to `reject`; `best` and `latest` collapse them and disclose the choice (`attempt_selection`, `n_attempts_total`, `n_bounties_with_multiple_attempts`, `GenerationArgs.max_attempts`).

**A sample is keyed on its input alone.** BountyBench scores by running the exploit or patch, so the only available "reference" was the observed outcome — which gave one bounty a different `sample_hash` per model. `input.reference` and `answer_attribution` are now empty, and the verdict stays in `Evaluation` where the verifier put it.

**Transcripts pair up.** BountyBench records each command twice, as the model's action and again as the executing resource's action carrying the output. Each execution now becomes exactly one `ToolCall`, and a `tool` message only cites the id of a call that was actually emitted. Output over 10,000 characters is cut with an in-band `[truncated N characters]` marker.

**Timestamps are host-independent.** Source stamps carry no UTC offset; `--source-timezone` (default `UTC`) says how to read them, so one log yields one `evaluation_id` everywhere instead of one per converting machine's local zone.

**Identity is refused rather than invented.** A model the source records without a provider prefix (`claude-code`) has no routable developer, so it is refused until `--model-developer` names one instead of being published under `unknown`. The datastore route is resolved at parse time, so an identity containing path components is one recorded failure rather than a write outside `--output-dir`.

**Incomplete input stops the run.** A log that cannot be parsed aborts before anything is written, because publishing the rest would report a success rate over a silently reduced denominator. `--allow-partial` converts the remainder, writes `adapter_reports/bountybench_failures.json` and exits non-zero. Logs that failed at startup ran no iteration and are recorded as exclusions, which do not fail the run.

## Verification

- `tests/test_bountybench_adapter.py` — 17 tests, one per behaviour above, including that a rejected batch and a parse failure write nothing at all.
- Full suite: 428 passed, 20 skipped. `ruff check` clean.
- `scripts/local_validate.py` over a published 3-record run and its sidecars: 6 passed, 0 errors, 0 warnings. The same validator reports 5 errors on the record already on the datastore — [both runs here](https://github.com/evaleval/every_eval_ever/pull/188#issuecomment-5239781030).

## One thing for a maintainer to decide

`data/BountyBench/` already exists on the datastore, from the original contributor's submission: one record for `anthropic/claude-opus-4-6`, schema 0.2.2, `detailed_evaluation_results.file_path` = `./<uuid>_samples.jsonl`. That record does not validate against the current schema — `file_path` must be the repository-relative `data/<collection>/<developer>/<model>/<uuid>_samples.jsonl`, which is one of the defects fixed here. It also has no workflow in its `evaluation_id`, a model-specific `evaluation_result_id`, and `developer: "Anthropic"` where the routed directory is `anthropic`.

This adapter writes the collection as `bountybench`, matching the `eval_library.name` already in that record and the 90 of 98 lowercase collections. So the datastore submission that follows this PR should delete `data/BountyBench/` in the same PR that adds `data/bountybench/`, rather than leaving the same benchmark under two collection names. Say the word if you would rather keep the capitalised directory and I will switch `COLLECTION`.


borgr:
The original contributor added it to the HF. I misclicked and merged instead of comment, so I manually committed it to here.

utils/BountyBench/convert_bountybench_logs.py
```diff
@@ -0,0 +1,782 @@
...
+    logger.info("Found %d raw logs", len(all_logs))
+
+    if not all_logs:
+        logger.error("No logs found")
+        sys.exit(1)
+
+    # Group by (model, workflow)
+    groups: dict[tuple[str, str], list[dict]] = {}
+    for log in all_logs:
+        key = (log["model"], log["workflow"])
```
mrshu: **[high] aggregation identity** — Separate runs and configurations can become one best-of score

This grouping key contains only model and workflow. Later deduplication prefers successful attempts across the whole group, while aggregate settings come from only the first retained log.

Configuration identity can partition logs before deduplication. When same-config runs have no trustworthy boundary, rejecting the ambiguous group avoids inventing one aggregate.

utils/BountyBench/convert_bountybench_logs.py
```diff
@@ -0,0 +1,782 @@
...
+                                        name="bash",
+                                        arguments={"command": action_command},
+                                    )
+                                ],
+                            )
+                        )
+                        turn_idx += 1
+                elif action_command or result_text:
+                    # Tool execution: command + result
+                    tool_call_id = f"tc_{turn_idx}"
```
mrshu: **[high] transcript linkage** — A normal result can reference a tool call that was never emitted

The model command receives one generated ID. The following result-only action receives a new ID, but no call with that second ID exists. Other action shapes can represent the command twice.

Each source execution can become one ordered call/result pair. A result needs the exact validated or generated ID of its emitted call.

utils/BountyBench/convert_bountybench_logs.py
```diff
@@ -0,0 +1,782 @@
...
+    phase_messages = log["phase_messages"]
+    system_msg = ""
+    if phase_messages:
+        for am in phase_messages[0].get("agent_messages", []):
+            if am.get("agent_id") == "system":
+                system_msg = am.get("message", "")
+                break
+
+    raw_input = system_msg or f"BountyBench {task_name} bounty {log['bounty_number']}"
+    reference = ["success"] if log["success"] else ["failure"]
```
mrshu: **[high] sample identity** — The model's outcome changes its ground truth and sample hash

This reference is selected from observed success, and the hash includes it. Two models therefore receive different sample identities for the same bounty when only their outcomes differ.

When BountyBench provides no real reference answer, an empty reference is safer than an invented one. Observed success can stay only in `Evaluation`.

utils/BountyBench/convert_bountybench_logs.py
```diff
@@ -0,0 +1,782 @@
...
+    file_uuid = str(uuid.uuid4())
+    model = logs[0]["model"]
+    model_id = model_id_from_bb(model)
+
+    if "/" in model_id:
+        dev_slug, model_name = model_id.split("/", 1)
+    else:
+        dev_slug, model_name = "unknown", model_id
+
+    dest_dir = output_dir / "BountyBench" / dev_slug / model_name
```
mrshu: **[high] output containment** — A model ID can escape `--output-dir`

The log-controlled model string is split and appended directly to the destination. Absolute components or `..` components can make the later directory and file writes land outside the requested output root.

Invalid components can be rejected rather than sanitized. A resolved-path containment check would protect the boundary before any filesystem operation.

utils/BountyBench/convert_bountybench_logs.py
```diff
@@ -0,0 +1,782 @@
...
+        dev = model_id.split("/")[0]
+        return dev.replace("-", " ").title()
+    return "Unknown"
+
+
+def convert_timestamp_to_unix(ts: str) -> str:
+    """Convert ISO timestamp to Unix epoch string."""
+    if not ts:
+        return str(datetime.now().timestamp())
+    try:
```
mrshu: **[medium] timestamps** — Naive source times depend on the converter host timezone

BountyBench writes timestamps without an offset. Calling `timestamp()` on that naive value applies the local host timezone, and the resulting epoch is part of the evaluation ID.

A documented source timezone, with an override for other provenance, can normalize the same log to one UTC instant on every host.

utils/BountyBench/convert_bountybench_logs.py
```diff
@@ -0,0 +1,782 @@
...
+    if log["duration_ms"] is not None:
+        performance = Performance(
+            latency_ms=log["duration_ms"],
+            generation_time_ms=log["query_time_ms"] or None,
+        )
+
+    # Answer attribution
+    answer_attribution = [
+        AnswerAttributionItem(
+            turn_idx=max(0, n_turns - 1),
```
mrshu: **[medium] answer attribution** — The cited extraction source is not serialized

The instance has no `workflow_metadata.workflow_summary.success` path, and the selected final turn does not contain this evaluator result. Consumers cannot resolve the mandatory attribution.

An empty attribution list avoids presenting workflow success as extracted model output. The result remains available in `Evaluation.score` and `is_correct`.

utils/BountyBench/convert_bountybench_logs.py
```diff
@@ -0,0 +1,782 @@
...
+
+
+def collect_logs(logs_dir: Path) -> list[dict[str, Any]]:
+    """Collect all JSON logs from a directory tree."""
+    results = []
+    for f in sorted(logs_dir.rglob("*.json")):
+        try:
+            results.append(parse_bounty_log(f))
+        except Exception as e:
+            logger.warning("Error parsing %s: %s", f, e)
```
mrshu: **[medium] partial input** — A malformed log can silently reduce the aggregate denominator

This path warns and continues. One valid success plus one malformed log can therefore become a successful 1/1 aggregate, with no completeness field in the output.

Default conversion can stop before writes when any candidate log fails. A partial mode would need explicit opt-in, omission details, and a nonzero final status.

utils/BountyBench/convert_bountybench_logs.py
```diff
@@ -0,0 +1,782 @@
...
+
+        if args.dry_run:
+            for log in usable:
+                status = "✅" if log["success"] else ("✓" if log["complete"] else "✗")
+                print(
+                    f"  {status} {log['bounty_id']}: {log['cve']} (tokens: {log['input_tokens'] + log['output_tokens']:,})"
+                )
+            continue
+
+        if convert_run(usable, args.output_dir, args.source_org):
```
mrshu: **[medium] group processing** — One empty group prevents later valid groups from converting

After startup failures are filtered, an empty group makes `convert_run()` return false and this branch exits immediately. Processing order therefore decides whether independent later groups produce output.

Empty groups can be recorded before the dry-run branch. Later groups can continue, with one final nonzero status after all groups are examined.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The converter has a clear structure, and several identity, transcript, and input-safety details would benefit from follow-up.
**Result:** Four high-priority correctness issues and four medium-priority robustness issues remain.
**Scope:** This PR converts BountyBench run logs into aggregate and instance-level EEE records.
**Checks:** 9 concerns checked; 9 confirmed, 0 ruled out, 0 set aside, 0 lowered in priority.
**Second check:** targeted, 2 reviewers; 8 kept, 8 fix paths clarified, 0 set aside, 1 lowered in priority.

## Earlier review comments

None.

## What I noticed

- **[high] aggregation identity** `utils/BountyBench/convert_bountybench_logs.py:748` — Separate runs and configurations are grouped together. Best-success deduplication can combine attempts from different runs, while the aggregate reports settings from only the first retained log. (`RAVF001`; inline)
- **[high] transcript linkage** `utils/BountyBench/convert_bountybench_logs.py:508` — Normal command results can reference a tool-call ID that was never emitted. Some action shapes can also emit the same execution twice and inflate the tool-call count. (`RAVF002`; inline)
- **[high] sample identity** `utils/BountyBench/convert_bountybench_logs.py:562` — `input.reference` and `sample_hash` depend on the model's observed success. The same bounty gets different identities when models have different outcomes. (`RAVF003`; inline)
- **[high] output containment** `utils/BountyBench/convert_bountybench_logs.py:676` — A log-controlled model ID can contain absolute or parent path components. The converter can then create output outside `--output-dir`. (`RAVF004`; inline)
- **[medium] timestamps** `utils/BountyBench/convert_bountybench_logs.py:272` — Naive BountyBench timestamps use the converter host's timezone. Identical logs produce different timestamps, evaluation IDs, and instance links on different hosts. (`RAVF005`; inline)
- **[medium] answer attribution** `utils/BountyBench/convert_bountybench_logs.py:602` — Attribution cites `workflow_metadata.workflow_summary.success`, but that path is absent from the instance. The selected turn also does not contain the value. (`RAVF006`; inline)
- **[medium] partial input** `utils/BountyBench/convert_bountybench_logs.py:205` — Parse failures are warned and skipped. A remaining subset can produce a successful aggregate with a reduced denominator. (`RAVF007`; inline)
- **[medium] group processing** `utils/BountyBench/convert_bountybench_logs.py:774` — One model/workflow group with only startup failures exits immediately. Later independent valid groups are never processed. (`RAVF009`; inline)

<details>

- **[low] result identifiers** — The aggregate ID is always named `detect`, and instances omit the preferred result ID. A workflow-aware shared ID would improve consistency. (`RAVF008`)

</details>

## Things to try

<details>
<summary>Things to try (8 items)</summary>

- **[high] aggregation identity** — Configuration identity can partition logs before deduplication. Ambiguous same-config runs need an explicit boundary or a clear rejection. (`RAVF001`)
- **[high] transcript linkage** — Each source execution can become one ordered call/result pair. Results need the exact ID of an emitted call. (`RAVF002`)
- **[high] sample identity** — An empty reference is safer than an invented one when BountyBench has no ground truth. The hash can use only stable task input and genuine reference data. (`RAVF003`)
- **[high] output containment** — Invalid path components can be rejected. The resolved destination can also be required to remain below the resolved output root. (`RAVF004`)
- **[medium] timestamps** — A documented source timezone, with an override for other provenance, can normalize naive timestamps to UTC. (`RAVF005`)
- **[medium] answer attribution** — An empty attribution list avoids presenting evaluator success as extracted model output. Success can stay in `Evaluation`. (`RAVF006`)
- **[medium] partial input** — Default conversion can stop before any write when a candidate log fails to parse. Partial mode needs explicit opt-in and visible incompleteness. (`RAVF007`)
- **[medium] group processing** — Empty usable groups can be recorded, later groups can continue, and the command can return nonzero after all groups are examined. (`RAVF009`)

</details>

<details>

- Target: PR #188 (`add-bountybench-converter`, 1 file, +782/-0)
- Mix: 3 codex-exec
- Focus: schema correctness, instance linkage, provenance, filesystem safety, maintainability, and constructive suggestion-oriented language
- Checks: concerns=9; confirmed=9/ruled-out=0/set-aside=0/lowered=0
- Second check: targeted; reviewers=2; kept=8/clarified=8/set-aside=0/lowered=1; approval changed no
- Set aside: 0 items

</details>

---

borgr:
All nine findings are addressed. The converter was rewritten as an adapter under `every_eval_ever/adapters/bountybench/`, so the line references below point at the new file. Each fix has a test named after the behaviour it protects.

**RAVF001 — aggregation identity (high).** Fixed. Logs are partitioned by `(model_id, workflow, config_fingerprint)`, where the fingerprint is the sha256 of `resources_used.model.config`; deduplication runs inside a partition, so it can no longer combine attempts across configurations, and the settings the aggregate reports hold for every bounty in it. Repeated attempts within one configuration are ambiguous with no run id, so `--attempt-policy reject` (the default) stops and names an example rather than picking; `best`/`latest` collapse them and publish `attempt_selection`, `n_attempts_total` and `n_bounties_with_multiple_attempts`. Tests: `test_differing_configurations_are_not_averaged_together`, `test_repeated_attempts_are_rejected_by_default` (asserts nothing was written), `test_best_attempt_policy_discloses_the_selection`.

**RAVF002 — transcript linkage (high).** Fixed. `build_messages_from_phases` mints `call_N` only when a call is actually emitted and carries the open call across the whole transcript, so a result either cites an emitted id or carries none. An executor echoing a still-unanswered command is recognised as the same execution, so one command is one call. Tests: `test_each_execution_yields_one_paired_tool_call` (ids `call_0`/`call_1`, results `[['call_0'], ['call_1']]`, contiguous `turn_idx`), `test_tool_result_without_a_call_has_no_dangling_reference`.

**RAVF003 — sample identity (high).** Fixed. `sample_hash` covers `{raw, reference}` with `reference` empty, so it is independent of the outcome, and `input.reference` is empty rather than holding the observed result. Test: `test_sample_hash_ignores_the_outcome` — two models with opposite outcomes on one bounty produce one hash.

**RAVF004 — output containment (high).** Fixed, and earlier than suggested. `resolve_model_identity` resolves the datastore route through `datastore_path_components` at parse time, so an identity with absolute or parent components is one recorded `SourceRecordFailure` before any staging, not an abort of the whole batch at publication. The routed name is also what lands in `ModelInfo.developer`, so the record and its directory agree. Test: `test_a_traversing_model_identity_cannot_escape_the_output_dir`.

**RAVF005 — timestamps (medium).** Fixed. `--source-timezone` (default `UTC`) declares how the naive stamps are read and is recorded in `source_metadata.additional_details`. Test: `test_naive_timestamps_are_read_in_the_declared_source_zone` — the same log shifts by exactly 4 h between `UTC` and `America/New_York`.

**RAVF006 — answer attribution (medium).** Fixed. `answer_attribution` is `[]`; the verdict stays in `Evaluation.score`/`is_correct`, where the verifier produced it. No turn is cited for a value no turn contains.

**RAVF007 — partial input (medium).** Fixed. A parse failure now aborts before anything is written, with a message naming the count and the first failure. `--allow-partial` is the explicit opt-in: it converts the rest, writes `adapter_reports/bountybench_failures.json` before publication so a publication failure cannot take the record of the omissions with it, and exits non-zero via `raise_if_incomplete()`. Tests: `test_unparseable_logs_stop_the_run_before_any_write`, `test_allow_partial_publishes_the_rest_and_reports_the_failure`.

**RAVF008 — result identifiers (low).** Fixed rather than deferred, because it was also a collision: with the workflow absent from `evaluation_id`, two workflows of one model at one timestamp produced the same id. `evaluation_result_id` is now the workflow slug (`detect`/`exploit`/`patch`), `evaluation_name` is `bountybench.<slug>`, `evaluation_id` carries both the slug and the fingerprint, and instances repeat `evaluation_result_id` so they link back. Test: `test_workflows_of_one_model_get_distinct_evaluation_ids`.

**RAVF009 — group processing (medium).** Fixed. Every partition is examined before anything is published; one with no usable bounty is recorded as an exclusion and the loop continues. Startup failures are exclusions rather than failures, so they do not fail the run, and only a run with nothing convertible at all exits non-zero. Tests: `test_startup_failures_are_excluded_without_failing_the_run`, `test_a_run_with_nothing_convertible_fails_loudly`.

Three defects not in the review, found while rewriting:

- `detailed_evaluation_results.file_path` was a bare `./<uuid>_samples.jsonl`. The schema requires the repository-relative `data/<collection>/<developer>/<model>/<uuid>_samples.jsonl`, and both the publisher and the merge gate check it. The record already on the datastore fails today's validator for exactly this reason — see the PR description.
- A model without a provider prefix was published under developer `unknown`, which `datastore_path_components` now rejects outright. It is refused until `--model-developer` names the publisher.
- `deployment_type` and `model_availability` were absent, so the merge gate's model-deployment check had nothing to read. They are now flags, defaulting to `externally_managed` (the hosted APIs BountyBench drives) and `unknown`.

Verification: 17 adapter tests, full suite 406 passed / 20 skipped, `ruff check` clean, and the merge gate with `run_semantic_checks=True` reports 0 errors and 0 warnings across a published 3-record run and its sidecars.

nelaturuharsha:
@borgr the last few issues need to be resolved - I'd suggest running it through the local validator to see if all is good. If you could post the same, then we can move to merge.

borgr:
@nelaturuharsha Done — `main` is merged in (`663e3d5`) and here is the run through `scripts/local_validate.py`, the standalone validator #238 added.

**The adapter's output.** Six logs in (three workflows, two models, one that failed at startup), three aggregates plus three sidecars out:

```
$ uv run python -m every_eval_ever.adapters.bountybench.adapter \
    --logs-dir logs --output-dir data/bountybench \
    --source-org 'EvalEval Coalition' --bountybench-version 0.3.0
INFO parsed 6 of 6 candidate logs
INFO anthropic/claude-opus-4 / PatchWorkflow   / 306a3f037b93: 1 log(s) -> 1 bount(ies), 1 usable
INFO openai/gpt-4o          / DetectWorkflow  / dc56f3739acf: 4 log(s) -> 4 bount(ies), 3 usable
INFO openai/gpt-4o          / ExploitWorkflow / dc56f3739acf: 1 log(s) -> 1 bount(ies), 1 usable
Wrote 3 BountyBench evaluation log(s).

$ uv run python scripts/local_validate.py data
data: validating 6 file(s) found recursively
╭─ Summary ─────────────────────────────────────────────────╮
│ 6 passed, 0 warning-only, 0 failed (0 errors, 0 warnings)  │
│ Merge-ready for the checks available locally.              │
╰───────────────────────────────────────────────────────────╯
exit 0
```

`every_eval_ever check-duplicates` on the same three aggregates: no duplicates. Suite on the merged branch: **428 passed, 20 skipped**, 17 of them the BountyBench tests; `ruff check` clean, and no file this PR touches appears in `ruff format --check` (23 files on `main` already do).

The logs are synthetic, because BountyBench does not publish its runs — `bountybench/bountybench` has no `logs/` tree. They are built to hit every identity rule at once: two workflows of one model that used to collide, a repeated bounty, a log that failed at startup, and a 20,000-character tool output that gets truncated.

**The one real artifact, through the same validator.** `data/BountyBench/` on the datastore, from the original submission — the record these rules exist to prevent:

```
$ uv run python scripts/local_validate.py data
data: validating 2 file(s) found recursively
╭─ Summary ──────────────────────────────────────────────────╮
│ 0 passed, 0 warning-only, 2 failed (5 errors, 0 warnings)   │
╰────────────────────────────────────────────────────────────╯

detailed_evaluation_results -> file_path [string_pattern_mismatch]
detailed_evaluation_results.file_path [semantic_rule_error]
  expected exactly 'data/BountyBench/anthropic/claude-opus-4-6/e24f7e2f-…_samples.jsonl'
  so the aggregate and samples share one UUID and folder,
  got './e24f7e2f-c66f-4db1-80c4-913df59e4c77_samples.jsonl'   (2 occurrences)
model_info.additional_details [semantic_rule_error]
  missing 'deployment_type' (expected self_deployed|externally_managed|unknown)
model_info.additional_details [semantic_rule_error]
  missing 'model_availability' (expected open_weights|closed_weights|unknown)
```

That is the difference in one screen: 5 errors on what is published, 0 on what this adapter writes.

On those two axes, since they are what the published record is missing: `deployment_type` is `externally_managed` and that is not a guess — every provider BountyBench can drive (`ServiceProvider`: anthropic, google, openai, together, xai, helm) is a remote API-key service, so the harness has no self-hosted path. `model_availability` stays `unknown`, because a log records which model was called and nothing about its weights. Both are `--deployment-type` / `--model-availability` flags for an operator whose own run knows better.

**On the nine review threads:** each is addressed [item by item here](https://github.com/evaleval/every_eval_ever/pull/188#issuecomment-5224439480), and all nine sit on `utils/BountyBench/convert_bountybench_logs.py` — a path this branch no longer adds at all, so GitHub holds them open against a file that is not in the diff. @mrshu, worth a glance if you have a minute; the line references in that comment point at the adapter that replaced it.

One decision is still a maintainer's rather than mine, from the end of the description: this adapter writes `data/bountybench/`, matching the `eval_library.name` already inside the published record and the 90 of 98 lowercase collections. So the datastore PR that follows should delete `data/BountyBench/` rather than leave one benchmark under two spellings. Say the word if you would rather keep the capitalised directory and I will switch `COLLECTION` instead.

every_eval_ever/adapters/bountybench/adapter.py
```diff
@@ -0,0 +1,924 @@
...
+BOUNTYBENCH_GITHUB = 'https://github.com/bountybench/bountybench'
+TOOL_OUTPUT_CHAR_LIMIT = 10_000
+WORKFLOW_SLUGS = {
+    'DetectWorkflow': 'detect',
+    'ExploitWorkflow': 'exploit',
+    'PatchWorkflow': 'patch',
+}
+
+
+def workflow_slug(workflow: str) -> str:
```
mrshu, line 96: **RAV-RUN2-R1-F004 [medium] workflow-identity** — The fallback maps distinct accepted labels such as `DetectWorkflow` and `Detect` to the same slug. Matching model, config, and timestamp values can then publish one evaluation identity.

The existing `WORKFLOW_SLUGS` map provides a clear boundary. Unknown names can enter the normal source-failure path until their semantics are mapped.

every_eval_ever/adapters/bountybench/adapter.py
```diff
@@ -0,0 +1,924 @@
...
+        if policy == 'latest':
+            return (not log['startup_failure'], log['start_time'])
+        return (
+            log['success'],
+            log['complete'],
+            not log['startup_failure'],
+            log['start_time'],
+        )
+
+    kept = [max(group, key=rank) for group in attempts.values()]
```
mrshu, line 344: **RAV-RUN2-R1-F005 [medium] attempt-selection** — The shared ranking makes `best` choose the latest final tie. It also makes `latest` prefer an earlier usable attempt over a later startup failure.

Policy-specific rankings over numeric `start_epoch` would align selection with the README and avoid lexical timestamp ordering.

every_eval_ever/adapters/bountybench/adapter.py
```diff
@@ -0,0 +1,924 @@
...
+            model_id,
+            workflow,
+            fingerprint,
+            len(group),
+            len(kept),
+            len(usable),
+        )
+        if not usable:
+            exclusions.append(
+                SourceRecordExclusion(
```
mrshu: **RAV-RUN2-R1-F006 [medium] source-accounting** — Attempt selection happens before startup exclusions. A startup row can disappear when a retry wins, while an all-startup group also adds a synthetic exclusion and can count one row twice.

A ledger derived once from the original group would preserve one accounting entry per source row.

every_eval_ever/adapters/bountybench/adapter.py
```diff
@@ -0,0 +1,924 @@
...
+            logger.info(
+                'partial conversion report: %s',
+                save_failure_report(
+                    result, default_failure_report_path(args.output_dir)
+                ),
+            )
+
+        published = publish_evaluation_logs(
+            logs,
+            args.output_dir.parent,
```
mrshu, line 871: **RAV-RUN2-R1-F008 [medium] output-path** — Publication receives `args.output_dir.parent` and appends `bountybench`. A request such as `/safe/job/chosen` therefore writes to `/safe/job/bountybench`.

Early validation of the documented collection-directory shape would keep staging, reports, and publication on the requested path.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This pass is comment-only; two confirmed high-priority identity and transcript issues remain.
**Result:** The rewrite fixes most earlier roots and follows shared publication patterns. Seven focused suggestions remain.
**Scope:** Convert BountyBench run logs into per-configuration aggregates and linked instance records.
**Checks:** 5 concerns checked; all 5 were confirmed.
**Second check:** 2 reviewers kept 7 findings, clarified their fix paths, and removed 1 workflow-preference item.

<details>
<summary>Earlier review comments (9 grouped roots)</summary>

- **Still present:** aggregation identity, transcript linkage, and a narrower group-processing failure remain relevant.
- **Fixed:** sample identity, output containment for model paths, timestamp normalization, answer attribution, strict partial-input handling, and result linkage.

</details>

## What I noticed

|---|---|---|---|---|
| RAV-RUN2-R1-F001 | high | aggregation identity | `every_eval_ever/adapters/bountybench/adapter.py:269` | The configuration fingerprint omits `max_iterations`, so different agent budgets can share one aggregate. |
| RAV-RUN2-R1-F002 | high | transcript linkage | `every_eval_ever/adapters/bountybench/adapter.py:525` | Consecutive model commands overwrite the one pending-call slot. Results detach from original calls and duplicate executions appear. |
| RAV-RUN2-R1-F004 | medium | workflow identity | `every_eval_ever/adapters/bountybench/adapter.py:97` | Unknown workflow labels can normalize to the same evaluation identity. (inline) |
| RAV-RUN2-R1-F005 | medium | attempt selection | `every_eval_ever/adapters/bountybench/adapter.py:306` | `best` selects the latest final tie, while `latest` means latest usable attempt rather than latest source attempt. (inline) |
| RAV-RUN2-R1-F006 | medium | source accounting | `every_eval_ever/adapters/bountybench/adapter.py:713` | Startup rows can disappear before selection, and all-startup groups can count one source row twice. (inline) |
| RAV-RUN2-R1-F007 | medium | group processing | `every_eval_ever/adapters/bountybench/adapter.py:735` | A missing required timestamp raises outside the expected source-data boundary and stops later valid groups. |
| RAV-RUN2-R1-F008 | medium | output path | `every_eval_ever/adapters/bountybench/adapter.py:822` | Publication replaces an arbitrary requested directory's final component with `bountybench`. (inline) |

## Things to try

<details>
<summary>Suggested next steps (7 items)</summary>

- **[high] aggregation identity** — A canonical identity payload can include `model_config` and the effective iteration budget. Missing and explicit defaults can normalize consistently. (`RAV-RUN2-R1-P001`)
- **[high] transcript linkage** — Exact-command FIFO queues can preserve source order and repeated commands. Unmatched executor actions can keep their current separate-call behavior. (`RAV-RUN2-R1-P002`)
- **[medium] workflow identity** — Only mapped workflow names can enter evaluation identity. Unknown names can use the normal source-failure path. (`RAV-RUN2-R1-P004`)
- **[medium] attempt selection** — Numeric `start_epoch` and explicit policy-specific ordering can match the documented rules. (`RAV-RUN2-R1-P005`)
- **[medium] source accounting** — Startup exclusions can derive once from the original group, before attempt ranking. The synthetic group exclusion can disappear. (`RAV-RUN2-R1-P006`)
- **[medium] group processing** — Required timestamps can validate during parsing. Only defined source-data errors can use partial-mode continuation. (`RAV-RUN2-R1-P007`)
- **[medium] output path** — Write-producing runs can validate that the requested collection directory ends in `bountybench` before staging or reports. (`RAV-RUN2-R1-P008`)

</details>

<details>

- Target: PR #188 (`add-bountybench-converter`, 4 files, +1416/-0) at `663e3d5836a819b67bec05ebdfa39e4f37ac55c9`
- Run ordinal: 2
- Mix: 3 codex-exec
- Finding counts: 0 critical, 2 high, 5 medium, 0 low, 0 nit
- Checks: concerns=5; confirmed=5
- Second check: targeted; reviewers=2; kept=7, clarified=7, removed=1; approval unchanged

</details>

---

borgr:
Thanks — all seven findings from the RUN2 review (`663e3d5`) are addressed on the branch. Two commits:

**`e97d6d1` — identity, transcript, and accounting**

- **F001 (aggregation identity)** — `config_fingerprint` now folds `max_iterations` into the payload alongside `model_config`, so runs under different agent budgets no longer average into one aggregate. `build_aggregate` drops the now-redundant mixed-limit note (a partition has one budget) and reports it via `eval_limits.message_limit`.
- **F002 (transcript linkage)** — `build_messages_from_phases` replaces the single pending-call slot with a FIFO of open calls; each result claims the matching call by command, then by arrival order. Consecutive model commands no longer overwrite one another, so no orphaned calls, detached results, or duplicated executions.
- **F005 (attempt selection)** — `best` now breaks ties toward the **earliest** start, as the README documents (was `max(start_time)` = latest), keyed on the numeric `start_epoch`. `latest` is documented as "the last-started attempt that produced a scorable run".
- **F006 (source accounting)** — startup failures are now recorded once per log from the whole group, before per-bounty selection can drop an attempt; the synthetic all-startup group exclusion that double-counted is gone.
- **F007 (group processing)** — a missing `start_time` is now a per-log parse failure (handled by `--allow-partial`), instead of raising inside `build_aggregate` and aborting later valid groups.
- **F008 (output path)** — `run()` rejects an `--output-dir` whose final component isn't the collection, rather than silently redirecting the write to `bountybench/`.

**`eecd8cd` — F004 (workflow identity)**

- An unmapped `workflow_name` was regex-slugged, so two distinct names could normalize to one slug and collide in `evaluation_id`. `parse_bounty_log` now rejects any workflow outside `Detect`/`Exploit`/`Patch` as a recorded failure; a new BountyBench workflow needs a deliberate `WORKFLOW_SLUGS` entry before its logs convert. The regex fallback (and the `re` import) are removed.

Each fix has a regression test in `tests/test_bountybench_adapter.py`; full suite green (470 passed, 17 skipped).

borgr:
Reviewed all 13 unresolved automated review threads against the current head. Every one is stale or already addressed: the eight on `utils/BountyBench/convert_bountybench_logs.py` predate the move to `every_eval_ever/adapters/bountybench/adapter.py`, and the remaining ones are handled in the current code (verifier-sourced verdict documented at `adapter.py:670-673`, publication and identity paths reworked). Resolving all 13.

borgr:
Brought up to date with main (`12ac73642`). One conflict, in the `adapters/README.md` table, where main's `paperswithcode` row met this branch's `bountybench` row — kept both, alphabetical.

Verified on the merge: `uv run pytest tests` 884 passed, 31 skipped; `uv run ruff check` clean. All four CI rows green on this head.

Labelled `auto-review` for a fresh pass. The open review-anvil items are the two high-priority ones on identity and transcript handling from the `663e3d5` round, which the `e97d6d1`/`f10d4d7`-era commits answered — worth re-reading against this head rather than the old diff.

borgr:
Scope note for the review pass.

The last report targeted \`663e3d583\` and named two confirmed high-priority items, identity and transcript. Both were answered after it, in \`e97d6d196\` (aggregation identity, transcript linkage, accounting) and \`eecd8cd0d\` (reject unknown workflows instead of deriving an identity). Against that head, \`adapters/bountybench/adapter.py\` moved 139 lines and the tests gained 141 — that is the delta. A raw diff against \`663e3d583\` is mostly main.

PR #190, opened 2026-06-20
[Adapter] AlpacaEval 1.0 and 2.0, consolidated with the merged adapter and resolved through the eval-card-registry

karthikchundi-commits:
## Review lane

**Needs a human.** Cross-package (adapter + shared resolver + vendored snapshot), well over ~1000 hand-written lines, and a material change in outcome: the same 324 leaderboard rows now publish different scores, ids, and provenance than what is on the datastore today.

Design agreed in: nothing up front — this began as a single adapter and grew a shared resolver once #230 turned out to need the same registry. That growth is the part to push back on if you want it split; the split point is `helpers/eval_card_registry.py` + `tools/refresh_eval_card_registry.py` + its tests, which is the only piece #230 depends on.

## Checklist

- [x] `validate` clean, **no warnings**, run at the final `data/<collection>/<dev>/<model>/` path — 324/324
- [x] every unconvertible source row in `adapter_reports/`, non-zero exit, no silent skip
- [x] offline unit tests added; full `pytest tests` green (513 passed / 20 skipped)
- [x] `ruff check` clean
- [ ] ids resolve in the registry — five do not yet; the alias PR is prepared and open as [eval-card-registry#55](https://github.com/evaleval/eval-card-registry/pull/55)
- [x] content spot-checked: no answer leak, not double-counted, `evaluation_id` stable across two independent runs

**Coverage:** 325 source rows → 324 records, 0 failed, 1 excluded (`NullModel`, itemized with its source row in `adapter_reports/`).

**Operator asked about policy calls?** Yes, twice: minting five new canonical registry ids (decided — report the gap, file #55, never auto-create), and replacing rather than overwriting 324 published records (decided — stage it as `refs/pr/178`, never commit to `main`).

Finishes this PR by consolidating it with the AlpacaEval adapter that landed on
`main` in the meantime, keeping whichever side is right field by field, fixing
what neither got right, and regenerating all 324 records to check the result.
@karthikchundi-commits' two commits are preserved underneath.

## Why the two implementations collided

This branch predates the repo reorganisation that moved adapters to
`every_eval_ever/converters/<source>/`, so as opened it *deletes* the adapter on
`main` and adds its own at the old `utils/alpaca_eval/` path. Both read the same
two leaderboard CSVs and emit the same four metrics per row, so this is the same
data converted twice — not new coverage. Both are now folded into
`every_eval_ever/converters/alpaca_eval/`, and `utils/alpaca_eval/` is removed so
only one adapter exists.

## Field-by-field

A = the adapter on `main`, B = this PR.

| # | Field / behaviour | A (`main`) | B (#190) | Kept | Result |
|---|---|---|---|---|---|
| 1 | Upstream ref | `…/main/…` (moving) | `…/main/…` (moving) | neither | pinned SHA `cd543a14`; `--ref` to move it |
| 2 | Score scale | ÷100 → proportion, bounds `[0,1]` | raw percent, bounds `[0,100]` | **B** | percent — the registry declares `win-rate` on `[0,100]`, which is the scale the CSV holds |
| 3 | Score rounding | `round(x, 6)` | `round(x, 4)` | neither | unrounded; rounding was discarding CSV precision |
| 4 | `metric_id` | `alpaca_eval.*` | absent | **A**'s intent | registry's `win-rate` where a canonical exists, `alpaca_eval.*` where it does not, plus `metric_name`/`metric_kind`/`metric_unit`/`metric_parameters` |
| 5 | SE method | `bootstrap` (**wrong**) | `analytic` (**right**) | **B** | `analytic` — upstream is `preferences.sem()` |
| 6 | SE scale | ÷100 | raw | **B** | raw percent, matching the score |
| 7 | v1 LC SE | reads absent column → `None` | hardcoded `None` | tie | column-driven; the v1 CSV has no `lc_standard_error` |
| 8 | `avg_length` unit | "tokens" (**wrong**) | "tokens" (**wrong**) | neither | characters — upstream is `output.str.len().mean()` |
| 9 | `avg_length` bound | `max_score=100000.0` (invented) | same | neither | unbounded (`"Infinity"` on the wire, per `MetricConfig.validate_bound_wire_type`), `metric_kind='length'`, not a quality score |
| 10 | `evaluation_name` | metric label (`Win Rate`) | metric key (`win_rate`) | neither | the registry benchmark id `alpacaeval-2-0` for v2, `alpaca_eval.v1` for v1 — it names the *evaluation*; metric identity lives in `metric_config` |
| 11 | `evaluation_id` | `alpaca_eval_v1/<id>/<now>` | `alpaca_eval/<id>/<now>` | neither | `<collection>/<slug>@<ref[:12]>` — stable across reruns, slug-keyed so `_concise` / `-best-of-16` variants don't collide |
| 12 | Model identity | private 37-entry substring map | shared `get_developer` | **B**'s reuse, **neither**'s accuracy | evidence ladder over upstream `models_configs/<slug>/configs.yaml`, rung recorded as `identity_source` |
| 13 | Unresolved identity | `id` = bare model name (77 rows) | `developer='unknown'` (137 rows) | neither | fail the row with provenance rather than guess — 0 rows actually fail |
| 14 | `submission_mode` | dropped | kept | **B** | kept, as `leaderboard_mode` |
| 15 | Source organisation | "Stanford CRFM / Tatsu Lab" (CRFM is unrelated) | "tatsu-lab" | **B** | "Tatsu Lab (Stanford University)" |
| 16 | `source_data` | `SourceDataUrl` → repo root | same | neither | `SourceDataHf` → `tatsu-lab/alpaca_eval`, split `eval`, 805 samples — the data actually evaluated |
| 17 | `eval_library.version` | leaderboard version `"1.0"`/`"2.0"` | hand-typed `"0.6"` | neither | `0.6.6`, parsed from upstream `__init__.py`; leaderboard version moved to `additional_details` |
| 18 | Judge | annotator name only | absent | **A** | judge `ModelInfo` + verbatim prompt + temperature + parser, via `llm_scoring` |
| 19 | Generation settings | absent | absent | tie | `GenerationConfig` per row from the upstream model config |
| 20 | Raw counts | dropped | dropped | tie | `n_wins`/`n_wins_base`/`n_draws`/`n_total` in `score_details.details` |
| 21 | `NullModel` | skipped | **published as a model** | **A** | excluded with a recorded reason (a constant-string gaming probe, not a model) |
| 22 | Rows missing `win_rate` | skipped silently | published if any other metric parses | **A** | recorded as a failure with row provenance |
| 23 | Errors | propagate | `except Exception: print(traceback)` | **A** | `SourceConversionResult`; failures/exclusions carry `CSV row N ('slug')` |
| 24 | Entry point | `cli.py convert alpaca_eval` | standalone `python -m`, hardcoded `data/` | **A** | CLI + `--version`, `--ref`, `--save_raw_json`, `--input_json`, `--no_registry_resolve`, `--registry_live` |
| 25 | Collection names | `alpaca_eval_v1` / `alpaca_eval_v2` | `alpaca_eval` / `alpaca_eval_2` | **A** | `alpaca_eval_v1` / `alpaca_eval_v2`, matching what is already published |
| 26 | Tests | 352 lines, mocks `requests.get` | deletes the test file | **A** | 52 tests over in-memory snapshots, no network |
| 27 | Unused imports | — | `ConfidenceInterval` imported, never used | **A** | — |

Net: B wins on 5 points (2, 5, 6, 14, 15, plus reusing the shared developer
helper), A on 10, and on 9 points neither was right.

## Canonical ids come from the eval-card-registry

Like the PapersWithCode adapter, organization, metric and benchmark ids are
resolved against the [eval-card-registry](https://evaleval-entity-registry.hf.space)
rather than invented here, so this source speaks the same vocabulary as the rest
of EEE. The resolver is `helpers/eval_card_registry.py` — repo-shared, not
adapter-private, because the validator needs the same vocabulary (see *Shared
resolver* below).

Measured over the 324 records:

| Entity | Resolved | Result |
|---|---|---|
| organization → `model_info.developer` | 233 / 324 (231 snapshot + 2 alias), all `reviewed` | 50 of 82 HF id prefixes have a canonical org (32 prefixes / 91 records have none); 12 prefixes differ from the developer they resolve to — either the registry records the namespace for that org (`meta-llama`→`meta` ×9, `Qwen`→`alibaba` ×7, `zai-org`→`zai` ×2), or it normalises the spelling (`HuggingFaceH4`→`huggingface` ×6, `WizardLM(Team)`→`wizardlm` ×7, `NousResearch`→`nous-research`, `baichuan-inc`→`baichuan`, `deepseek-ai`→`deepseek`, `Nanbeige`→`nanbeige`), or a repo moved namespace while its publisher did not (`ff670`→`OpenBuddy` ×2, `dphn`→`cognitivecomputations`) |
| metric | `win_rate` → `win-rate` (reviewed), 324 results | the other three keep `alpaca_eval.*` — no canonical exists |
| benchmark | v2 → `alpacaeval-2-0` (reviewed), 222 records | v1 keeps `alpaca_eval.v1` — no canonical exists |
| harness | none | `alpaca_eval` has no canonical entry |

`win-rate`'s declared bounds are `[0, 100]`, which is what settles row 2 above:
the scale is the registry's call, not the adapter's.

**`developer` and `id` are deliberately different identities.** The registry
records the organization and the HuggingFace namespace it publishes under as
separate facts (`meta`/`meta-llama`, `alibaba`/`qwen`, `zai`/`zai-org`), so
`model_info.id` keeps the namespace — that is the repo id that resolves — and
`developer` carries the canonical org id. The datastore folder comes from the id
prefix, so paths stay where a reader would look for them.

**Resolution reads a vendored snapshot, not `POST /resolve`.** That endpoint
defaults to `mode="resolve"`, which auto-creates a draft canonical for anything
it cannot place; bulk-resolving a leaderboard through it would add hundreds of
drafts to a shared registry as a side effect of a read-only conversion, and the
registry is already 18157 draft models to 5298 reviewed ones. So
`helpers/data/eval_card_registry.json` vendors the read-only list endpoints,
refreshed by `tools/refresh_eval_card_registry.py` (GET-only, with `--check` for
review), with provenance in `_meta` — the PapersWithCode pattern.
`--registry_live` opts into a
live check anyway and sends `mode="exact"`, the one mode that creates nothing;
a test asserts the request body, and a live failure degrades provenance rather
than failing the run. It asks about the five recorded gaps too, not only
spellings the snapshot has never seen, so a canonical minted after the last
refresh is picked up without waiting for one — 37 lookups per run rather than 32,
each spelling asked once. Measured today, all five still answer `no_match`, so no
published record differs between a live and an offline run. `--no_registry_resolve` turns resolution off entirely and
marks the records `registry_disabled` rather than letting them pass as
source-derived.

## Identity, measured over the same 324 rows

| | distinct developers | rows with no developer | bare `model_id` (no org) | agrees with merged |
|---|---|---|---|---|
| A (`main`) | 20 | 77 | 77 | 72 / 324 |
| B (#190) | 17 | 137 | 0 (137 under `unknown/`) | 56 / 324 |
| merged | 79 | 0 | 0 | 324 / 324 |

The ladder resolves **227/227** upstream model configs. Rungs used, per record:
`upstream_model_name` 134, `hf_model_link` 70, `vendor_api` 48, `github_link` 44,
`vendor_site` 13, `hf_link_org` 9, `name_pattern` 6.

Representative corrections (published → this branch):

```
stanford/alpaca-7b              -> tatsu-lab/alpaca-7b-wdiff
CohereForAI/cohere              -> cohere/command-nightly
mistralai/Mistral-7B-ReMax-v0.1 -> ziniuli/Mistral-7B-ReMax-v0.1
chatglm2-6b                     -> zai-org/chatglm2-6b
causallm-14b                    -> CausalLM/14B
openai/gpt4_1106_preview        -> openai/gpt-4-1106-preview
openai/aligner-2b_gpt-4-…       -> aligninc/aligner-2b_gpt-4-…
```

## Judgement calls worth reviewing

1. **Repo-id casing.** `completions_kwargs.model_name` is hand-typed and
   sometimes miscased (`01-ai/Yi-34b-Chat`, ten `baai/Infinity-Instruct-…`,
   `FreedomIntelligence/PlatoLM-7b` — none of those repos exist). A reference
   link is a URL that resolves, so where a link names the same repo up to case,
   the link's spelling wins; applied snapshot-wide, so a `-best-of-16` row that
   links only its *reward* model can still borrow the served model's own entry.
   Fixes 12 rows. Corroborated two ways: HF's `api/models/<id>` canonical-id
   endpoint agrees on all 12, and the datastore already spells them `BAAI` (14
   records) and `01-ai` (39). Result: 205 distinct ids over 324 records, **zero
   case-split ids**, and zero casing conflicts among the 126 linked repos.
2. **A colon disqualifies a repo id.** Upstream serves Llama 2 70B as
   `replicate/llama70b-v2-chat:e951f18…`; adopting that would publish Replicate
   as the developer. Such rows fall through to the link rungs and keep the served
   reference in `upstream_model_name`.
3. **Best-of-n rows must not adopt their reward model.** `ultralm-13b-best-of-16`
   links `openbmb/UltraRM-13b`: the org is trusted, the repo name is not.
4. **`gpt4_gamed` is kept, `NullModel` is excluded.** Both come from the
   leaderboard-gaming study, but `gpt4_gamed` is a real model with a real win
   rate whereas `NullModel` emits a constant string. Confirm this split is what
   you want.
5. **`evaluation_timestamp` is left unset.** The CSVs carry no per-row date, and
   the snapshot date would misdate 2023 submissions.
6. **`meta` vs `meta-llama` is now settled, not papered over.** Both spellings
   appear in the datastore (`meta` 333 / `meta-llama` 311, `alibaba` 309 / `Qwen`
   88, `deepseek` 359 / `deepseek-ai` 29 across 885 developer dirs). The registry
   resolves this properly, and it also avoids the trap that blocked a blanket
   fix: `helpers.get_developer` maps `LLM-Opt` → `meta` on a substring hit on
   `opt`, whereas the registry correctly declines to resolve it.
7. **Repo ids are published under the name HuggingFace serves today, not the one
   the 2023 config typed.** The leaderboards predate several namespace renames.
   The old id still answers via a redirect, so nothing *looks* broken — but the
   datastore's per-publisher folder comes from the id namespace
   (`helpers/io.py:datastore_path_components`), so a stale id silently fails to
   join with records other sources already publish under the current one.
   `data/hf_canonical_ids.json` maps referenced → current, swept GET-only by
   `refresh_hf_canonical_ids.py` (`--check` for review). Of 135 published repo
   ids: 124 confirmed, **9 renamed**, 11 left alone because `401` conflates
   *gated* with *nonexistent*. The 9 move 13 records:

   ```
   OpenBuddy/openbuddy-falcon-40b-v9-bf16  -> ff670/openbuddy-falcon-40b-v9-bf16
   THUDM/chatglm2-6b                       -> zai-org/chatglm2-6b
   WizardLM/WizardLM-13B-1.0               -> WizardLMTeam/WizardLM-13B-V1.0
   WizardLM/WizardLM-13B-V1.2              -> WizardLMTeam/WizardLM-13B-V1.2
   WizardLM/WizardLM-70B-V1.0              -> WizardLMTeam/WizardLM-70B-V1.0
   cognitivecomputations/dolphin-2.2.1-…   -> dphn/dolphin-2.2.1-mistral-7b
   meta-llama/Meta-Llama-3.1-{8B,70B,405B}-Instruct -> meta-llama/Llama-3.1-…
   ```

   Three things to review here. **(a)** A rename is applied only to rungs whose
   id *is* a repo id — `hf_model_link` or the `upstream_model_name` upstream
   recorded. The other rungs *construct* an id, and a constructed id that
   happens to redirect is not evidence: HF redirects `cohere/command-nightly` to
   `CohereLabs/Command-nightly`, but that row was served by Cohere's API under a
   rolling alias, so adopting the repo id would contradict its own
   `model_availability`. **(b)** `developer` deliberately does *not* follow the
   new namespace. A redirect cannot distinguish an org renaming itself from a
   repo transferred to someone else — `ff670` is an individual's namespace, and
   the publisher is still OpenBuddy. Same `id`/`developer` duality as
   `meta-llama` published by `meta`. **(c)** A rename is not extrapolated:
   `WizardLM/WizardLM-13B-V1.1` and `WizardLMTeam/WizardLM-13B-V1.1` *both*
   answer `401`, so that row keeps the id the source typed rather than a guessed
   namespace. **(d)** The map is rebuilt from scratch each run, so only `200` and
   `401` count as answers: a `429` or `5xx` for one id would delete the renames
   above, and such a sweep leaves the file untouched, names the ids and exits
   non-zero — `--check` included, since otherwise it would advise the very
   refresh that does the deleting. Note also that the map corrects the whole id, not just the
   namespace — `WizardLM-13B-1.0` gains the `V` the real repo has.
8. **Uncertainty and count cells are strict; provenance counts are not.** A
   populated cell that cannot be read is not the same fact as an absent one, so a
   standard error that is not a usable number (or is negative — `StandardError`
   has no `ge=0` bound) and an `n_total` that is not a positive whole count fail
   their own row with a named reason, rather than publishing a score with no
   uncertainty at all or quoting the dataset's 805 as a denominator the row never
   claimed. 69 of the 325 rows do carry a denominator other than 805 (795–805), so
   the fallback is only for a leaderboard with no such column. `n_wins`,
   `n_wins_base` and `n_draws` go the other way and keep their source text: they
   are provenance and are stringified either way, so parsing them could only drop
   a value it failed to read. Live, all six cells are clean on all 325 rows — this
   closes a hole rather than repairing observed loss.
9. **File names stay `uuid4()`; `evaluation_id` is what is reproducible.**
   Reruns produce identical records, paths and ids, but a fresh file name, so a
   rerun is not byte-identical and this PR no longer claims it is. Making the
   name content-derived is a repo-wide convention change, not an adapter choice:
   `helpers/io.py:require_uuid4` rejects anything that is not a v4 UUID, so the
   correct deterministic construct (UUIDv5) cannot be published today, and
   hashing a digest into v4 bits would produce a UUID that lies about its
   version.

## Verification of the regenerated data

All 324 records regenerated from the pinned snapshot:

- `validate` → **all 324 pass**; `check-duplicates` → **no duplicates**.
- Two independent full runs agree on **all 324 records, every field, and every
  output path**, with two exceptions that are repo conventions rather than
  adapter behaviour: the `uuid4()` file name and `retrieved_timestamp` (one value
  per leaderboard — when the run fetched). A live fetch reproduces the saved
  snapshot byte-identically, and `--input_json` replays it with no network.
- **Numerics exact.** Over 1210 results: every score equals its CSV cell, every
  score is inside its declared bounds, every SE equals the CSV SE, no SE is
  invented where the column is absent, and per-metric counts equal the non-empty
  CSV cell counts exactly (v1: win_rate 102, lc 58, discrete 60, avg_length 102;
  v2: 222 each).
- **Identity.** 205 distinct ids over 324 records; 0 unknown developers, 0
  `unknown/` path components (published has 77), 0 case-split ids. 300 folders
  for 324 records — 12 folders hold 2 records and 6 hold 3, all legitimate
  variant axes (`_concise`, `-best-of-16`, `_gamed`, `-togetherai`) kept distinct
  by `evaluation_id`. Ladder rungs, per record: `upstream_model_name` 134,
  `hf_model_link` 70, `vendor_api` 48, `github_link` 44, `vendor_site` 13,
  `hf_link_org` 9, `name_pattern` 6.
- **The rename step is a pure relabelling.** Against a run with the map disabled:
  exactly 13 records change `model_info.id`, no record changes anything else, 13
  folders move 1:1 with no merge or split, and none of the 9 new ids collides
  with an id another row already publishes. `developer` is unchanged on all 13.
- **Against the 324 published records** (read-only diff, joined on leaderboard
  slug): same 324 rows, no additions or removals. Every score equals published ×
  the scale change (×100 for the three win rates, unchanged for `avg_length`),
  within published's 6-dp rounding; all 371 SE values likewise, with `method`
  corrected `bootstrap` → `analytic` in every case. Bounds move `[0,1]` → `[0,100]`
  for the win rates and `100000.0` → unbounded for `avg_length`. The rest of the
  delta is identity and provenance: 248 model ids and 226 developers corrected,
  1210 results gain `metric_id`, every result gains judge/prompt/generation
  provenance, every record gains a reproducible `evaluation_id`, and the 77
  `unknown/` records gain real developers.
- Republishing therefore **replaces** rather than overwrites: of 324 folders, 76
  are kept, 248 retired and 224 new. The `alpaca_eval_v1`/`alpaca_eval_v2`
  collections contain exactly these 324 files and no `_samples.jsonl`, so the
  swap is self-contained.
- **Datastore replacement is staged as a PR**, not committed to `main`:
  https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/178
  (`refs/pr/178`) holds exactly these 324 records and deletes the 324 stale
  paths that are on `main`; `main` itself is untouched until someone merges it.
  Refreshed against this branch on 2026-08-07: the 324 records were rewritten in
  place at their existing PR paths — same ids, same paths, content only — and all
  324 re-validate clean at those paths, 0 errors and 0 warnings.
- Tests: **513 passed / 20 skipped** on the full suite (`uv run pytest tests`,
  what CI runs) — 99 adapter tests plus 20 for the shared resolver. `ruff check`
  clean.

## Registry-side follow-ups → filed as a PR to the registry

Minting a canonical is a lasting namespace decision, so the adapter reports gaps
instead of creating them, and the gaps are now a reviewable PR rather than a
paragraph here: **evaleval/eval-card-registry#55** adds the five entries this
source needs — benchmark `alpacaeval-1-0`, harness `alpaca-eval`, and metrics
`length-controlled-win-rate`, `discrete-win-rate`, `average-response-length` —
additively, with the seed dry-run at 817 passed, the same as unmodified `main`.

`gaps()` in this PR therefore still reports those five, and stops once #55 lands
*and* a maintainer syncs the Space: the live `/resolve` API updates on that sync,
not on merge, so the vendored snapshot is refreshed and the data regenerated as a
follow-up rather than pre-emptively.

Also reported on #55 as read-only findings, because they need org curation this
PR should not do:

- `baichuan-inc/baichuan-13b-chat` is a miscased reviewed entry (the repo is
  `baichuan-inc/Baichuan-13B-Chat`).
- The **generated** org list carries two reviewed rows for one uploader wherever
  HuggingFace renamed a namespace — `wizardlm` / `WizardLMTeam` and
  `cognitivecomputations` / `dphn`. Both pairs surface in this data, as a record
  whose id prefix is the new namespace and whose `developer` is the old org. The
  curated list already handles the same class correctly with `hf_org`
  (`zai`/`zai-org` for the THUDM rename, `meta`/`meta-llama`), so the fix shape
  is known; what blocks it is the seed loader collision described on #55.

Two findings from an earlier pass are **withdrawn**: `llm-stats-alpacaeval-2`
being a draft duplicate of `alpacaeval-2-0`, and `lmsys/vicuna-13b` being stale.
Both came from the live `/resolve` API, which lags the seed — in `seed/` the
first is already an alias of `alpacaeval-2-0` and the second does not exist. The
gap list above was re-derived from `seed/`, not from the API, for that reason.

## Shared resolver: this PR now owns it, #230 drops its copy

Three copies of the same registry vocabulary existed across open work: this
adapter's, plus `helpers/org_registry.py` and a second snapshot on the unmerged
`validator-developer-slug-drift` branch (**#230**). They are collapsed here into
one module, one snapshot and one refresh tool:

```
converters/alpaca_eval/registry.py                 -> helpers/eval_card_registry.py
converters/alpaca_eval/data/registry_snapshot.json -> helpers/data/eval_card_registry.json
converters/alpaca_eval/refresh_registry_snapshot.py-> tools/refresh_eval_card_registry.py
```

The move adopts #230's *stricter* alias rules wholesale (confirmed only; the
target must be a listed canonical id; an identity wins, including another org's;
one org per normalized spelling, ambiguity dropped) plus its `second_name_of`
read-side helper. Adopting them costs nothing measurable: strict and lenient
derivations select **exactly the same 99 aliases** — 0 dropped, 0 added, 0
retargeted. Identity derivation is also made order-independent, which changes 4
collision keys (`DeepAuto-AI`/`deepautoai` and three truncated `Gemini-…` ids),
none of them an AlpacaEval publisher.

Because that module does not exist on `main`, **#230 cannot be rebased onto it
until this merges.** Ordering: this PR first; #230 then deletes
`helpers/org_registry.py`, `helpers/data/org_registry.json`,
`tools/refresh_org_registry.py` and `tests/test_org_registry.py`, and imports
`second_name_of` from `helpers.eval_card_registry`.

## Nothing here merges alone

This PR, the datastore PR
([EEE_datastore#178](https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/178)),
registry PR #55 and #230 are one change split across four repos. All four stay
open until the set is reviewed together, so the code, the vocabulary it resolves
against and the data it produces land in an order that never leaves the datastore
pointing at ids the registry has not got.


utils/alpaca_eval/adapter.py
```diff
@@ -0,0 +1,337 @@
...
+    "/weighted_alpaca_eval_gpt4_turbo_leaderboard.csv"
+)
+OUTPUT_DIR_V2 = "data/alpaca_eval_2"
+
+SOURCE_DATA_V2 = SourceDataUrl(
+    dataset_name="alpaca_eval_2",
+    source_type="url",
+    url=["https://github.com/tatsu-lab/alpaca_eval"],
+)
+
```
mrshu: **[high] data semantics** — This module maps the same two CSVs as the established AlpacaEval converter, but it changes score units, metric IDs, versions, evaluation IDs, and source metadata. The same leaderboard rows can therefore produce incompatible records.

The requested `utils` entry point can stay lightweight and delegate row conversion to the established converter. It can then map shared logs to the datastore directories needed here.

utils/alpaca_eval/adapter.py
```diff
@@ -0,0 +1,337 @@
...
+        count += 1
+    return count
+
+
+def fetch_alpaca_eval_2(retrieved_timestamp: str) -> int:
+    print("Fetching AlpacaEval 2.0 leaderboard...")
+    rows = fetch_csv(ALPACA_EVAL_2_URL)
+    source_metadata = make_source_metadata(
+        source_name="AlpacaEval 2.0 Leaderboard",
+        organization_name="tatsu-lab",
```
mrshu: **[medium] remote input** — The current V2 CSV contains a populated `NullModel` placeholder, and this loop writes it as a normal model.

The established converter already filters this placeholder. Delegating conversion to that shared path would keep the rule in one place.

utils/alpaca_eval/adapter.py
```diff
@@ -0,0 +1,337 @@
...
+    return count
+
+
+def main():
+    retrieved_timestamp = str(time.time())
+    print("=" * 60)
+    print("AlpacaEval leaderboard adapter")
+    print("=" * 60)
+
+    try:
```
mrshu: **[medium] error handling** — A handled fetch, conversion, or write failure reaches the final `Done!` message and exits with status 0. Automation can treat a partial dataset as complete.

If this entry point keeps independent work, both outcomes can feed a final status that `__main__` propagates. Delegation may remove this duplicate path.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The requested adapter entry point is useful, but
its conversion semantics currently conflict with the existing AlpacaEval
converter.
**Result:** Three findings remain. The second check narrowed the consolidation
path so the requested `utils` entry point can stay lightweight.
**Scope:** This PR adds a manually run AlpacaEval 1.0 and 2.0 datastore adapter.
**Checks:** 3 concerns checked; 3 confirmed.
**Second check:** 2 Codex reviewers checked 3 findings; 3 kept and 3 fix paths
clarified.

## Earlier review comments

None.

## What I noticed

|---|---|---|---|---|
| RAVF001 | high | data semantics | `utils/alpaca_eval/adapter.py:69` | This adds a second mapping for the same CSVs. It uses different score units, metric IDs, versions, evaluation IDs, and source metadata. |
| RAVF002 | medium | remote input | `utils/alpaca_eval/adapter.py:284` | The current V2 CSV contains `NullModel`, and this loop writes it as a normal model. |
| RAVF003 | medium | error handling | `utils/alpaca_eval/adapter.py:315` | Handled fetch or write failures print an error and `Done!`, then the module exits with status 0. |

## Things to try

- **[high] data semantics** — The `utils` module can remain as the requested
  manual entry point while it delegates row conversion to the established
  converter. One implementation can then define scores, identifiers, metadata,
  and placeholder filtering. (`RAVF001`, `RAVF002`)
- **[medium] output layout** — The thin entry point can map the shared logs to
  the datastore directories required by this workflow. `submission_mode` can
  be preserved there if it is part of that contract. (`RAVF001`)
- **[medium] error handling** — If this entry point retains independent work,
  it can collect both outcomes and propagate a final nonzero status through
  `__main__`. Delegation may remove the duplicate failure path instead.
  (`RAVF003`)

<details>

- Target: PR #190 (`add-alpaca-eval-adapter`, 2 files, +337/-0)
- Mix: 3 codex-exec
- Finding counts: 0 critical, 1 high, 2 medium, 0 low, 0 nit
- Checks: concerns=3; confirmed=3; ruled-out=0; set-aside=0; lowered=0
- Second check: targeted; reviewers=2; kept=3; clarified=3; set-aside=0;

</details>

---

every_eval_ever/converters/alpaca_eval/adapter.py
```diff
@@ -1,265 +1,733 @@
...
+    then records the spelling the source used.
+    """
+    engine = (
+        InferenceEngine(name=resolved.inference_engine)
+        if resolved.inference_engine
+        else None
+    )
+    return ModelInfo(
+        name=resolved.slug,
+        id=resolved.model_id,
```
mrshu, line 687: **RAV-RUN2-R1-F001 [medium] model-identity** — Developer IDs use the registry, but `model_info.id` uses an adapter-private identity ladder. Canonical model aliases cannot affect this join key.

Only model IDs emitted by the pinned snapshot need vendored resolution. The source-derived ID can remain a marked fallback.

every_eval_ever/converters/alpaca_eval/adapter.py
```diff
@@ -304,97 +787,137 @@ def fetch_leaderboard_result(
...
                 continue
-            model_id = f'{developer}/{model_name}'
 
-            evaluation_id = (
-                f'{benchmark_key}/{model_id}/{retrieved_ts}'
+            # Stable and rerun-idempotent: the leaderboard row plus the pinned
+            # upstream revision. Same-model variants (`_concise`, `_verbose`,
+            # `-best-of-16`, `_gamed`) legitimately share a model id, so the
+            # slug is what keeps their records distinct.
+            evaluation_id = f'{cfg["collection"]}/{slug}@{ref[:12]}'
```
mrshu, line 976: **RAV-RUN2-R1-F002 [medium] reproducibility** — Branch and tag refs are used in stable IDs and provenance URLs without commit resolution. Two network runs can fetch different input under the same identity.

Network-mode refs can resolve once to a commit SHA. Offline replay can keep the stored immutable ref without network access.

every_eval_ever/cli.py
```diff
@@ -370,12 +370,31 @@ def _cmd_convert_helm(args: argparse.Namespace) -> int:
...
         LEADERBOARDS,
         AlpacaEvalAdapter,
     )
-
-    adapter = AlpacaEvalAdapter()
+    from every_eval_ever.converters.alpaca_eval.upstream import UpstreamSnapshot
+    from every_eval_ever.helpers.eval_card_registry import Registry, gaps
+
+    snapshot = None
+    if args.input_json:
```
mrshu, line 392: **RAV-RUN2-R1-F003 [medium] cli** — The module entry point builds the old argument namespace. This handler now reads missing fields and fails before conversion.

The module entry point can use the shared parser, with one offline entry-point test covering its defaults.

every_eval_ever/converters/alpaca_eval/adapter.py
```diff
@@ -1,265 +1,733 @@
...
+#: string submitted to show the judge can be gamed, not a model.
+_NULL_MODEL_RE = re.compile(r'null.?model', re.IGNORECASE)
+
+
+# ---------------------------------------------------------------------------
+# Row helpers
+# ---------------------------------------------------------------------------
+
+
+def _to_float(value: Any) -> Optional[float]:
```
mrshu, line 307: **RAV-RUN2-R1-F005 [medium] input-validation** — Populated invalid `win_rate` values bypass row-level failure handling. They can omit the headline metric or abort strict JSON publication.

Numeric fields can validate as finite values under their metric rules before result construction. Invalid rows can enter `SourceConversionResult`.

every_eval_ever/converters/alpaca_eval/adapter.py
```diff
@@ -1,265 +1,733 @@
...
+    if not config:
+        return None
+    kwargs = identity_mod.completions_kwargs(config)
+    max_tokens = None
+    for key in _MAX_TOKEN_KEYS:
+        candidate = _to_int(kwargs.get(key))
+        if candidate is not None and candidate >= 1:
+            max_tokens = candidate
+            break
+    prompt_template = config.get('prompt_template')
```
mrshu: **RAV-RUN2-R1-F007 [medium] generation-provenance** — `GenerationArgs.prompt_template` contains an upstream path instead of prompt content. Offline records cannot reconstruct the generation prompt.

New snapshots can store pinned prompt content. Legacy snapshots can leave this typed value unset and mark the missing content.

every_eval_ever/helpers/eval_card_registry.py
```diff
@@ -0,0 +1,387 @@
...
+
+        The snapshot records a ``None`` entry for a query the registry has no
+        canonical for, which is a different fact from "not in the snapshot": the
+        first is a known gap this adapter has already reported, the second means
+        the snapshot needs a refresh. Both fall back, only the second tries the
+        network.
+        """
+        if not self.enabled:
+            return _unresolved(name, entity_type, 'registry_disabled')
+        entries = _snapshot()[section]
```
mrshu, line 251: **RAV-RUN2-R1-F009 [medium] registry-resolution** — Live mode skips keys whose vendored value is null. A known gap cannot discover a new registry entry until the snapshot refreshes.

Live mode can query present-null entries while offline behavior remains unchanged.

every_eval_ever/tools/refresh_eval_card_registry.py
```diff
@@ -0,0 +1,344 @@
...
+    ``model_info.developer``.
+
+    Canonical ids are read in sorted order and win over namespaces, so a
+    spelling two organizations both answer to resolves the same way on every
+    refresh. Four such collisions exist today, all between a canonical id and
+    its own differently-punctuated twin (``DeepAuto-AI``/``deepautoai``), plus
+    ``ai21``'s ``ai21labs`` namespace against the ``ai21-labs`` canonical id.
+    """
+    identities: Dict[str, str] = {}
+    for record in sorted(orgs, key=lambda record: str(record.get('id', ''))):
```
mrshu: **RAV-RUN2-R1-F011 [medium] registry-identity** — Normalized organization collisions can rewrite an exact canonical ID to another canonical organization.

Exact canonical IDs can win first. Normalized spellings can resolve only when one organization owns them.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — Seven confirmed medium concerns remain in the consolidated adapter.
**Result:** The nine earlier concerns are fixed. This head has seven medium concerns and four low suggestions.
**Scope:** This PR consolidates AlpacaEval conversion with registry resolution, provenance, and publication.
**Checks:** 11 concerns checked; 10 confirmed and 1 narrowed.
**Second check:** targeted, 2 reviewers; 11 kept, 9 clarified or lowered, 0 set aside, 0 removed.

## Earlier review comments

- All nine earlier comments are fixed or stale. The duplicate adapter is gone, `NullModel` is excluded, and incomplete conversion exits nonzero.

## What I noticed

- **RAV-RUN2-R1-F001 [medium] model-identity** `every_eval_ever/converters/alpaca_eval/adapter.py:602` — Developer IDs use the registry, but `model_info.id` uses an adapter-private identity ladder. Canonical model aliases cannot affect the join key. (inline)
- **RAV-RUN2-R1-F002 [medium] reproducibility** `every_eval_ever/converters/alpaca_eval/adapter.py:838` — Explicit branch or tag refs are not resolved before use in stable IDs and provenance URLs. Two network runs can fetch different input under the same identity. (inline)
- **RAV-RUN2-R1-F003 [medium] cli** `every_eval_ever/cli.py:383` — The module entry point builds the old argument namespace. The changed handler reads missing fields and fails before conversion. (inline)
- **RAV-RUN2-R1-F005 [medium] input-validation** `every_eval_ever/converters/alpaca_eval/adapter.py:302` — Populated invalid `win_rate` values bypass row-level failure handling. They can omit the headline metric or abort strict JSON publication. (inline)
- **RAV-RUN2-R1-F007 [medium] generation-provenance** `every_eval_ever/converters/alpaca_eval/adapter.py:643` — `GenerationArgs.prompt_template` contains an upstream path instead of prompt content. Offline records cannot reconstruct the generation prompt. (inline)
- **RAV-RUN2-R1-F009 [medium] registry-resolution** `every_eval_ever/helpers/eval_card_registry.py:245` — Live mode skips keys whose vendored value is null. A known gap cannot discover a new registry entry until the snapshot refreshes. (inline)
- **RAV-RUN2-R1-F011 [medium] registry-identity** `every_eval_ever/tools/refresh_eval_card_registry.py:151` — Normalized organization collisions can rewrite an exact canonical ID to another canonical organization. (inline)

<details>
<summary>Non-blocking low findings (4 items)</summary>

- **RAV-RUN2-R1-F004 [low] output-lifecycle** `every_eval_ever/cli.py:624` — The fixed default smoke directory keeps UUID-named output from earlier runs.
- **RAV-RUN2-R1-F006 [low] metric-semantics** `every_eval_ever/converters/alpaca_eval/adapter.py:438` — The primary description always says 805 judged instructions, even when `n_total` is smaller.
- **RAV-RUN2-R1-F008 [low] source-metadata** `every_eval_ever/converters/alpaca_eval/adapter.py:893` — Every row is marked third-party, including a model developed by the evaluating organization.
- **RAV-RUN2-R1-F010 [low] registry-provenance** `every_eval_ever/helpers/eval_card_registry.py:267` — A prior live error causes later successful exact misses to report `registry_unavailable`.

</details>

<details>
<summary>Things to try (11 items)</summary>

- **[medium] model-identity** — Only model IDs emitted by this pinned snapshot can receive vendored registry resolution. The source ID can remain a marked fallback. (`RAV-RUN2-R1-P001`; covers `RAV-RUN2-R1-F001`)
- **[medium] reproducibility** — Network-mode refs can resolve once to a commit SHA. Offline replay can keep the stored immutable ref without network access. (`RAV-RUN2-R1-P002`; covers `RAV-RUN2-R1-F002`)
- **[medium] cli** — The module entry point can use the shared top-level parser and one offline entry-point test. (`RAV-RUN2-R1-P003`; covers `RAV-RUN2-R1-F003`)
- **[low] output-lifecycle** — A temporary-directory context can own and remove default smoke output while preserving explicit output paths. (`RAV-RUN2-R1-P004`; covers `RAV-RUN2-R1-F004`)
- **[medium] input-validation** — Populated numeric cells can validate as finite values within their metric rules before result construction. Invalid rows can enter `SourceConversionResult`. (`RAV-RUN2-R1-P005`; covers `RAV-RUN2-R1-F005`)
- **[low] metric-semantics** — The primary description can use the validated row denominator. (`RAV-RUN2-R1-P006`; covers `RAV-RUN2-R1-F006`)
- **[medium] generation-provenance** — New snapshots can store pinned prompt content. Legacy snapshots can leave the typed value unset and mark missing content. (`RAV-RUN2-R1-P007`; covers `RAV-RUN2-R1-F007`)
- **[low] source-metadata** — Canonical evaluator and model-developer organizations can determine the relationship. (`RAV-RUN2-R1-P008`; covers `RAV-RUN2-R1-F008`)
- **[medium] registry-resolution** — Live mode can query present-null entries while offline behavior stays unchanged. (`RAV-RUN2-R1-P009`; covers `RAV-RUN2-R1-F009`)
- **[low] registry-provenance** — Each cached live lookup can carry a hit, miss, or error result. Aggregate errors can remain separate. (`RAV-RUN2-R1-P010`; covers `RAV-RUN2-R1-F010`)
- **[medium] registry-identity** — Exact canonical IDs can win first. Normalized spellings can resolve only when one organization owns them. (`RAV-RUN2-R1-P011`; covers `RAV-RUN2-R1-F011`)

</details>

<details>

- Target: PR #190 at `d081d835d21d582a0e056e0e0fcbd77344fc8abb` (15 files, +6557/-459)
- Run ordinal: 2
- Mix: 3 codex-exec
- Earlier review comments: 9 comments; all fixed or stale
- Finding counts: 0 critical, 0 high, 7 medium, 4 low, 0 nit
- Checks: concerns=11; confirmed=10/ruled-out=0/set-aside=0/narrowed=1
- Second check: targeted; reviewers=2; kept=11/clarified-or-lowered=9/set-aside=0/removed=0; approval changed no

</details>

---

every_eval_ever/converters/alpaca_eval/adapter.py
borgr: Disagreeing, with numbers.

I measured this rather than reasoning about it. Of the 205 distinct real model ids on the two leaderboards, resolving through the registry's model vocabulary gives: **91 exact hits, 1 alias hit, 113 unresolved (55%)**. Of the 91 hits, only 8 differ from the HuggingFace id — and **all 8 are worse**:

- `anthropic/claude-2.1` → `claude-21`
- `openai/gpt-3.5-turbo-0301` → `gpt-3-5-turbo-0301`
- `baichuan-inc/Baichuan-13B-Chat` → lowercased

Those are punctuation normalization producing ids that resolve nowhere. 20 of the 91 are still `draft`. So adopting the model vocabulary would replace 205 working repo ids with 91 ids of which 8 are broken, and drop the other 113 on the floor.

The planned model-registry change (`MODEL_REGISTRY_PROPOSAL.md`) rules this out at the source: it permits case-only normalization and explicitly prohibits punctuation normalization, org substitution, and family inference. The 8 bad cases above are exactly what it forbids.

Three further costs, for the record:

1. 232 of 252 aggregate files embed the publisher spelling in `evaluation_id`, so this is a data rewrite, not a `git mv`.
2. Canonical model ids can be renamed or merged upstream. An identifier that moves is not an identifier — which is why the proposal freezes canonical ids and appends changed upstream ids as identifiers instead.
3. Vendoring costs 900 KB for a minimal id map, 17 MB for the full one, against 23,455 models / 39,755 aliases of which 19,236 are confirmed. 30 of 37 collision groups are case-only, which the registry says nothing about.

What the adapter does instead, and I think this is right: **organizations** resolve through the registry (1,166 canonical, all `reviewed`), `model_info.id` keeps the HuggingFace repo id that actually resolves, and `raw_model_id` / `raw_model_namespace` record the source's own spelling under the names the proposal reserves for them — so if a registry ever does own `model_info.id`, re-resolution needs no refetch.

One thing your finding did surface: resolution was not saying how *strongly* it matched. `5609eef73` splits that out — `snapshot_identifier` means a spelling the registry records matched case-insensitively, `snapshot_normalized` means punctuation was discarded too. Measured across both live leaderboards, all 39 formerly-`snapshot` organizations are recorded identifiers and the normalized tiers decide nothing; a test fails if that changes.

every_eval_ever/helpers/eval_card_registry.py
borgr: Disagreeing — this one is deliberate, and it's documented at the `_keyed` docstring.

A present-`null` entry and an absent key are different facts. A `null` is the refresh tool recording "I asked the registry about this exact spelling and it had no canonical" — a known gap, already reported by the CLI on every run and by `gaps()`. An absent key means the snapshot has never been asked, i.e. it needs a refresh. Only the second is worth a network call.

The reason not to retry the first: `POST /resolve` in `mode="exact"` is the only side-effect-free mode, and it answers from the same data the list endpoints served. Asking it about a spelling the list endpoints just told us has no canonical returns the same `null` — so the cost is one HTTP round trip per row per known gap, for a result we already have. With 4 metric gaps × 324 rows that's ~1,300 pointless requests against a shared HF Space.

Minting the canonical is the actual fix, and it isn't the adapter's to make: a new canonical id is a lasting namespace decision and belongs in a PR to the registry. That's what `gaps()` exists to surface — the current list is `metric:length_controlled_winrate`, `metric:discrete_win_rate`, `metric:avg_length`, `benchmark:AlpacaEval 1.0`, `harness:alpaca_eval`.

If the concern is a *stale* snapshot silently hiding new registry entries, that's covered from the other side: `refresh_eval_card_registry --check` fails when the committed snapshot no longer matches the registry, so drift is caught in review rather than papered over per-lookup at conversion time.

I did take the related point from F010 — a clean miss and an outage were both collapsing to one strategy. Fixed in `c8a33c8d2`.

every_eval_ever/converters/alpaca_eval/adapter.py
```diff
@@ -1,265 +1,844 @@
...
+        metric_id=metric.metric_id,
+        metric_name=spec.metric_name,
+        metric_kind=spec.metric_kind,
+        metric_unit=spec.metric_unit,
+        metric_parameters={
+            'baseline_model': cfg['baseline'],
+            'annotator': cfg['annotator'],
+            'leaderboard_version': cfg['leaderboard_version'],
+        },
+        lower_is_better=metric.lower_is_better,
```
mrshu, line 525: **RAV-RUN3-R1-F001 [medium] metric-semantics** — `avg_length` is described as neutral but receives `lower_is_better=False`. Consumers can read that Boolean as higher-is-better.

The adapter approach is consistent with the current schema; the remaining gap is a shared representation for neutral measures.

every_eval_ever/converters/alpaca_eval/adapter.py
```diff
@@ -1,265 +1,844 @@
...
+                )
+                if standard_error is not None
+                else None
+            ),
+            num_samples=n_total,
         )
+    details = {f'source_{spec.column}': value}
+    if spec.judge_scored:
+        details.update(
+            {column: _to_int(_cell(row, column)) for column in _COUNT_COLUMNS}
```
mrshu: **RAV-RUN3-R1-F002 [medium] input-validation** — Populated malformed uncertainty and count cells become absent values here. The later denominator fallback also turns invalid or zero `n_total` into 805.

Separate rules for uncertainty, counts, and positive denominators would keep bad rows in the failure report. Empty optional cells can remain allowed.

every_eval_ever/converters/alpaca_eval/refresh_hf_canonical_ids.py
```diff
@@ -0,0 +1,224 @@
...
+    """Return ``(status, current id)`` for one repo id.
+
+    The id is ``None`` whenever HuggingFace does not hand one back — a gated or
+    nonexistent repo (both ``401``), or anything else non-200.
+    """
+    response = requests.get(
+        HF_API_URL.format(repo_id=repo_id), timeout=REQUEST_TIMEOUT
+    )
+    if response.status_code != 200:
+        return response.status_code, None
```
mrshu, line 115: **RAV-RUN3-R1-F005 [medium] registry-refresh** — Every non-200 response drops the current ID from the rebuilt map. A transient 429 or server error can therefore remove a confirmed rename when the file is overwritten.

An incomplete sweep can leave the existing output unchanged and return a nonzero result. Expected 401 responses can keep their current handling.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This pass is comment-only; five medium suggestions remain on identity, data semantics, and refresh safety.
**Result:** The consolidation is much stronger at this head. Most earlier concerns are fixed, and the remaining items have focused fix paths.
**Scope:** Consolidate AlpacaEval 1.0 and 2.0 conversion, identity, registry resolution, CLI support, and pinned source data.
**Checks:** 5 concerns checked; 4 confirmed and 1 lowered in priority.
**Second check:** 2 reviewers checked 8 findings; 5 kept, 2 clarified or lowered, and 1 removed.

<details>
<summary>Earlier review comments (11 grouped items)</summary>

- **Still open:** `RAV-RUN2-R1-F001` model identity and `RAV-RUN2-R1-F009` live lookup for known registry gaps remain relevant.
- **Fixed:** moving refs, module CLI parsing, score-cell validation, prompt content, evaluator relationship, temporary output, denominator text, live-error provenance, and organization collisions.
- **Fixed:** the old duplicate converter, `NullModel` publication, partial-success exit, and conflicting output layout are no longer present.

</details>

## What I noticed

|---|---|---|---|---|
| RAV-RUN2-R1-F001 | medium | model identity | `every_eval_ever/converters/alpaca_eval/adapter.py:662` | `model_info.id` still bypasses shared model aliases, so the same model can keep separate join keys across sources. |
| RAV-RUN2-R1-F009 | medium | registry resolution | `every_eval_ever/helpers/eval_card_registry.py:249` | Live mode still skips present-null snapshot entries, so it cannot discover a new canonical until refresh. |
| RAV-RUN3-R1-F001 | medium | metric semantics | `every_eval_ever/converters/alpaca_eval/adapter.py:504` | `avg_length` is described as neutral but receives a binary direction value that signals higher is better. (inline) |
| RAV-RUN3-R1-F002 | medium | input validation | `every_eval_ever/converters/alpaca_eval/adapter.py:535` | Populated invalid uncertainty and count cells can disappear, and invalid denominators can become 805. (inline) |
| RAV-RUN3-R1-F005 | medium | registry refresh | `every_eval_ever/converters/alpaca_eval/refresh_hf_canonical_ids.py:104` | A transient non-200 response can remove a confirmed rename when the refreshed map replaces the prior file. (inline) |

<details>
<summary>Non-blocking low findings (2 items)</summary>

- **RAV-RUN3-R1-F003 [low] CLI help** — `--input-json` says it is offline, but `--registry-live` remains an explicit network exception. The help text can name that exception.
- **RAV-RUN3-R1-F004 [low] registry reporting** — The CLI prints registry status before live lookups. A final status line would show the actual query and error totals.

</details>

## Things to try

<details>
<summary>Suggested next steps (6 items)</summary>

- **[medium] model identity** — Exact aliases for model IDs emitted by supported converters can stay small. Source IDs and identity evidence can remain in provenance. (`RAV-RUN3-R1-P001`; covers `RAV-RUN2-R1-F001`)
- **[medium] metric semantics** — The neutral-metric representation needs a shared contract before this adapter changes its published coverage. (`RAV-RUN3-R1-P002`; covers `RAV-RUN3-R1-F001`)
- **[medium] input validation** — Populated uncertainty and count cells can use their own numeric rules. Invalid rows can then enter the failure report. (`RAV-RUN3-R1-P003`; covers `RAV-RUN3-R1-F002`)
- **[medium] registry resolution** — Present-null entries can use the existing cached live lookup while offline behavior stays unchanged. (`RAV-RUN3-R1-P004`; covers `RAV-RUN2-R1-F009`)
- **[low] registry reporting** — One final status line before publication checks can replace the initial zero-count line. (`RAV-RUN3-R1-P005`; covers `RAV-RUN3-R1-F004`)
- **[medium] registry refresh** — Transient failures can stop an incomplete destructive refresh. Existing output can stay unchanged when the sweep is incomplete. (`RAV-RUN3-R1-P006`; covers `RAV-RUN3-R1-F005`)

</details>

- **RAV-RUN3-R1-F001 [medium] metric semantics** — The issue is confirmed, but the specific representation stays open because the current schema has no neutral direction.

<details>

- Target: PR #190 (`add-alpaca-eval-adapter`, 16 files, +8729/-514) at `ed5b7dcbe7a2cd5755338c02431782f03a12be82`
- Run ordinal: 3
- Mix: 3 codex-exec
- Earlier review comments: 11 grouped items; 2 still present and 9 fixed or stale
- Finding counts: 0 critical, 0 high, 5 medium, 2 low, 0 nit
- Checks: concerns=5; confirmed=4, lowered=1; none set aside by verification
- Second check: targeted; reviewers=2; kept=5, clarified/lowered=2, removed=1; approval unchanged
- Set aside: 1 fix choice; reason=shared neutral-metric contract

</details>

---

borgr:
Four of the five are fixed in `d03822052`, plus both lows. `RAV-RUN3-R1-F001` I am
leaving where your own set-aside puts it. One of the two carried-over items —
`RAV-RUN2-R1-F009` — I argued against last round on a cost figure that was simply
wrong, so it is fixed here too.

Re-verified live on this head: a fresh fetch and an `--input_json` replay both give
**324 logs from 325 source rows** (0 failures, 1 exclusion — `NullModel`, itemized),
`validate` **all 324 passed** at the `data/<collection>/<dev>/<model>/` path, and every
record is field-for-field identical to the previous head apart from the `uuid4()` file
name and `retrieved_timestamp`. `pytest tests` **491 passed / 20 skipped** (this
adapter's file 83 → 99 tests, the shared resolver's 19 → 20); `ruff check` clean.

### RAV-RUN3-R1-F002 · populated invalid cells could disappear

Confirmed, and there were four distinct losses in that code, not one:

1. A malformed `standard_error` / `lc_standard_error` parsed to `None`, so the score
   published with **no uncertainty at all** — indistinguishable from a source that
   never reported one.
2. A **negative** SE published as given. `StandardError.value` has no `ge=0` bound in
   `eval_types.py`, so nothing downstream would have caught it.
3. An unusable or non-positive `n_total` parsed to `None`, so `num_samples` vanished
   *and* `samples = _to_int(...) or HF_DATASET_SAMPLES` quoted **805** in the metric
   description — a denominator the row never claimed. This is the same class as the
   denominator text you had me fix in run 2, one layer down.
4. A malformed count parsed to `None` and `_stringify` then dropped the key, so the
   value disappeared with nothing recorded anywhere.

1–3 are now row failures with named reasons (`n_total is not a positive count:
'805.5'`), so they land in `SourceConversionResult.failures` and the report, per P003.
Two details are judgement calls rather than mechanics:

- **The counts went the other way.** `n_wins`, `n_wins_base` and `n_draws` keep their
  **source text** instead of being validated. They are provenance, no score depends on
  them, and they are stringified either way — so parsing them could only drop a value
  it failed to read, which is defect 4 again in a different costume. Same divergence I
  took on #197's `_optional_number`. `'767 (est.)'` now survives to `details`.
- **The SE is checked with its metric, not standalone.** An SE whose score cell is
  absent is never published, so it cannot be a reason to drop a row; a test pins that.
  And `0.0` is a legitimate SE — the v2 baseline ties with itself on every instruction
  — so the check is `error is None or error < 0`, not a truthiness test.

The fallback to 805 is now correct by precondition rather than by luck: a populated
`n_total` that is not a positive count fails above, so only an *absent* column falls
back. That matters here — 69 of the 325 rows carry a denominator other than 805
(795–805).

Live, all six cells are clean on all 325 rows (0 unreadable, 0 negative, 0
non-integer, 0 zero), so this closes a hole rather than repairing observed loss, and no
published record changes. Six tests, including the two negatives above.

### RAV-RUN3-R1-F005 · a transient non-200 could delete a confirmed rename

Confirmed and worse than a lost update: the file is rebuilt from scratch each run, so a
single `429` for `WizardLM/WizardLM-13B-V1.2` would have dropped its rename and
silently returned the adapter to publishing a stale id — with the write reporting
success.

Statuses now split by whether they are an *answer*: `200` carries the id, `401` is
HuggingFace's answer for gated-or-nonexistent (both, indistinguishably). Anything else
means the sweep never got an answer, and `build_map` returns those ids with their
statuses; `main` then leaves the file exactly as it is, names up to ten of them on
stderr and exits 1.

`--check` is gated by the same list, deliberately — an incomplete sweep cannot tell a
deleted rename from an unasked question, so without the gate `--check` would print
"differs from HuggingFace; rerun without `--check`" and advise the refresh that does
the deleting. Three tests: a `503` writes nothing and exits 1; a `429` under `--check`
leaves the committed bytes intact and never claims a difference; a `401` is still a
complete sweep and writes the rename. On live data only `{200: 124, 401: 11}` occur, so
the gate cannot get stuck on a healthy sweep.

### RAV-RUN2-R1-F009 · live mode skipped present-null entries — you were right

I disagreed last round on the grounds that re-asking would cost "~1,300 pointless
requests" (4 gaps × 324 rows). That was wrong: `_live_lookup` caches on
`(entity_type, raw_value)`, so each spelling is asked **once per run** regardless of
row count. The measured cost of your P004 is five extra requests — a full live run goes
from 32 lookups to 37. I built the argument on a number I had not measured, in a reply
that led with measurements everywhere else.

So it is implemented as P004 describes: a present-`null` entry now takes the same
cached live path as an absent key, and offline behaviour is unchanged — both are still
`no_canonical`, and nothing is contacted. A test pins both halves.

It also does not change the data today: I asked the live registry for all five gaps in
`mode="exact"` just now, and it answers `no_match` for every one, so a live run and an
offline run produce byte-identical records. What the fix buys is the case your finding
was actually about — once #55 lands and someone syncs the Space, `--registry_live`
picks the new canonicals up instead of waiting for a snapshot refresh.

The rest of that reply stands: minting a canonical is still a registry-side PR, and
`gaps()` still reports the five on every run.

### RAV-RUN3-R1-F001 · `avg_length` direction — taking your set-aside

Agreed on the diagnosis, and I am not changing it here, for the reason you gave:
*"the specific representation stays open because the current schema has no neutral
direction"*. `lower_is_better=False` is a two-valued field being asked a three-valued
question, and the honest options are all schema changes — an optional third state, or
dropping the field for neutral metrics and losing it for the two win rates that
genuinely have a direction. Either is a shared contract (P002), not an adapter call, and
guessing one here would make this adapter's coverage disagree with every other. What the
adapter can carry without a schema it already does — the metric description says
*"Reported for length-bias context, not as a quality score"* in words.

Happy to make the change in this PR the moment the contract exists — it is one field.

### RAV-RUN2-R1-F001 · model identity — still disagreeing, same numbers

Unchanged from [my measured reply last round](https://github.com/evaleval/every_eval_ever/pull/190#discussion_r3732608677),
and the short version so it is not just a link: of the 205 distinct model ids here, the
registry's model vocabulary gives 91 exact hits, 1 alias, and 113 unresolved (55%); of
the 91, the 8 that differ from the HuggingFace id are **all worse**
(`anthropic/claude-2.1` → `claude-21`, `openai/gpt-3.5-turbo-0301` →
`gpt-3-5-turbo-0301`, `baichuan-inc/Baichuan-13B-Chat` lowercased), and 20 are still
`draft`. `MODEL_REGISTRY_PROPOSAL.md` forbids exactly that punctuation normalization,
so adopting the vocabulary today would replace 205 resolving ids with 91 of which 8
resolve nowhere, and drop 113.

P001's "exact aliases … can stay small" is the right shape, but it needs those aliases
to exist first. What is here already matches it on the read side: organizations resolve
through the registry, `model_info.id` keeps the repo id that resolves, and
`raw_model_id` / `raw_model_namespace` carry the source's own spelling under the names
the proposal reserves — so if a registry does come to own `model_info.id`,
re-resolution needs no refetch of the source.

### RAV-RUN3-R1-F003, F004 · the two lows

`--input-json`'s help no longer claims offline: *"Nothing is fetched unless
`--registry_live` is also given."*

For F004 I kept the up-front line and **added** a final one rather than replacing it, as
P005 suggests — the initial line announces the configuration and the snapshot date
before a multi-minute fetch, and the `no canonical entry for:` list reads as its
continuation. It is the *counters* on it that were meaningless, and those are now
reported after conversion, live runs only: `eval-card-registry live lookups: 37
queries, 0 resolved`, with the sticky `live_error` appended when there was one. Two
tests, one asserting the line reports a non-zero query count where the up-front line
read `live_queries: 0`, one that the line is absent when live mode is off.

mrshu:
# ⚒️ review-anvil report

**Review decision:** COMMENT — This pass is comment-only; two medium identity and refresh suggestions remain.
**Result:** The latest changes resolve most earlier feedback. The remaining items have narrower, well-bounded fix paths.
**Scope:** Consolidate AlpacaEval conversion, identity, registry resolution, CLI support, and pinned source data.
**Checks:** 1 concern checked; it was confirmed.
**Second check:** 2 reviewers kept 2 material findings and clarified both fix paths.

<details>
<summary>Earlier review comments (19 grouped roots)</summary>

- **Still present:** `RAV-RUN2-R1-F001` model aliasing and `RAV-RUN3-R1-F005` 401 refresh behavior remain relevant.
- **Still present, low:** `RAV-RUN3-R1-F003` remains in the converter README only; CLI help now names the network exception.
- **Fixed:** duplicate conversion, score validation, ref pinning, prompt provenance, registry lookup, collision handling, status reporting, and prior CLI/output concerns.
- **Intentionally set aside:** neutral `avg_length` direction remains a shared schema decision.

</details>

## What I noticed

|---|---|---|---|---|
| RAV-RUN2-R1-F001 | medium | model identity | `every_eval_ever/converters/alpaca_eval/adapter.py:687` | Final model IDs still bypass shared exact aliases, so cross-source join keys can remain split. |
| RAV-RUN3-R1-F005 | medium | registry refresh | `every_eval_ever/converters/alpaca_eval/refresh_hf_canonical_ids.py:139` | A later 401 can remove a confirmed rename from the rebuilt map and restore a stale source ID. |

<details>
<summary>Non-blocking low and nit suggestions (2 items)</summary>

- **RAV-RUN3-R1-F003 [low] docs** — The README can state that `--registry_live` is the network exception to offline snapshot replay.
- **RAV-RUN4-R1-F001 [nit] CLI** — A nullable parser default would avoid treating one concrete temporary path as the omission marker.

</details>

## Things to try

- **[medium] model identity** — One exact, idempotent alias boundary can run after source-specific resolution. Source IDs and evidence can stay in provenance. (`RAV-RUN4-R1-P001`; covers `RAV-RUN2-R1-F001`)
- **[medium] registry refresh** — A 401 can retain only the prior mapping for IDs still emitted now. A valid 200 remains authoritative. (`RAV-RUN4-R1-P002`; covers `RAV-RUN3-R1-F005`)

<details>

- Target: PR #190 (`add-alpaca-eval-adapter`, 16 files, +9065/-514) at `1983a5f858a9111576dfcd10bd0bc65bd93028c0`
- Run ordinal: 4
- Mix: 3 codex-exec; 2 reviewed the exact head and 1 could not reach GitHub
- Finding counts: 0 critical, 0 high, 2 medium, 1 low, 1 nit
- Checks: concerns=1; confirmed=1
- Second check: targeted; reviewers=2; kept=2, clarified=3, removed=0; approval unchanged

</details>

---

karthikchundi-commits:
Pushed `e8c8d081a` for **RAV-RUN3-R1-F005** (registry refresh): `build_map` now takes the previously committed `renamed_repos` map and retains an entry when the current sweep only gets 401 for it, per P002 - a fresh 200 still always overrides whatever the prior map said. `main` now loads the committed file once and reuses it for both the retention pass and `--check`, instead of loading it twice.

Added two new cases in `test_alpaca_eval_adapter.py`: a 401 on an id with a prior confirmed rename retains it and is no longer counted `unverifiable`; and a 200 overriding a stale prior mapping. (The "401 with no prior mapping" case was already covered by the existing `test_a_gated_repo_is_still_a_complete_sweep`.) `pytest tests/test_alpaca_eval_adapter.py`: 101/101 passing (99 existing + 2 new). Full suite `pytest`: 510 passed, 5 pre-existing failures unrelated to this change (Windows path-separator comparisons in `test_vectara_hallucination_leaderboard_adapter.py`, `test_community_evals_converter.py`, `test_local_validate.py`, `test_skill_conversion.py` - none touch alpaca_eval or the refresh script). `ruff check` clean; `ruff format --diff` shows only pre-existing drift in code this PR doesn't touch, nothing near my changes.

**RAV-RUN2-R1-F001 (model identity) - not changed, and I don't think it should be guessed at here.** I looked for the "one exact, idempotent alias boundary" P001 describes and couldn't find a model-id alias table anywhere in the codebase to run it against - there's `org_aliases`/`org_alias_spellings` in `eval_card_registry.json` for *developer/organization* resolution (already wired in via `developer.canonical_id`), but nothing equivalent keyed on model id, and I couldn't find `MODEL_REGISTRY_PROPOSAL.md` in this repo to check what "exact" is meant to mean precisely. Given @borgr's measured numbers on the closely related broader-vocabulary approach (91 exact / 1 alias / 113 unresolved of 205, and 8 of those 91 resolving *worse* due to forbidden punctuation normalization), I'd rather not improvise a narrower version of that mechanism without the actual alias data or the proposal doc in hand - the risk of getting "exact" subtly wrong here seems higher than the value of a guess. This feels like the same shape as `RAV-RUN3-R1-F001`'s set-aside: happy to implement the alias boundary the moment there's a concrete table or doc to point it at.

karthikchundi-commits:
Following up — it's been a couple of weeks since `e8c8d081a`.

Current state:
- **RAV-RUN3-R1-F005** (registry refresh) is addressed in `e8c8d081a` with two new tests; all four CI checks green; PR is `MERGEABLE`.
- **RAV-RUN2-R1-F001** (model identity) is the only remaining item. My position from the last comment still stands: there's no model-id alias table or `MODEL_REGISTRY_PROPOSAL.md` in this repo to implement the "one exact, idempotent alias boundary" against, and given @borgr's measured numbers on the closely related broader approach I'd rather not improvise a narrower version blind. Happy to add it the moment there's a concrete table or proposal doc to point it at — same shape as the `RAV-RUN3-R1-F001` set-aside @borgr already accepted.

Is this in a mergeable state now, with F001 tracked as a follow-up? Or if F001 is a hard blocker, could you point me at the alias data or the proposal doc so I can implement it here? Happy either way — just want to know what's needed to move it forward.

borgr:
F001 is done. Pushed to this branch as `5e6a233`, all five checks green.

The alias data you asked for does exist in the eval-card-registry. It is an organization table rather than a model-id table. Every org record carries `hf_org`, the namespace that org publishes model repos under, and `GET /api/v1/orgs` already returns it to the refresh tool this PR adds. An org table is enough here because every id F001 flags is *constructed* — `get_model_id(slug, site_developer)` on the `vendor_site` rung, where `site_developer` came from a domain (`ai.meta.com` -> `meta`). A website names the company while repos live under a namespace, so `meta/llama-2-70b-chat-hf` joins with nothing and `meta-llama/Llama-2-70b-chat-hf` is the repo that exists.

What the commit adds:

- `org_hf_namespaces` in `refresh_eval_card_registry` derives canonical org id -> `hf_org`, recorded only where the registry declares one, so an org with no HuggingFace presence stays absent rather than mapping to itself.
- `hf_namespace_of` reads it off the vendored snapshot.
- `resolve_identity` consults it on the `vendor_site` rung and nowhere else, and the spelling the site gave moves to `model_id_as_referenced`.
- `hf_namespaces={}` switches the lookup off, mirroring the existing `hf_canonical={}` seam. Four tests cover the lift, the off switch, an org with no recorded namespace, and the confinement to `vendor_site`.

Measured effect. 14 of the 1141 organizations that declare a namespace declare one differing from their own id, so this rewrites very little. Across the 324 records it moves 2, both Llama-2 70B chat rows, and the rest are byte-identical. Two datastore directories relocate from `meta/` to `meta-llama/`. Casing (`llama-` against `Llama-`) is left alone on purpose, because `refresh_hf_canonical_ids` follows HuggingFace's redirect and vendors the spelling it lands on, so that sweep is where casing belongs.

Declining to improvise the broader version was the right call. Three other routes measured dead first:

- Substring matching against the registry's `seed/models` gives `higgs-llama-3-70b-v2` -> `meta-llama` (it is Boson AI) and `openpipe-moa-gpt-4-turbo-v1` -> `openai` (it is OpenPipe). `get_developer` fails the same way, by substring.
- Adopting a unique exact-name repo from HuggingFace search adopts nothing. Eleven namespaces publish a repo named `Llama-2-70b-chat-hf`.
- Borrowing a sibling config's link does not reach these rows. 126 configs link a repo, none of those names is ambiguous, and no config's slug is linked by a sibling under a different name. The 70B slug has no sibling match at all.

The `hf_org` route avoids all three because the mapping is curated per organization rather than matched.

Two things left, both outside this branch and neither blocking review here. eval-card-registry #55 mints the five canonicals this adapter still reports as `no_canonical` — metrics `length_controlled_winrate`, `discrete_win_rate` and `avg_length`, benchmark `AlpacaEval 1.0`, harness `alpaca_eval`. Merging #55 publishes to the Space automatically, and one snapshot re-refresh then fills all five. The HuggingFace datastore PR 178 regenerates once after that, picking up the five canonicals and the two moved directories together rather than twice.

PR #191, opened 2026-06-21
feat: add generalized Kaggle Community Benchmarks adapter

mrshu:
Kaggle publishes ~1,080 community benchmarks, each an `owner/slug` resource with a public leaderboard. Only `cohere-labs/global-mmlu-lite` has an adapter today. This converts all of them into `data/kaggle/`: one record per (benchmark, model), one result per task result within it.

```bash
# every published benchmark
uv run python -m every_eval_ever.adapters.kaggle.adapter \
  --all --output-dir data/kaggle

# one benchmark, output outside the repo
uv run python -m every_eval_ever.adapters.kaggle.adapter \
  --benchmark cohere-labs/global-mmlu-lite-korean --output-dir /tmp/eee-kaggle
```

## One index read, for every path

Both paths read Kaggle's unauthenticated `benchmarks.BenchmarkService/ListBenchmarks` RPC first (via the anonymous XSRF handshake the adapter performs for you). That RPC is the only place a benchmark's scoring configuration is published, so a `--benchmark` run and an `--all` run describe the same benchmark identically rather than the targeted path silently losing `lower_is_better`.

## Source and naming

The source is Kaggle, the platform that published the leaderboard — `source_name` `Kaggle Benchmarks`, `source_organization_name` `Kaggle` — with the uploader in `source_metadata.additional_details.benchmark_owner`. One `data/kaggle/` collection holds every benchmark.

A benchmark's overall score and its sub-benchmarks share a record, so they are named apart and a consumer summing a file cannot double-count:

| | `evaluation_name` | `source_data.dataset_name` | `metric_id` |
|---|---|---|---|
| aggregate | `kaggle.<benchmark>` | `<benchmark>` | `kaggle.<benchmark>.<metric>` |
| sub-benchmark | `kaggle.<benchmark>.<sub>` | `<sub>` | `kaggle.<benchmark>.<sub>.<metric>` |

The owner is a segment because a benchmark is an `owner/slug` resource and 15 slugs are published by more than one owner (`animalimagerecognition` by six), so the slug alone gave 17 `metric_id`s to 45 unrelated benchmarks; the corpus now has 0 shared `metric_id`s across 7,470 distinct ids.

`metric_name`/`metric_kind` name the metric (`aggregationType` `PERCENTAGE_PASSED` → `pass rate`/`pass_rate`), uniformly across a record; Kaggle's column title is kept as `metric_config.additional_details.kaggle_task_title`. `eval_library.name` is `unknown` — no harness ran these.

## What the records do and do not claim

Kaggle does not publish a metric's unit or scale, and its rendering hints do **not** imply one. Across the published corpus, values under `displayType=PERCENTAGES` / `aggregationType=PERCENTAGE_PASSED` run from **-180.29 to 74,453,100.0**, with only 68.5% in `[0, 1]`.

So numeric results carry **no** `metric_unit`, **no** `score_type` and **no** bounds; the raw `aggregationType`/`displayType`/task version travel in `metric_config.additional_details` so a later pass can normalize what this adapter refuses to guess. Boolean task results are genuinely pass/fail and are `binary` on `[0, 1]`. `sortOrder` *is* published (`DESCENDING` for all 1,080), is validated against `ASCENDING`/`DESCENDING` at construction, and is the only source of `lower_is_better`.

`confidenceInterval` is a symmetric half-width and `unevenConfidenceInterval` an asymmetric `{plus, minus}` pair (446 results); both are author-supplied and occasionally nonsensical (219 negative half-widths, 257 equal to the score), so a half-width that is not a non-negative finite number yields no uncertainty rather than bounds that do not bracket the score.

## Model identity

`modelVersionSlug` mixes the model with the effort it was run at, so `model_info.id` is resolved against the [eval-card-registry](https://github.com/evaleval/eval-card-registry) through a new shared `helpers/registry.py` resolver (extracted from `open_medical_llm`, which now uses it too), and the tier is reported as a generation setting:

| Kaggle slug | `model_info.id` | generation config |
|---|---|---|
| `claude-opus-4-6-default` | `anthropic/claude-opus-4.6` | `reasoning_effort: default` |
| `claude-sonnet-4-5-thinking-20250929` | `anthropic/claude-sonnet-4.5-20250929` | `reasoning: true`, `reasoning_effort: thinking` |
| `grok-4.20-0309-non-reasoning` | `xai/grok-4.20-non-reasoning` | `reasoning: false`, `reasoning_effort: non-reasoning` |
| `qwen3-235b-a22b-thinking-2507` | `Qwen/Qwen3-235B-A22B-Thinking-2507` | — |

The registry decides the identity; nothing is inferred from a model's name. The slug as published is offered first, and only when the registry has no canonical id for it is the effort-stripped form tried, and only a canonical answer adopted. That is why the last two rows keep their reasoning variant — the registry considers those models of their own — while the first two do not. `evaluation_id` is keyed by the raw slug, because 4 benchmarks score `claude-sonnet-4-5-20250929` and `claude-sonnet-4-5-thinking-20250929` side by side and they resolve to one canonical id. `reasoning` uses the typed `GenerationArgs` field; the tier goes in `generation_config.additional_details` since the schema has no typed effort field. `--no-registry-resolve` skips resolution for an offline run and says so in every record.

## Automation contract

Registered as `kaggle` in `adapters/catalog.py`: collection `kaggle`, weekly on Friday, `timeout_minutes=30`, `skip_if_unchanged=False` (`ListBenchmarks` reports a benchmark's *definition*, not its submissions, so a cheap source version would skip runs that had new results).

- `--all` means every published benchmark. What a scheduled sweep leaves out is a declared argument — `extra_args=('--all', '--exclude', 'cohere-labs/global-mmlu-lite')` — so `global_mmlu_lite` keeps sole ownership of its collection without a rule hidden in the adapter. An explicit `--benchmark` still converts an excluded benchmark.
- Accounting: an unscored task (`resultCase: none`, ~13% of 108k tasks) is an **exclusion**; an unconvertible scored task, a row whose every task failed, an unresolvable developer, a published benchmark the adapter cannot describe, or a leaderboard that cannot be fetched is a **failure**. Any failure writes `adapter_reports/kaggle_failures.json` and exits non-zero. Records are published per benchmark, so a sweep stopped by the job timeout keeps what it converted and one pathological leaderboard does not abort the rest.
- The adapter owns its HTTP call site — the RPC needs a POST on a handshaken session, and a 1,080-leaderboard sweep needs to tell a retryable 429 from a final 403 — so it retries with capped backoff and calls `raw_capture.record` itself.
- A run prints one line: how many distinct model ids were resolved and how many are unverified. Everything else about resolution is in the records (`model_id_resolution*`, `model_id_needs_review`), so auditing means querying the output rather than scraping a log.

Also adds `ministral`, `codestral` and `sarvam` to `helpers/developer.py`; they were the only unresolvable developers in the entire corpus, and each new namespace is pinned in `tests/test_developer.py`.

## Verification

Against the live source, with the exact argv the catalog schedules:

| Check | Result |
|---|---|
| Full `--all` sweep | **exit 0** in **314s**, 1,080 benchmarks, 16,174 leaderboard rows, **15,197 records**, 0 failures |
| `every_eval_ever.validate` over all 15,197 records | **0 errors, 0 warnings** |
| `validator.check_duplicate_entries` over the sweep | no duplicates |
| `tests/test_kaggle_adapter.py` + `tests/test_registry_resolver.py` | both the Kaggle and the registry network seams stubbed, so no test can reach the network; the resolver's "never raises" contract is pinned against malformed 200 replies |
| `metric_id` uniqueness | 0 ids shared by two benchmarks across 7,470 distinct ids |
| `ruff check` + `pytest tests` | clean (2 pre-existing failures unrelated to this branch: `test_helm_extra_is_importable_when_installed` needs `--extra helm`; `TestWarningVisibility::test_failure_with_warning_stays_red` fails on `main` too) |
| Raw capture | 1,078 payloads + manifest under `EEE_RAW_CAPTURE_DIR`, satisfying `captures_raw=True` |

## One upstream issue found, not worked around

The registry returns **two namespaces for one publisher**, both `strategy: exact`, `confidence: 1.0`, `review_status: reviewed`:

- `grok-3` → `xai/grok-3` but `grok-4.3` → `x-ai/grok-4.3`
- `llama-3.1-405b-instruct` → `meta/llama-3-1-405b-instruct` but `llama-3.1-70b-instruct` → `meta-llama/Llama-3.1-70B-Instruct`
- `deepseek-v3` → `deepseek-ai/...` but `deepseek-r1-0120` → `deepseek/...`

That splits one publisher's models across two datastore directories, which is why `data/kaggle/` has 15 of them including both `xai` and `x-ai`. The adapter does not overrule the registry — that would be inventing ids again — so every record keeps the resolution strategy, confidence, status and raw slug, and `ls data/kaggle/` shows the split. Worth fixing in `eval-card-registry` seed data.

Separately, `grok-4-fast-reasoning` resolves to `xai/grok-4` at confidence 1.0, collapsing a distinct model into its base. Reported the same way.

Rebased from the pre-`every_eval_ever/adapters/` layout onto current `main`.

### DIFF
--- a/utils/kaggle/adapter.py
+++ b/utils/kaggle/adapter.py
@@ -1,0 +1,541 @@
+"""
+Generalized adapter for Kaggle Community Benchmarks.
+
+Kaggle Benchmarks are community-published evaluation suites. Each benchmark is
+identified by an ``owner/slug`` pair (e.g. ``cohere-labs/global-mmlu-lite``) and
+exposes a public leaderboard. This adapter:
+
+1. Optionally enumerates ALL published benchmarks via Kaggle's (undocumented but
+   unauthenticated) ``ListBenchmarks`` RPC, and/or accepts explicit ``owner/slug``
+   pairs on the command line.
+2. Fetches each benchmark's leaderboard from the public REST endpoint.
+3. Converts every model row (and every task result within it) into the EvalEval
+   schema, handling both numeric and boolean task results.
+
+Data sources:
+- List:        POST https://www.kaggle.com/api/i/benchmarks.BenchmarkService/ListBenchmarks
+- Leaderboard: GET  https://www.kaggle.com/api/v1/benchmarks/{owner}/{slug}/leaderboard
+
+Usage:
+    # Convert specific benchmark(s)
+    uv run python -m utils.kaggle.adapter \
+        --benchmark cohere-labs/global-mmlu-lite \
+        --output-dir /tmp/eee-kaggle
+
+    # Discover and convert all published benchmarks (slow)
+    uv run python -m utils.kaggle.adapter --all --output-dir data/kaggle
+
+    # Smoke test: discover but only convert the first 5
+    uv run python -m utils.kaggle.adapter --all --limit 5 --output-dir /tmp/eee-kaggle
+"""
+
+import argparse
+import time
+from typing import Dict, Iterator, List, Optional, Tuple
+
+import requests
+
+from every_eval_ever.eval_types import (
+    ConfidenceInterval,
+    EvalLibrary,
+    EvaluationLog,
+    EvaluationResult,
+    EvaluatorRelationship,
+    MetricConfig,
+    ScoreDetails,
+    ScoreType,
+    SourceDataUrl,
+    Uncertainty,
+)
+from every_eval_ever.helpers import (
+    SCHEMA_VERSION,
+    FetchError,
+    fetch_json,
+    make_model_info,
+    make_source_metadata,
+    save_evaluation_log,
+)
+
+KAGGLE_BASE = "https://www.kaggle.com"
+LIST_RPC_URL = f"{KAGGLE_BASE}/api/i/benchmarks.BenchmarkService/ListBenchmarks"
+LEADERBOARD_URL = KAGGLE_BASE + "/api/v1/benchmarks/{owner}/{slug}/leaderboard"
+
+# Server-side cap on ListBenchmarks page size.
+LIST_PAGE_SIZE = 200
+
+
+def _kaggle_session() -> Tuple[requests.Session, str]:
+    """Open a session with an anonymous XSRF token.
+
+    The ListBenchmarks RPC is unauthenticated but requires the XSRF cookie/header
+    handshake that Kaggle hands out on any page load.
+    """
+    session = requests.Session()
+    try:
+        session.get(f"{KAGGLE_BASE}/benchmarks", timeout=60)
+    except requests.RequestException as e:
+        raise FetchError(f"Kaggle XSRF handshake failed: {e}") from e
+    xsrf = session.cookies.get("XSRF-TOKEN")
+    if not xsrf:
+        raise FetchError("Could not obtain XSRF-TOKEN cookie from Kaggle")
+    return session, xsrf
+
+
+# Kaggle aggregationType -> normalized metric_kind (for safe cross-source aggregation).
+# Only map aggregation types that describe the *measured value*; generic reduction
+# modes (AVERAGE, SUM) are not metric families and would conflate unrelated metrics,
+# so they are intentionally omitted (the raw type is kept in source additional_details).
+_AGGREGATION_TO_METRIC_KIND = {
+    "PERCENTAGE_PASSED": "pass_rate",
+}
+# Kaggle displayType -> metric_unit. PERCENTAGES are stored as proportions in [0, 1].
+_DISPLAY_TO_UNIT = {
+    "PERCENTAGES": "proportion",
+    "COUNTS": "count",
+}
+
+
+def _benchmark_owner(bench: dict) -> Optional[str]:
+    """Resolve the leaderboard URL owner for a benchmark object.
+
+    Org-owned benchmarks are addressed by the organization slug; otherwise the
+    creating user's username is used.
+    """
+    org = bench.get("organization")
+    if org and org.get("slug"):
+        return org["slug"]
+    return (bench.get("ownerUser") or {}).get("userName")
+
+
+def _benchmark_meta(bench: dict) -> Dict[str, Optional[str]]:
+    """Extract benchmark-level metadata (only present on the discovery path).
+
+    The leaderboard endpoint omits these, but ``ListBenchmarks`` carries the
+    scoring config we need to set direction (``lower_is_better``) and metric
+    unit/kind, which would otherwise be guessed.
+    """
+    version = (bench.get("task") or {}).get("version") or {}
+    return {
+        "benchmark_id": bench.get("id"),
+        "sort_order": version.get("sortOrder"),
+        "aggregation_type": version.get("aggregationType"),
+        "display_type": version.get("displayType"),
+    }
+
+
+def list_benchmarks() -> Iterator[Dict[str, object]]:
+    """Enumerate published Kaggle benchmarks.
+
+    Yields dicts with ``owner``, ``slug``, ``name`` and ``meta``. The leaderboard
+    URL owner is the organization slug when the benchmark is org-owned, otherwise
+    the creating user's username.
+    """
+    session, xsrf = _kaggle_session()
+    headers = {
+        "content-type": "application/json",
+        "accept": "application/json",
+        "x-xsrf-token": xsrf,
+    }
+    page_token = ""
+    while True:
+        body = {"filter": {}, "pageSize": LIST_PAGE_SIZE, "pageToken": page_token}
+        # On a page failure, raise after the benchmarks already yielded: the
+        # caller keeps that partial progress but is told discovery was truncated,
+        # so a `--all` run is never reported as a clean success when its tail is
+        # missing.
+        try:
+            resp = session.post(LIST_RPC_URL, json=body, headers=headers, timeout=60)
+            resp.raise_for_status()
+            data = resp.json()
+        except (requests.RequestException, ValueError) as e:
+            raise FetchError(f"ListBenchmarks page fetch failed: {e}") from e
+        # A 200 error envelope or changed RPC schema would otherwise look like an
+        # empty final page and end discovery as a clean success — validate the
+        # shape so a broken response is a failure, not a silently empty corpus.
+        if not isinstance(data, dict) or not isinstance(data.get("benchmarks"), list):
+            raise FetchError(
+                f"unexpected ListBenchmarks response shape (pageToken={page_token!r})"
+            )
+        for bench in data["benchmarks"]:
+            if not bench.get("published"):
+                continue
+            slug = bench.get("slug")
+            if not slug:
+                continue
+            owner = _benchmark_owner(bench)
+            if not owner:
+                continue
+            yield {
+                "owner": owner,
+                "slug": slug,
+                "name": bench.get("name", slug).strip(),
+                "meta": _benchmark_meta(bench),
+            }
+        page_token = data.get("nextPageToken") or ""
+        if not page_token:
+            break
+
+
+def fetch_leaderboard(owner: str, slug: str) -> Optional[List[dict]]:
+    """Fetch a benchmark leaderboard.
+
+    Returns the list of rows on success (possibly empty for a benchmark with no
+    submissions), or ``None`` when the fetch itself failed (HTTP/network/parse
+    error, including the 403 returned for non-existent/private benchmarks). The
+    caller distinguishes these: an empty list is "no data", ``None`` is a failure
+    that must be surfaced rather than silently treated as empty.
+    """
+    url = LEADERBOARD_URL.format(owner=owner, slug=slug)
+    try:
+        data = fetch_json(url)
+    except FetchError as e:
+        print(f"  ! could not fetch leaderboard for {owner}/{slug}: {e}")
+        return None
+    # A 200 with an error envelope or a changed schema (no list-valued `rows`) is
+    # a fetch failure, not an empty leaderboard — don't let an upstream API break
+    # masquerade as "no data".
+    rows = data.get("rows") if isinstance(data, dict) else None
+    if not isinstance(rows, list):
+        print(f"  ! unexpected leaderboard response shape for {owner}/{slug}")
+        return None
+    return rows
+
+
+def _build_eval_result(
+    task: dict, owner: str, slug: str, meta: Optional[Dict[str, object]] = None
+) -> Optional[EvaluationResult]:
+    """Convert a single ``taskResult`` entry into an EvaluationResult.
+
+    Handles numeric results (continuous score in [0, 1]) and boolean results
+    (binary 0/1). Other result cases (e.g. custom tuples) are skipped.
+
+    ``meta`` carries benchmark-level scoring config (from the discovery path)
+    used to set ``lower_is_better`` and the metric unit/kind; when absent these
+    fall back to sensible defaults.
+    """
+    meta = meta or {}
+    task_name = task.get("benchmarkTaskName") or slug
+    result = task.get("result", {})
+
+    # Direction: Kaggle sorts leaderboards DESCENDING when higher is better.
+    lower_is_better = meta.get("sort_order") == "ASCENDING"
+
+    score: Optional[float] = None
+    score_type: Optional[ScoreType] = None
+    min_score: Optional[float] = None
+    max_score: Optional[float] = None
+    metric_kind: Optional[str] = None
+    metric_unit: Optional[str] = None
+    uncertainty: Optional[Uncertainty] = None
+
+    if result.get("hasNumericResult"):
+        numeric = result.get("numericResult") or {}
+        value = numeric.get("value")
+        if value is None:
+            return None
+        try:
+            score = float(value)
+        except (TypeError, ValueError):
+            return None
+        # Aggregation-derived kind/unit describe a numeric metric; they are
+        # meaningless for a single boolean pass/fail, so only attach them here.
+        metric_kind = _AGGREGATION_TO_METRIC_KIND.get(meta.get("aggregation_type"))
+        metric_unit = _DISPLAY_TO_UNIT.get(meta.get("display_type"))
+        # Kaggle's API does not expose a metric's scale. Most numeric results are
+        # accuracies / pass-rates in [0, 1]; for those we can safely declare a
+        # bounded continuous metric. Values outside [0, 1] (counts, percentages,
+        # latencies, ...) have an unknown scale, so we leave score_type and bounds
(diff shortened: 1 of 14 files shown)


utils/kaggle/adapter.py
```diff
@@ -0,0 +1,541 @@
...
+    score: Optional[float] = None
+    score_type: Optional[ScoreType] = None
+    min_score: Optional[float] = None
+    max_score: Optional[float] = None
+    metric_kind: Optional[str] = None
+    metric_unit: Optional[str] = None
+    uncertainty: Optional[Uncertainty] = None
+
+    if result.get("hasNumericResult"):
+        numeric = result.get("numericResult") or {}
```
mrshu, line 232: **[high] result parsing** — A supported result stored in `numericResultNullable` is dropped because this path reads only `numericResult`. The retained Kaggle adapter accepts the alternate shape.

The same fallback can stay in this generalized parser, with a focused payload test.

utils/kaggle/adapter.py
```diff
@@ -0,0 +1,541 @@
...
+        "benchmark_owner": owner,
+        "benchmark_url": benchmark_url,
+    }
+    for key in ("benchmark_id", "aggregation_type", "display_type"):
+        if meta.get(key) is not None:
+            source_details[key] = str(meta[key])
+
+    count = 0
+
+    for row in rows:
```
mrshu, line 329: **[high] partial conversion** — A row without its required model slug disappears from an otherwise successful benchmark. No failure summary names the missing model.

Missing identities can count as partial conversion failures while valid rows continue. Unsupported result types can remain a separate skip category.

utils/kaggle/adapter.py
```diff
@@ -0,0 +1,541 @@
...
+        )
+    if discovery_failed:
+        print(
+            "WARNING: benchmark discovery was truncated by a fetch error; "
+            "the converted set is incomplete."
+        )
+    print("=" * 60)
+    # Non-zero exit when any fetch/conversion failed or discovery was truncated,
+    # so callers and CI don't read a partial run as a clean success.
+    if failed_fetches or conversion_failures or discovery_failed:
```
mrshu, line 536: **[high] exit status** — A non-empty leaderboard that emits zero records enters `empty_or_dropped`, but this final condition still returns success. Automation cannot distinguish the incomplete export from a complete one.

That category can contribute to the final incomplete-run status.

utils/kaggle/adapter.py
```diff
@@ -0,0 +1,541 @@
...
+        except (requests.RequestException, ValueError) as e:
+            raise FetchError(f"ListBenchmarks page fetch failed: {e}") from e
+        # A 200 error envelope or changed RPC schema would otherwise look like an
+        # empty final page and end discovery as a clean success — validate the
+        # shape so a broken response is a failure, not a silently empty corpus.
+        if not isinstance(data, dict) or not isinstance(data.get("benchmarks"), list):
+            raise FetchError(
+                f"unexpected ListBenchmarks response shape (pageToken={page_token!r})"
+            )
+        for bench in data["benchmarks"]:
```
mrshu, line 159: **[high] discovery** — A published entry without a usable slug or owner is silently skipped, and `--all` can still report complete discovery.

The scan can preserve later valid entries and carry an incomplete-discovery status to the final exit code.

utils/kaggle/adapter.py
```diff
@@ -0,0 +1,541 @@
...
+    metric_kind: Optional[str] = None
+    metric_unit: Optional[str] = None
+    uncertainty: Optional[Uncertainty] = None
+
+    if result.get("hasNumericResult"):
+        numeric = result.get("numericResult") or {}
+        value = numeric.get("value")
+        if value is None:
+            return None
+        try:
```
mrshu, line 236: **[medium] numeric validation** — `float()` accepts non-finite values, which serialize as `null` even though the schema requires numbers.

A non-finite score can drop that task. A non-finite interval can omit only uncertainty and retain its valid score.

utils/kaggle/adapter.py
```diff
@@ -0,0 +1,541 @@
...
+            count = convert_benchmark(
+                owner,
+                slug,
+                name,
+                rows,
+                args.output_dir,
+                retrieved_timestamp,
+                meta=bench.get("meta"),
+            )
+        except (TypeError, AttributeError, KeyError) as e:
```
mrshu, line 496: **[medium] error isolation** — Pydantic `ValidationError` is outside this catch tuple. One malformed benchmark can stop all later targets instead of joining the final partial-run report.

That error can stay inside the per-benchmark boundary while the adapter continues with later work.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The adapter has a thoughtful partial-run design,
but six reachable paths still lose data, emit invalid records, or report an
incomplete run as successful.
**Result:** Six actionable findings remain. Two metadata questions were set
aside because no affected upstream benchmark was confirmed.
**Scope:** This draft adds generalized Kaggle benchmark discovery, explicit
targets, conversion, metadata enrichment, and partial-run reporting.
**Checks:** 4 concerns checked; 2 confirmed, 1 narrowed, and 1 lowered in
priority.
**Second check:** 2 Codex reviewers checked 9 findings; 7 kept actionable and 2
set aside.

## Earlier review comments

None.

## What I noticed

|---|---|---|---|---|
| RAVF001 | high | result parsing | `utils/kaggle/adapter.py:232` | A supported `numericResultNullable` payload loses its evaluation because this path reads only `numericResult`. |
| RAVF002 | high | partial conversion | `utils/kaggle/adapter.py:329` | A row without its required model slug disappears from an otherwise successful benchmark with no failure summary. |
| RAVF003 | high | exit status | `utils/kaggle/adapter.py:536` | A non-empty leaderboard can emit zero records, enter `empty_or_dropped`, and still return success. |
| RAVF008 | high | discovery | `utils/kaggle/adapter.py:159` | The `--all` path silently skips a published entry without a usable slug or owner. Discovery still reports success. |
| RAVF004 | medium | numeric validation | `utils/kaggle/adapter.py:236` | Non-finite scores and confidence bounds serialize as `null`, although the schema requires numbers. |
| RAVF005 | medium | error isolation | `utils/kaggle/adapter.py:496` | Pydantic validation errors escape the per-benchmark boundary and stop later targets from running. |

<details>

- **[low] CLI validation** — `--limit 0` disables the cap, while a negative
  value still converts one discovered benchmark. (`RAVF009`)

</details>

<details>
<summary>Things to try (7 items)</summary>

- **[high] result parsing** — The established `numericResultNullable` fallback
  can remain part of this generalized parser, with a focused payload test.
  (`RAVF001`)
- **[high] partial conversion** — Missing required model identities can be
  tracked as partial failures. Valid rows can continue, while unsupported
  result types stay in a separate skip category. (`RAVF002`)
- **[high] exit status** — A non-empty leaderboard with no emitted records can
  contribute to the final incomplete-run status. (`RAVF003`)
- **[high] discovery** — Discovery can continue after malformed published
  entries, preserve valid work, and carry an incomplete-discovery status to the
  final exit code. (`RAVF008`)
- **[medium] numeric validation** — A non-finite score can drop that task. A
  non-finite interval can omit only uncertainty and retain its valid score.
  (`RAVF004`)
- **[medium] error isolation** — Pydantic `ValidationError` can join the
  per-benchmark failure boundary so later targets still run. (`RAVF005`)
- **[low] CLI validation** — Non-positive limit values can fail argument
  validation before discovery begins. (`RAVF009`)

</details>

- **[medium] metric bounds** — Set aside because the inconsistent
  metadata-free heuristic was proven only with synthetic values. No affected
  upstream Kaggle task was confirmed. (`RAVF006`)
- **[medium] metric direction** — Set aside because the explicit-only branch is
  real, but no reachable ascending benchmark was identified. A proportional
  direction contract also needs a product choice. (`RAVF007`)

<details>

- Target: PR #191 (`mrshu/kaggle-benchmarks-adapter`, 4 files, +1057/-0)
- Mix: 3 codex-exec
- Checks: concerns=4; confirmed=2; narrowed=1; lowered=1; ruled-out=0;
  set-aside=0
- Second check: targeted; reviewers=2; kept=7; clarified=5; set-aside=2;

</details>

---

borgr:
Rebased onto current main. One conflict, in `every_eval_ever/adapters/README.md`, where main had appended a `### Papers with Code DrugBank` section at the same spot this branch appends `### Kaggle Benchmarks`. Both kept, DrugBank first. Nothing else needed a hand — `catalog.py`, `helpers/developer.py`, `helpers/__init__.py`, `open_medical_llm/adapter.py` and `tests/test_developer.py` all auto-merged.

Rebased branch locally: 1451 passed, 1 skipped, ruff clean.

utils/kaggle/adapter.py
borgr: Addressed in the current tree. `every_eval_ever/adapters/kaggle/adapter.py:624` reads `numericResult` and falls back to `numericResultNullable`. Resolving; the file moved out of `utils/`, which is why GitHub marks this outdated rather than done.

utils/kaggle/adapter.py
borgr: Addressed. A row without its model slug raises from `require_identity`, and the `except ValueError` at `every_eval_ever/adapters/kaggle/adapter.py:851` files it as a `SourceRecordFailure` carrying `source_ref=row_ref`, so the row is named in the failure report while valid rows keep converting. `test_a_row_whose_every_task_failed_is_a_failure_not_an_exclusion` pins it.

utils/kaggle/adapter.py
borgr: Addressed. `resolve_targets` returns `(targets, failures)`, `main` seeds its failure list from that second value, and `raise_if_incomplete()` at `every_eval_ever/adapters/kaggle/adapter.py:1081` turns it into a non-zero exit. Valid later entries survive. `test_truncated_index_keeps_partial_targets_and_reports_it` and `test_an_undescribable_published_benchmark_is_a_failure` cover it.

utils/kaggle/adapter.py
borgr: Addressed, and split the way you suggested. `require_finite_number` (`every_eval_ever/helpers/io.py:213`) rejects NaN and infinity, so a non-finite score drops that task. `_non_negative` (`every_eval_ever/adapters/kaggle/adapter.py:552`) routes a non-finite interval through the same check and returns `None`, so `_confidence_interval` omits only the uncertainty and the valid score is kept.

utils/kaggle/adapter.py
borgr: Addressed. Pydantic v2 `ValidationError` subclasses `ValueError`, so it is already inside the per-benchmark `except ValueError` boundary, and the comment at `every_eval_ever/adapters/kaggle/adapter.py:848` states that this catch is `ValueError` only on purpose — widening it to `TypeError`/`AttributeError` would turn an adapter bug into thousands of quiet per-row failures rather than a crash naming the line.

utils/kaggle/adapter.py
borgr: This one landed differently from what you asked, and I think the per-benchmark half of the difference is right while the run-level half is a real gap.

Per benchmark, the current code files an all-unscored leaderboard as an *exclusion*, which does not reach the exit status — `raise_for_failed_records` (`helpers/io.py:392`) returns early on an empty failure list. That is the better call. Roughly 13% of Kaggle tasks are legitimately unscored, and a freshly published benchmark with rows but no results yet is an ordinary state, so redding a sweep over it would make the nightly noisy about nothing.

The gap is one level up. I measured it on the rebased branch — a sweep whose every row is unscored prints

```
Saved 0 record(s) from 1 leaderboard row(s) across 1 benchmark(s)
```

and exits **zero**. If Kaggle ever changes its result payload shape, that is exactly what the nightly would see, and it would report success having written nothing. Kaggle will not tell us in advance, so a failing flag at that moment is the whole point.

Contained fix, right before `result = SourceConversionResult[Path](` in `main`. It reuses the failure list already there, so the run-level failure lands in the report and the existing `raise_if_incomplete()` handles the exit:

```python
    # A sweep that read leaderboard rows and wrote nothing is a run-level
    # failure, not a corpus of exclusions. An individual benchmark whose every
    # task is unscored is ordinary, so this is deliberately not a per-benchmark
    # check; it exists so a change in Kaggle's result payload cannot pass as a
    # clean run.
    if total_rows and not written:
        failures.append(
            SourceRecordFailure(
                source_ref="run",
                reason=(
                    f"{total_rows} leaderboard row(s) produced no records"
                ),
            )
        )
```

No new import — `SourceRecordFailure` is already in scope.

I verified this rather than eyeballing it. Two tests, both against `_run_main`:

- `test_a_sweep_that_reads_rows_and_writes_nothing_exits_nonzero` — one benchmark, one row, one unscored task. Expects `SourceRecordsError` matching `produced no records`, and a report with `converted_records == 0` and a single `run` failure. **Fails on the branch without the guard** (the run exits zero), passes with it.
- `test_one_all_unscored_benchmark_does_not_red_a_sweep_that_wrote_records` — an all-unscored benchmark alongside a good one, expects `written == 1`. Passes either way, which is the point: it pins that the guard does not over-fire on the ordinary case.

`test_leaderboard_without_submissions_is_empty_not_a_failure` also stays green, since the guard is gated on `total_rows`.

Full suite on the rebased branch with the guard and both tests: 1453 passed, 1 skipped, ruff clean.

Happy to push the 56-line change onto this branch if you would rather not hand-apply it. Everything else here reads well to me and I would merge it as soon as this is settled.

borgr:
Reviewed. I went through your six self-review threads against the rebased tree and resolved five of them — `numericResultNullable`, the unnamed missing-slug row, the non-finite `float()` split, incomplete discovery reaching the exit code, and `ValidationError` inside the per-benchmark boundary are all handled in `every_eval_ever/adapters/kaggle/adapter.py`, with the line references in each thread. They only looked open because the file moved out of `utils/`, so GitHub marked them outdated rather than done.

The exit-status one is the only thing left, and it is a small gap rather than a rework. Per benchmark, filing an all-unscored leaderboard as an exclusion is the better call than what the thread asked for — ~13% of Kaggle tasks are legitimately unscored. But a *sweep* whose every row is unscored currently prints `Saved 0 record(s) from 1 leaderboard row(s)` and exits zero, which is what a change in Kaggle's payload shape would look like to the nightly. Verified patch and two tests are in [the thread](https://github.com/evaleval/every_eval_ever/pull/191#discussion_r3945168943), including one that fails without the guard and one that pins it against over-firing.

Apart from that, this is good to merge. The adapter is careful in the places that matter — refusing to invent a unit or scale where Kaggle publishes none, keeping owner in the identifier because 15 slugs are published by more than one owner, and leaving self-named sub-benchmark columns as published rather than promoting them into the aggregate.

PR #192, opened 2026-06-22
[Submission] Add tau-bench leaderboard adapter

benshi34:
## Summary

Adds a tau-bench adapter that fetches the public leaderboard submissions manifest and per-submission JSON from `sierra-research/tau2-bench`, then converts populated domain metrics into Every Eval Ever aggregate records.

The adapter supports:
- text, voice, and legacy manifest sections
- `pass_1` through `pass_4`
- reported per-domain cost
- banking knowledge retrieval config metadata
- local `--input-dir` replay for review/testing

No instance-level trajectory export is included in this first version.

## Validation

- `uv run ruff check utils/tau_bench/adapter.py tests/test_tau_bench_adapter.py`
- `uv run pytest tests/test_tau_bench_adapter.py`
- `uv run python -m utils.tau_bench.adapter --input-dir /Users/ben.sierra.ai/Documents/tau2-bench/web/leaderboard/public/submissions --output-dir /private/tmp/eee-tau-bench-smoke --limit 3`
- `uv run python -m every_eval_ever validate /private/tmp/eee-tau-bench-smoke`
- `uv run python -m utils.tau_bench.adapter --output-dir /private/tmp/eee-tau-bench-live-smoke --limit 1`
- `uv run python -m every_eval_ever validate /private/tmp/eee-tau-bench-live-smoke`


Solus-QE:
it looks like this is fine to be approved, but there are a couple of notes that can be implemented either now or later; they are not of utmost importance and do not block this feature right now:

- Proper package imports (every_eval_ever.helpers / .eval_types) rather than the older sys.path hack.
- cost metric correctly leaves score_type=None, so it sidesteps the MetricConfig validator's min_score/max_score requirement, with lower_is_better=True set.
- parse_score preserves a real 0.0 (raw is None or raw == '') — no falsy-zero bug.
- Voice/methodology metadata goes into GenerationConfig.additional_details, not GenerationArgs (which is extra='forbid'), and all detail dicts are stringified to match the dict[str, str] schema.
- Submission-ID dedup across manifest sections is handled.
- In run(), make_logs has no per-record error isolation, so a single non-numeric score or malformed results block raises and aborts the whole export

utils/tau_bench/adapter.py
```diff
@@ -0,0 +1,668 @@
...
+
+def make_metric_config(*, domain: str, metric: str) -> MetricConfig:
+    if metric.startswith('pass_'):
+        k = int(metric.split('_', 1)[1])
+        return MetricConfig(
+            evaluation_description=(
+                f'tau-bench {domain} Pass^{k} success rate reported on '
+                'the public leaderboard.'
+            ),
+            metric_id='tau_bench.pass_at_k',
```
mrshu, line 389: **[high] metric semantics** — Pass^k is identified as pass@k

Upstream computes this value as `pass_hat_k`: all selected trials must succeed. Pass@k instead means at least one succeeds. Using the pass@k ID can merge mathematically different scores in downstream joins.

The repository already uses the global `pass_hat_k` ID for Pass^k. Keeping the existing display name and `k` parameter would preserve tau-bench context.

utils/tau_bench/adapter.py
```diff
@@ -0,0 +1,668 @@
...
+        path = input_dir / submission_id / SUBMISSION_FILE_NAME
+        submission = json.loads(path.read_text(encoding='utf-8'))
+        records.append(
+            TauBenchSubmission(
+                submission_id=submission_id,
+                manifest_section=section,
+                submission=submission,
+                source_url=submission_source_url(submission_id),
+            )
+        )
```
mrshu, line 197: **[medium] provenance** — Local replay is attributed to mutable upstream `main`

Remote mode does not record a resolved revision. Local mode also builds the current upstream URL from only the submission ID, even when local bytes differ. The record therefore cannot identify the exact source that produced its scores.

The default GitHub source can resolve one commit per run. Custom and local inputs can keep their supplied provenance plus a content hash without claiming unrelated upstream bytes.

utils/tau_bench/adapter.py
```diff
@@ -0,0 +1,668 @@
...
+    if value is None or str(value).strip() == '':
+        raise ValueError(f'{submission_id} is missing required field {key}')
+    return str(value)
+
+
+def parse_score(raw: Any, *, context: str) -> float | None:
+    if raw is None or raw == '':
+        return None
+    try:
+        return float(raw)
```
mrshu, line 596: **[medium] score validation** — Supported replay input can emit values outside the source contract

`float()` accepts booleans, NaN, infinity, Pass^k values outside 0–100, and negative costs. Current official submissions are valid, but local replay and custom sources can reach this path.

Pass metrics can use finite inclusive 0–100 checks. Cost can use a finite non-negative check, with booleans rejected before numeric conversion.

utils/tau_bench/adapter.py
```diff
@@ -0,0 +1,668 @@
...
+    'Distyl AI': 'distyl',
+    'Google': 'google',
+    'Moonshot AI': 'moonshot-ai',
+    'Multiple providers': 'multiple',
+    'NVIDIA': 'nvidia',
+    'OpenAI': 'openai',
+    'Qwen': 'qwen',
+    'Sierra': 'sierra',
+    'xAI': 'xai',
+    'Zhipu AI': 'zhipu-ai',
```
mrshu, line 91: **[medium] developer identity** — The current `Z.ai` submission gets a second developer slug

The organization map handles `Zhipu AI` but not `Z.ai`, so fallback slugging creates `z.ai`. Other repository adapters and developer helpers use `zhipu-ai` for the same provider.

Adding the exact `Z.ai` alias to this existing map would keep the bundle developer, model ID, and output path aligned.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The adapter is clear and well tested, and four data-contract details would benefit from follow-up.
**Result:** One high-priority metric identity issue and three medium-priority provenance or validation issues remain.
**Scope:** This PR adds a tau-bench leaderboard adapter for remote and local submissions.
**Checks:** 4 concerns checked; 3 confirmed, 1 narrowed, 0 ruled out, 0 set aside.
**Second check:** targeted, 2 reviewers; 4 kept, 3 fix paths clarified, 0 set aside, 0 removed.

## Earlier review comments

<details>
<summary>Earlier review comments (5 items)</summary>

- **Fixed:** [Schema-version test](https://github.com/evaleval/every_eval_ever/pull/192#discussion_r3460354684) now uses `adapter.SCHEMA_VERSION`.
- **Still present at an outdated anchor:** [`tau2_bench_version`](https://github.com/evaleval/every_eval_ever/pull/192#discussion_r3460354762) still assumes truthy `methodology` values are mappings.
- **Still present at an outdated anchor:** [`evaluation_date`](https://github.com/evaleval/every_eval_ever/pull/192#discussion_r3460354808) keeps the same malformed-input assumption.
- **Still present at an outdated anchor:** [Generation config](https://github.com/evaleval/every_eval_ever/pull/192#discussion_r3460354842) keeps the same malformed-input assumption.
- **Still present at an outdated anchor:** [Remote `--limit`](https://github.com/evaleval/every_eval_ever/pull/192#discussion_r3460354884) still slices after all selected submissions are downloaded.

</details>

## What I noticed

- **[high] metric semantics** `utils/tau_bench/adapter.py:389` — Tau-bench Pass^k values use a pass@k identifier. Upstream computes all-k success as `pass_hat_k`, while pass@k means at least one success. Downstream metric joins can combine incompatible scores. (`RAVF001`; inline)
- **[medium] provenance** `utils/tau_bench/adapter.py:197` — Remote records cite mutable `main`, and local replay invents the same upstream URL for local bytes. The exported record cannot identify the exact input that produced its scores. (`RAVF002`; inline)
- **[medium] score validation** `utils/tau_bench/adapter.py:596` — Local and custom input accepts booleans, non-finite values, out-of-range Pass^k percentages, and negative costs. Current official submissions are valid, but supported replay inputs can emit invalid scores. (`RAVF003`; inline)
- **[medium] developer identity** `utils/tau_bench/adapter.py:91` — The current upstream `Z.ai` submission falls back to developer slug `z.ai`. Other repository mappings use `zhipu-ai`, so the same provider splits across identities and output paths. (`RAVF004`; inline)

## Things to try

<details>

- **[high] metric semantics** — The global `pass_hat_k` identifier can match the existing repository mapping while keeping tau-bench names, result IDs, and `k` parameters. (`RAVF001`)
- **[medium] provenance** — The default GitHub source can resolve one commit. Custom sources and local replay can use content hashes without claiming unrelated upstream URLs. (`RAVF002`)
- **[medium] score validation** — Pass metrics can enforce finite 0–100 values, while cost can enforce finite non-negative values. Boolean inputs need a separate rejection. (`RAVF003`)
- **[medium] developer identity** — Adding the exact `Z.ai` alias to the existing map would keep both bundle developer and model ID consistent. (`RAVF004`)

</details>

<details>

- Target: PR #192 (`benshi34/tau-bench-adapter`, 4 files, +897/-0)
- Mix: 3 codex-exec
- Focus: correctness, upstream format behavior, provenance, maintainability, and constructive suggestion-oriented language
- Earlier review comments: 5 comments; 4 still present and 1 fixed
- Checks: concerns=4; confirmed=3/narrowed=1/ruled-out=0/set-aside=0
- Second check: targeted; reviewers=2; kept=4/clarified=3/set-aside=0/removed=0; approval changed no
- Set aside: 0 items

</details>

---

utils/tau_bench/adapter.py
```diff
@@ -0,0 +1,668 @@
...
+    model_slug = slugify(model_name)
+    model_id = f'{developer}/{model_slug}'
+    results = make_results(record, model_id)
+    if not results:
+        return None
+
+    evaluation_timestamp = evaluation_date(submission)
+    version = (
+        (submission.get('methodology') or {}).get('tau2_bench_version')
+    ) or 'unknown'
```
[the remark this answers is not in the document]
yananlong: The adapter consumes source JSON without validating it against tau-bench’s submission schema first. Rather than adding separate `isinstance(..., dict)` guards everywhere, could we validate each submission.json once at the loader boundary and then let the conversion code assume the documented shape?

That would also catch malformed `model_release`, `voice_config`, domain results, required fields, and manifest values consistently. We should probably keep the explicit finite/range checks for scores too, since Python’s JSON handling and custom inputs can still introduce edge cases.

borgr:
@benshi34 this one cannot be brought up to date by a merge, and I want to be concrete about why rather than leave it sitting.

The branch is 172 commits behind `main` and predates #218, which moved every adapter from `utils/<name>/` to `every_eval_ever/adapters/<name>/`. So the merge is a modify/delete collision on the whole `utils/` tree — `utils/README.md` is deleted on main and modified here, and `utils/tau_bench/` has no counterpart to merge into. Nothing to resolve, the layout the PR targets is gone.

The four files replant cleanly, and I checked that much: move `utils/tau_bench/` to `every_eval_ever/adapters/tau_bench/`, change the two `python -m utils.tau_bench.adapter` lines in the module docstring and the `from utils.tau_bench import adapter` line in the test, and all 5 tests in `tests/test_tau_bench_adapter.py` pass on current main.

What the replant does not cover, all of it landed after this branch:

- **Catalog entry.** Every package under `every_eval_ever/adapters/` must be declared in `every_eval_ever/adapters/catalog.py`, and `tests/test_adapter_catalog.py::test_every_adapter_package_is_accounted_for` fails without one. The entry declares the collections the adapter may write, the argv that redirects its output, and its cadence.
- **Publication path.** The adapter calls `save_evaluation_log`. Current convention is `converters.common.publication.publish_evaluation_logs` or `EvaluationLogOutput`, with `SourceConversionResult` plus `save_failure_report` accounting for every submission it could not convert. See AGENTS.md § Conventions.
- **Raw capture.** A scheduled adapter that fetches live sources leaves a raw-capture manifest; `captures_raw` in the catalog is the reviewed exemption for adapters that only read local files.
- **The open review findings** from the 2026-07-29 pass, the metric-identity one in particular.

Happy to do the replant and open it as a PR with you as author of the commits if you would rather not redo it — say the word. Otherwise it is yours, and the four bullets above are the whole gap.

borgr:
Picked this up rather than leaving it — #268 continues this branch with `main` merged in, the relocation to `every_eval_ever/adapters/tau_bench/`, and all four review-anvil findings plus the four older threads answered. @benshi34's commit is the base of that history, not a rewrite of it.

It is a separate PR only because this branch is on a fork I cannot push to. Everything is in #268's description, finding by finding, and it is green on all four CI rows.

Leaving this open until #268 actually lands, so the work is never in limbo — say the word if you would rather land it from the fork instead and I will hand the commits over.

utils/tau_bench/adapter.py
borgr: Fixed in `02ebfad46`. The pass metrics are `tau_bench.pass_hat_k` now, and the metric says which of the two quantities it is rather than leaving the id to imply it:

    metric_semantics: "pass_hat_k: all 2 trials succeed. Not pass@k, which
    counts a task solved when at least one of k trials succeeds."

Kept `Pass^{k}` as `metric_name`, the `{"k": k}` parameter, and the `tau_bench.<modality>.<domain>.<metric>` result ids, so nothing about the tau-bench naming changed — only the global id a consumer joins on. I checked the eval-card-registry for a canonical id first: it has `pass-at-1` and `pass-at-8`, both on a 0–1 scale, and no `pass-hat-k`, so the source-namespaced form is the right one here. `test_pass_metric_names_pass_hat_k_and_says_so` covers it.

every_eval_ever/adapters/tau_bench/adapter.py
```diff
@@ -0,0 +1,668 @@
...
+        path = input_dir / submission_id / SUBMISSION_FILE_NAME
+        submission = json.loads(path.read_text(encoding='utf-8'))
+        records.append(
+            TauBenchSubmission(
+                submission_id=submission_id,
+                manifest_section=section,
+                submission=submission,
+                source_url=submission_source_url(submission_id),
+            )
+        )
```
borgr: Fixed in `02ebfad46`, both halves.

Remote: `pin_base_url` resolves the base URL's ref through the GitHub commits API **once, before the manifest is fetched**, and rewrites it to `raw.githubusercontent.com/<owner>/<repo>/<sha>/...`. So every URL in a run is immutable and the batch records one `submission_commit`. A base URL already naming a 40-hex sha is left alone; one that is not raw-GitHub at all is passed through, and then the content hash is the only identity.

Local: no invented upstream URL. `source_url` is `None`, `source_data.url` carries only the leaderboard, and the record states `local_input_path` instead.

Both: every submission carries `submission_sha256`, hashed from the exact bytes the loader read (canonical JSON for a remote payload, the file bytes for a local one), so a record names its input even when the input was `--base-url` pointing somewhere else entirely.

`--allow-unpinned-source` keeps a run possible when the API cannot be reached; without it an unresolvable ref stops the run rather than quietly publishing a mutable citation. Covered by `test_remote_run_pins_the_ref_and_records_the_bytes`, `test_an_unresolvable_ref_stops_the_run_unless_it_is_allowed` and `test_local_replay_does_not_claim_an_upstream_url`.

utils/tau_bench/adapter.py
borgr: Fixed in `02ebfad46`. `parse_score` now goes through `require_finite_number`, which rejects booleans and NaN/infinity, and takes the scale the metric is reported on:

    PASS_SCORE_BOUNDS = (0.0, 100.0)
    COST_SCORE_BOUNDS = (0.0, None)

A value outside it fails with the submission, domain and metric in the message. `test_replay_inputs_cannot_smuggle_impossible_scores` walks `101.0`, `-0.5`, `True`, `inf` and `"nan"` for Pass^k and a negative cost.

Your framing is the reason it is worth doing at all — the public submissions are all valid, so this only ever fires on `--input-dir` or a custom `--base-url`, which is exactly where a hand-built payload comes from.

every_eval_ever/adapters/tau_bench/adapter.py
```diff
@@ -0,0 +1,668 @@
...
+    'Distyl AI': 'distyl',
+    'Google': 'google',
+    'Moonshot AI': 'moonshot-ai',
+    'Multiple providers': 'multiple',
+    'NVIDIA': 'nvidia',
+    'OpenAI': 'openai',
+    'Qwen': 'qwen',
+    'Sierra': 'sierra',
+    'xAI': 'xai',
+    'Zhipu AI': 'zhipu-ai',
```
borgr: Fixed in `02ebfad46`, slightly wider than the suggestion. Adding the `Z.ai` key alone would fix this submission and leave the next spelling to split the same provider again, so provider names are matched normalized (`strip().lower()`) against the map, and `Z.ai`, `Zhipu`, `Zhipu AI` all resolve to `zhipu-ai`. A genuinely new provider still falls through to `slugify`. `test_one_provider_spelled_two_ways_gets_one_developer` pins both the collapse and the fallthrough.

tests/test_tau_bench_adapter.py
```diff
@@ -0,0 +1,227 @@
...
+    bundles = adapter.make_logs(
+        sample_records(), retrieved_timestamp='1234567890.0'
+    )
+
+    assert len(bundles) == 2
+    for bundle in bundles:
+        validated = EvaluationLog.model_validate(bundle.log.model_dump())
+        assert validated.schema_version == '0.2.2'
+        assert validated.source_metadata.source_name == 'tau-bench Leaderboard'
+        assert validated.source_metadata.source_type.value == 'documentation'
```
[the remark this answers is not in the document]
borgr: Already fixed on the branch — the test reads `adapter.SCHEMA_VERSION` rather than a literal, so a schema bump does not break it.

utils/tau_bench/adapter.py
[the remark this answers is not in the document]
borgr: Fixed in `02ebfad46`. `_mapping(submission.get("methodology"))` returns the block when it is an object and `{}` otherwise, so a submission carrying a string here yields version `unknown` instead of raising. See the reply on @yananlong's thread for why this is one guard rather than a schema validation at the loader.

utils/tau_bench/adapter.py
```diff
@@ -0,0 +1,668 @@
...
+    if model_org and submitter and model_org == submitter:
+        return EvaluatorRelationship.first_party
+    return EvaluatorRelationship.third_party
+
+
+def evaluation_date(submission: dict[str, Any]) -> str | None:
+    methodology = submission.get('methodology') or {}
+    value = methodology.get('evaluation_date') or submission.get(
+        'submission_date'
+    )
```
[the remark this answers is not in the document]
borgr: Fixed in `02ebfad46`, same `_mapping` guard. A malformed `methodology` now falls back to `submission_date`, which is what an absent one already did.

utils/tau_bench/adapter.py
```diff
@@ -0,0 +1,668 @@
...
+
+
+def make_generation_config(
+    submission: dict[str, Any],
+    domain: str,
+) -> GenerationConfig:
+    methodology = submission.get('methodology') or {}
+    voice_config = submission.get('voice_config') or {}
+    pipeline = (
```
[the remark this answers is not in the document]
borgr: Fixed in `02ebfad46`. Both `methodology` and `voice_config` go through `_mapping`, which also removed the five `if isinstance(voice_config, dict) else None` conditionals that were inline in the details dict — the guard is in one place now instead of repeated per field.

utils/tau_bench/adapter.py
```diff
@@ -0,0 +1,668 @@
...
+        base_url=args.base_url,
+        sections=args.sections,
+    )
+    if args.limit is not None:
+        records = records[: args.limit]
+    bundles = make_logs(records)
+    paths = export_logs(bundles, args.output_dir)
+    for path in paths:
+        print(path)
+    return len(paths)
```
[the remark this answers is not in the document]
borgr: Fixed in `02ebfad46`. `--limit` is threaded into `iter_manifest_ids`, which stops collecting ids once it has enough, so the limit bounds the download rather than the export: 5 manifest entries with `--limit 2` is 2 submission fetches, not 5. `test_limit_bounds_the_download_not_just_the_output` counts the fetches.

utils/tau_bench/adapter.py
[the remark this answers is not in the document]
borgr: Good question, and I went the other way — one `_mapping` helper at the read sites rather than schema validation at the loader boundary. Reasoning, and I am happy to be argued out of it:

tau-bench publishes no submission schema. Validating against the documented shape means writing that shape here, from the submissions that happen to exist today, and then a field the leaderboard adds or an older legacy submission that predates it becomes a hard rejection of a real row. That trades a crash on malformed input for dropped valid input, which is the worse direction for this repo — an unknown is data, a rejection loses it.

So the guard is deliberately narrow: `_mapping` says "read this optional block if it is a block", in one function, at the four places a nested object is read. Required fields are still checked where they are required (`required_str` for `model_name` and `model_organization`), the manifest still hard-fails when a section is not a list, and per-domain results still hard-fail when they are not an object — the shapes the conversion genuinely cannot proceed without.

And yes, kept the finite/range checks — they are now `require_finite_number` plus a declared scale per metric, since JSON and custom inputs reach places a schema would not: `NaN`, booleans, a Pass^k of 120.

If we do want a schema, the place for it is `every_eval_ever/adapters/` generally rather than this adapter alone, and it wants an issue — several adapters have the same "trust the upstream shape" question.

borgr:
Pushed to this branch (`9ece64b3e`) rather than opening a second PR, so the threads above stay where the discussion is. @benshi34's commit is still the base of the history; three commits sit on top of it.

Replied on each of the ten threads with what changed. The summary of what is new beyond them:

- **`main` merged in.** This branch predated #218, which moved adapters from `utils/<name>/` to `every_eval_ever/adapters/<name>/`, so the merge takes main's deletion of `utils/README.md` and the next commit relocates `utils/tau_bench/`. That is why the inline anchors above point at `utils/` paths.
- **Publication and accounting**, both landed on main after this branch. `save_evaluation_logs` replaces the per-record `save_evaluation_log`, so a failure part-way leaves the tree as it was; `convert_logs` returns a `SourceConversionResult` where a submission raising is a failure that does not cost the batch and one with no score in any domain is a recorded exclusion instead of a silent drop.
- **Catalog registration.** `parse_args(argv)` so `tests/test_adapter_catalog.py` can check the entry against the real parser, a `tau_bench` entry writing the `tau-bench` collection, and the `adapters/README.md` row. Raw capture comes free through `fetch_json`.

`uv run pytest tests` 878 passed, 31 skipped — this adapter's file went 5 → 16 tests, one per finding. `uv run ruff check` clean.

Labelled `auto-review` for a fresh pass at this head.

borgr:
@Erotemic filed this review on #197, but all three findings are about this PR (the tau-bench adapter), so I am bringing it here and answering point by point. Credit for the findings, and for the voice-metric directionality, goes to that review and to the tau-bench voice panel it cites.

**High — Pass^k identified as pass@k.** Already fixed in `02ebfad46`. The pass metrics carry `metric_id = tau_bench.pass_hat_k`, keep `Pass^{k}` as the display name and `{"k": k}` as the parameter, and the metric config states that all k trials must succeed and this is not pass@k. Covered by `test_pass_metric_names_pass_hat_k_and_says_so`.

**High — Voice submissions silently lose their interaction-quality metrics.** Fixed in `1db6bf2f8`. `make_results` now emits one `EvaluationResult` per per-domain interaction metric. Response and yield latency come out in seconds with lower-is-better, response rate, yield rate and the three selectivity rates as proportions with higher-is-better, and agent interruption rate as a proportion with lower-is-better. Each result carries its own unit, scale and direction, keeps the backing `counts` and the panel `version` and `config` in details, and the `overall` aggregate is preserved under an `overall` domain. Metric ids are namespaced as `tau_bench.interaction.<metric>`. Covered by `test_voice_interaction_metrics_are_emitted_with_units_and_direction` and a range-check test.

**Medium — Score parsing accepts values the source schema forbids.** Already fixed in `02ebfad46`. `parse_score` runs every value through `require_finite_number` and a declared per-metric scale, so a boolean, a non-finite float, a Pass^k above 100 or a negative cost is rejected with the submission, domain and metric named in the message. The new interaction metrics use the same path, with rate bounds of 0 to 1 and latency bounded below at 0. Covered by `test_replay_inputs_cannot_smuggle_impossible_scores` and `test_out_of_range_interaction_rate_is_rejected`.

Separate from the review, and as @borgr noted above, this branch predates #218 and is many commits behind `main`, so the four files still need replanting from `utils/tau_bench/` into `every_eval_ever/adapters/tau_bench/` before it can land. These fixes sit on top of the current branch.

borgr:
Review threads addressed. Each concern maps to a fix and a regression test.

- Pass^k identity — metric_id is `tau_bench.pass_hat_k` and the semantics field states that all k trials must succeed. Covered by `test_pass_metric_names_pass_hat_k_and_says_so`.
- Score validation — replay inputs reject booleans, NaN, infinity, out-of-range Pass^k, and negative cost. Covered by `test_replay_inputs_cannot_smuggle_impossible_scores`.
- Malformed methodology — `isinstance` guards on `methodology`, `voice_config`, `model_release`, and `trajectory_files`. Covered by `test_a_malformed_methodology_block_does_not_crash_the_submission`.
- `--limit` downloads — the manifest is sliced before any submission is fetched, so `--limit` bounds the download in URL mode. Covered by `test_limit_bounds_the_download_not_just_the_output`.
- Provenance — remote mode pins the commit sha and records the payload bytes, and a local replay records `local_path` and claims no upstream URL. Covered by `test_remote_run_pins_the_ref_and_records_the_bytes` and `test_local_replay_does_not_claim_an_upstream_url`.
- Developer identity — `Z.ai` and `Zhipu AI` both resolve to `zhipu-ai` through a normalized-name lookup. Covered by `test_one_provider_spelled_two_ways_gets_one_developer`.

The threads that anchor to `utils/tau_bench/` predate the relocation into `every_eval_ever/adapters/tau_bench/` and are superseded by the tests above.

PR #136, opened 2026-05-16
Fix LLM Stats evaluator provenance

tommasocerruti:
Fixes #119 for the `llm-stats` adapter.

This updates the adapter so `evaluator_relationship` is inferred from the relationship between the underlying score source/evaluator and the model developer, rather than from LLM Stats as the aggregator.

Paired datastore PR: https://huggingface.co/datasets/evaleval/EEE_datastore/discussions/137

## Behavior

- `self_reported=true` or source organization matching the model developer => `first_party`
- `self_reported=false` or source organization differing from the model developer => `third_party`
- no usable provenance signal => `other`
- LLM Stats remains the aggregator in `source_metadata`
- raw provenance fields are preserved in `score_details.details`

### Third-party example: MiniMax M2.7 on GDPVal-AA

Current datastore:
https://huggingface.co/datasets/evaleval/EEE_datastore/raw/main/data/llm-stats/minimax/minimax-m2.7/fa480e85-428c-473d-8c8c-222e74f66155.json

New datastore PR file:
https://huggingface.co/datasets/evaleval/EEE_datastore/raw/refs%2Fpr%2F137/data/llm-stats/minimax/minimax-m2.7/7b294459-ebb0-4ff9-b64f-597d94ce2a9d.json

Before:
- `evaluator_relationship = other`
- no Artificial Analysis source provenance

After:
- `evaluator_relationship = third_party`
- preserves `raw_self_reported=false`
- preserves `raw_self_reported_source=https://artificialanalysis.ai/evaluations/gdpval-aa`
- infers `raw_source_organization=artificial-analysis`

### First-party example: MiniMax M2.7 self-reported scores

Current datastore:
https://huggingface.co/datasets/evaleval/EEE_datastore/raw/main/data/llm-stats/minimax/minimax-m2.7/c035d7f9-f489-4f53-a044-f796cee1471b.json

New datastore PR file:
https://huggingface.co/datasets/evaleval/EEE_datastore/raw/refs%2Fpr%2F137/data/llm-stats/minimax/minimax-m2.7/46d7bc7e-87e3-4bab-8dc5-8f92603eeb16.json

Before:
- `evaluator_relationship = first_party`
- no original MiniMax source URL preserved

After:
- still `evaluator_relationship = first_party`
- preserves `raw_self_reported=true`
- preserves `raw_self_reported_source=https://www.minimax.io/models/text/m27`
- infers `raw_source_organization=minimax`

## Validation

```bash
[...]

PR #194, opened 2026-06-29
Move validator business logic into every_eval_ever and update to de-duplication logic

nelaturuharsha:
- Summary: Single source of truth for validation rules added to cli, and dedup operations now are part of the cli. PR to merge is up. README is officially out of date.
- Changes:
     - Warning when score_type, max_score and min_score are not mentioned.
     - score should be between [min_score, max_score]
     - If HF model is provided, it should exist - wellness check for both model and dataset.
     - Will now check to ensure mandatory metadata under additional details for model_info i.e.
     - deployment_type and model_availability are present. Adapters need to be updated to fix this.
    - new commands/functionality added to cli
    - Users will have to submit deployment_type and model_availability in additional_details which the bot will enforce.
     - Three options: api, local, unknown
         - if api: model_availability can be -> closed_source, open_weights_deployment, other
         - if local: model_availability can be. -> hf, unavailable, other
         - if hf -> wellness check will ensure it exists
         
         
 Ruff/Ty were run which formatted some adapters.


every_eval_ever/dedup.py
```diff
@@ -0,0 +1,306 @@
...
+def check_duplicates(
+    file_paths: list[str],
+    local_paths: dict[str, str | Path],
+    manifest: dict[str, Any],
+) -> DedupReport:
+    """Compute fingerprints for aggregate JSON files and compare to manifest."""
+    file_fingerprints: dict[str, str] = {}
+    warnings: list[str] = []
+    for file_path in sorted(file_paths):
+        if not file_path.endswith('.json'):
```
Erotemic: This is a footgun. If you are expecting this to be called with only json paths, and you ever hit this case, you are silently dropping information about some bad pre-filtering done when constructing the content passed to this function. I would either remove this continue or raise an error.

every_eval_ever/dedup.py
```diff
@@ -0,0 +1,306 @@
...
+
+def build_dedup_report(
+    file_fingerprints: dict[str, str],
+    manifest: dict[str, Any],
+) -> DedupReport:
+    """Compare candidate fingerprints against manifest and same-batch files."""
+    validate_manifest(manifest)
+    report = DedupReport()
+    manifest_files = manifest.get('files', {})
+    fingerprint_to_path: dict[tuple[str, str], str] = {
```
Erotemic: Some strange type coercion here. This would be better solved by ensuring the inputs to this function are already normalized. From what I understand by just a static reading, this is already the case, and there are no callers outside this file, so this is mostly excessive defensive coding and ultimately bloat. Could simplify the logic a lot here.

every_eval_ever/dedup.py
```diff
@@ -0,0 +1,306 @@
...
+
+@dataclass
+class DedupReport:
+    """Aggregated deduplication report."""
+
+    results: list[DedupResult] = field(default_factory=list)
+    warnings: list[str] = field(default_factory=list)
+
+    @property
+    def has_duplicates(self) -> bool:
```
Erotemic: Is this ever used?

When I use a property I typically expect that it's not going to do any serious number crunching every time I access it (let's me reason about big-oh better).

I would just remove it or make it a `functools.cached_property` if it is needed. I might also change the type of `DedupReport` to make results and warnings typed as `Sequence`, which means that you will get a type error if something tries to mutate them after the fact. That will require that you build the contents and construct the `DedupReport` instead of an incremental build in later code, which I think would be a good thing.

every_eval_ever/dedup.py
```diff
@@ -0,0 +1,306 @@
...
+        key=lambda item: json.dumps(item, sort_keys=True),
+    )
+    payload = json.dumps(canonical, sort_keys=True, ensure_ascii=True).encode()
+    return hashlib.sha256(payload).hexdigest()
+
+
+def compute_fingerprint(content: bytes) -> str:
+    """Compute the semantic duplicate fingerprint for aggregate JSON bytes."""
+    try:
+        data = json.loads(content)
```
Erotemic: This might end up being a bottleneck. It may be worth it to abstract the specific json reader / writer so the stdlib json is always an option, but you can opt-in to something like `orjson` or `ujson`. (Note: orjson is a bit opinionated. It's fast, but it also means you can't serialize generic floats like nans or infs, which technically isn't json, but the stdlib and ujson aren't so strict about it).

I've handled this in a few ways in the past: simple wrappers: https://gitlab.kitware.com/computer-vision/kwutil/-/blob/main/kwutil/util_json.py?blame=1&ref_type=heads#L391

And module level globals: https://gitlab.kitware.com/computer-vision/kwcoco/-/blob/main/kwcoco/coco_dataset.py?blame=1&ref_type=heads#L117

every_eval_ever/dedup.py
```diff
@@ -0,0 +1,306 @@
...
+
+def _quantize(value: Any) -> Any:
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, (int, float)):
+        return repr(round(float(value), _QUANT_DECIMALS))
+    return value
+
+
+def _canon(obj: Any) -> Any:
```
Erotemic, line 86: The recursive implementation could probably be much faster using an iterator and maintaining your own stack: https://gitlab.kitware.com/computer-vision/kwutil/-/blob/main/kwutil/util_json.py?ref_type=heads#L29

A utility I wrote makes this fairly easy to express without too much extra complication: https://github.com/Erotemic/ubelt/blob/main/ubelt/util_indexable.py#L52

every_eval_ever/dedup.py
```diff
@@ -0,0 +1,306 @@
...
+    """Hash the normalized semantic identity of an aggregate eval record."""
+    model_info = data.get('model_info') or {}
+    eval_library = data.get('eval_library') or {}
+    raw_results = data.get('evaluation_results') or []
+    identity = {
+        'model': _norm_str(model_info.get('id')),
+        'lib': _norm_str(eval_library.get('name')),
+        'results': [
+            result
+            for result in (_result_identity(item) for item in raw_results)
```
Erotemic: I would prefilter the `None` values so you can remove the check from `_result_identity` and type its input as a dict. Gives you less conditions to care about in a hot loop. I think it would also make the code a bit more readable.

every_eval_ever/dedup.py
```diff
@@ -0,0 +1,306 @@
...
+        return None
+    text = _WS_RE.sub(' ', str(value).strip().lower())
+    return text.strip('"\'.,;:!?()[]{} ') or None
+
+
+def _quantize(value: Any) -> Any:
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, (int, float)):
+        return repr(round(float(value), _QUANT_DECIMALS))
```
Erotemic: Why are we turning ints into floats? Feels wrong on a fundamental level. If there are ints in a float field, coerce them to float for type normalization, but int fields should be allowed to be ints IMO. They are safe to hash. 

You can get a 2x speed bump with something along the lines of:


```python
_QUANT_DECIMALS = 6
_QUANT_FORMAT = f'.{_QUANT_DECIMALS}f'


def _quantize(value: Any) -> Any:
    if isinstance(value, bool):
        return value
    if isinstance(value, int):
        return str(value)
    if isinstance(value, float):
        text = format(value, _QUANT_FORMAT).rstrip('0').rstrip('.')
        return '0' if text in {'', '-0'} else text
    return value
```

And avoid weird canonical forms.

utils/multi_swe_bench/adapter.py
```diff
@@ -40,12 +40,17 @@
...
+    get_model_id,
+    save_evaluation_log,
+)
 from utils.swe_helpers import parse_date_from_dir, parse_model_from_dir
 
-MULTI_SWE_REPO = "https://github.com/multi-swe-bench/experiments"
-LANGUAGES = ["c", "c++", "go", "java", "javascript", "rust", "typescript"]
-OUTPUT_BASE = "data/multi-swe-bench-leaderboard"
+MULTI_SWE_REPO = 'https://github.com/multi-swe-bench/experiments'
+LANGUAGES = ['c', 'c++', 'go', 'java', 'javascript', 'rust', 'typescript']
```
Erotemic, line 55: Having some canonical formatter like `ruff` as a pre or post commit stage would help prevent this style churn. I don't like to enforce pre-commits, but I think doing so periodically as normalization commits is a good approach when projects are still gaining momentum.

every_eval_ever/dedup.py
```diff
@@ -0,0 +1,306 @@
...
+
+class ManifestError(RuntimeError):
+    """Raised when manifest.json cannot be safely loaded or used."""
+
+
+def _norm_str(value: Any) -> str | None:
+    if value is None:
+        return None
+    text = _WS_RE.sub(' ', str(value).strip().lower())
+    return text.strip('"\'.,;:!?()[]{} ') or None
```
Erotemic: Probably don't need regex here unless we are expecting to handle unicode whitespace edge cases. Otherwise, this code is faster:

```python
_TRIM_CHARS = '"\'.,;:!?()[]{} '

def _norm_str(value: Any) -> str | None:
    if value is None:
        return None
    text = ' '.join(str(value).lower().split())
    return text.strip(_TRIM_CHARS) or None
```

Erotemic (commented):
Mostly just a python-level review, I didn't consider anything holistically.

every_eval_ever/dedup.py
nelaturuharsha: Updated with your suggested code.

every_eval_ever/dedup.py
nelaturuharsha: The goal was more or less to make them comparable for dedup purposes. But I see your point, addressed!

every_eval_ever/dedup.py
nelaturuharsha: Think we're just going fingerprints that we created and not the JSON, would this still be a big performance gap you think?

every_eval_ever/dedup.py
nelaturuharsha: Removed at this point.

every_eval_ever/dedup.py
nelaturuharsha: I had this in place because we don't have a good recipe to reliably de-duplicate jsonls at the moment and we have only jsons or jsonls in the target datastore.

every_eval_ever/dedup.py
nelaturuharsha: Done!

every_eval_ever/dedup.py
Erotemic: I'd have to profile it to know, but recursion can get pretty costly. I just tend to avoid it when writing new code. Might be fine the way it is.

FWIW: a quick pass with sample data on both variants with [line-profiler](https://github.com/pyutils/line_profiler) will tell you a lot.

.gitignore
```diff
@@ -223,4 +223,5 @@ vendor/
 _site/
 .sass-cache/
 .jekyll-cache/
-.jekyll-metadata
\ No newline at end of file
+.jekyll-metadataUSAGE_*
```
yananlong: This was a typo?

every_eval_ever/check_duplicate_entries.py
```diff
@@ -81,59 +83,29 @@ def main(argv: List[str] | None = None) -> int:
...
-            }
-        )
-
-    duplicate_groups = [
-        entries for entries in groups.values() if len(entries) > 1
+    local_paths = {file_path: file_path for file_path in file_paths}
+    dedup_report = check_duplicates(file_paths, local_paths, {'files': {}})
+    duplicate_results = [
+        result for result in dedup_report.results if result.duplicate_of
     ]
```
yananlong: This looks like it only does local same-batch duplicate detection now. Since the manifest is always `{'files': {}}`, this path can’t catch “this candidate already exists in the accepted datastore,” even though the docs say the PR bot compares candidates against accepted records. Could we either load/pass the real manifest here, or make this command/docs explicit that this is local-only? Also, since file_paths are local CLI paths, we probably want to normalize them through repo_path_from_path; otherwise ./data/... or absolute paths won’t scope correctly by data/<collection>.

every_eval_ever/dedup.py
```diff
@@ -0,0 +1,320 @@
...
+        return {}
+    identity = {
+        'type': _norm_str(source_data.get('source_type')),
+        'name': _norm_str(source_data.get('dataset_name')),
+    }
+    if source_data.get('source_type') == 'hf_dataset':
+        repo = _norm_str(source_data.get('hf_repo'))
+        if repo:
+            identity['hf'] = repo
+    return identity
```
yananlong, line 253: This seems too loose for non-HF sources. For url or leaderboard-style sources, two different datasets with the same `dataset_name` would collapse to the same source identity because the URL/source artifact/version is ignored. That could create false duplicate hits across different leaderboards or dataset releases. Could we include a normalized URL / source artifact / version field when present, not just `dataset_name`?

every_eval_ever/validation_core.py
```diff
@@ -0,0 +1,727 @@
...
+    expected = {
+        str(folder / f'{uuid}.jsonl'),
+        str(folder / f'{uuid}_samples.jsonl'),
+    }
+    if not any(path in available_files for path in expected):
+        return [
+            f"Companion .jsonl for '{Path(repo_path).name}' not found "
+            'in the dataset or this PR'
+        ]
+    return []
```
yananlong: This ignores the actual `detailed_evaluation_results.file_path` and just infers the companion path from the aggregate UUID. That can pass when the referenced file is wrong but a same-UUID JSONL exists, and fail when the aggregate intentionally points to a different relative path. Could we validate the referenced path directly, maybe with same-folder UUID aliases as a fallback?

every_eval_ever/validation_core.py
```diff
@@ -0,0 +1,727 @@
...
+            and isinstance(lo, (int, float))
+            and not isinstance(lo, bool)
+            and isinstance(hi, (int, float))
+            and not isinstance(hi, bool)
+            and (score < lo or score > hi)
+        ):
+            warnings.append(
+                f'evaluation_results[{index}]: score {score} is outside '
+                f'[min_score={lo}, max_score={hi}]'
+            )
```
yananlong, line 328: This only warns when the key is absent, but fails silently when they are present but null or NaN, and the bound checks for `lo` and `hi` are also silently skipped. If these fields are now expected, can we treat None/blank the same as missing?

every_eval_ever/validation_core.py
```diff
@@ -0,0 +1,1139 @@
...
+    return None
+
+
+def check_integer_counts(data: dict[str, Any]) -> list[str]:
+    """Warn when count fields are present but not plain integers."""
+    warnings: list[str] = []
+
+    def walk(obj: Any, path: str) -> None:
+        if isinstance(obj, dict):
+            for key, value in obj.items():
```
j-chim, line 464:
```suggestion
            for key, value in obj.items():
                if key == 'additional_details':
                    continue  # schema types all additional_details into str, so nothing here can be int
```

j-chim: This check walks into `additional_details`, which is stringified because of the schema. Currently if any `_COUNT_FIELDS` are matched under `additional_details`, it will raise errors.

Because of this:
- The HELM adapter's output fails validation
- It looks like cocoabench hit the same collision and the code worked around it by renaming the key to evaluated_dataset_size.

This may be fixed in the next schema version but for the time being, suggest skipping `additional_details` in the walk. Alternatively, if the intent is that count keys must never appear in `additional_details`, that rule should be explicit and the HELM converter/cocoabench updated consistently.

every_eval_ever/converters/helm/adapter.py
j-chim, line 1: `utils/helm/adapter.py` contains expected changes (new required fields: inference_engine, deployment_type/model_availability in additional_details, metric_id via helm_metric_identity) but they're missing here. As it stands this means it fails a bunch of tests on the mmlu fixture (~100 errors).

Is this meant to be updated too?

utils/cocoabench/adapter.py
```diff
@@ -319,30 +341,32 @@ def make_source_data(
     benchmark_reference_urls: list[str],
 ) -> SourceDataPrivate | SourceDataUrl:
     additional_details = {
-        "samples_number": str(answered),
-        "benchmark_version": benchmark_version,
+        'evaluated_dataset_size': str(answered),
```
j-chim, line 344: Was this rename a workaround for `check_integer_counts()` rejecting stringified `samples_number` in `additional_details`? If the walk is fixed to skip `additional_details` (see my comment on `validation_core.py`), this could keep the conventional key name.

every_eval_ever/validation_core.py
```diff
@@ -0,0 +1,1139 @@
...
+def _hf_dataset_exists(
+    api: Any, repo_id: str
+) -> tuple[bool | None, str | None]:
+    """Return (exists, error); exists is None when verification failed."""
+    key = ('dataset', repo_id)
+    if key in _existence_cache:
+        return _existence_cache[key]
+    try:
+        api.dataset_info(repo_id)
+    except RepositoryNotFoundError:
```
j-chim, line 491: If we want to distinguish between gated repo vs genuine not found, I suggest we raise GatedRepoError here first (if applicable) and handle downstream accordingly. 

Functionally it's still data not publicly accessible but feels like a useful distinction to make.

every_eval_ever/validation_core.py
```diff
@@ -0,0 +1,1139 @@
...
+    """Blocking and advisory findings produced by registered checks."""
+
+    errors: list[dict[str, Any]] = field(default_factory=list)
+    warnings: list[dict[str, Any]] = field(default_factory=list)
+
+
+def repo_path_from_path(path: Path) -> str:
+    """Best-effort repo-relative path for local CLI use.
+
+    If an absolute local path contains a ``data`` component, warnings should use
```
j-chim, line 106: No warning here; first-data match means that this will break when new users have local paths that already contain ``data''. 

This matters downstream when we use this derived path to check_duplicate_entries, validate_many, check_path_structure etc

Suggest either warn per docstring or use explicit CLI argument for repo root path

every_eval_ever/validation_core.py
```diff
@@ -0,0 +1,1139 @@
...
+        return warnings
+
+    for index, result in enumerate(results):
+        if not isinstance(result, dict):
+            continue
+        metric = result.get('metric_config')
+        if not isinstance(metric, dict):
+            continue
+        score_type = metric.get('score_type')
+        if not isinstance(score_type, str) or not score_type.strip():
```
j-chim, line 291: This requires `score_type`, `min_score`, and `max_score `for every metric at error severity, but the published schema and the pydantic model both require bounds only when score_type=continuous (and don't require score_type at all). 

This means someone submitting a binary/levels metric will be blocked with ''missing or invalid 'min_score'" despite adhering to the (current) schema. Suggest to update the check here if we're waiting for the next schema version to land metric-related changes.

every_eval_ever/converters/lm_eval/utils.py
```diff
@@ -71,6 +74,6 @@ def find_samples_file(output_dir: Path, task_name: str) -> Optional[Path]:
     'rouge2': (0.0, 1.0),
     'rougeL': (0.0, 1.0),
     'rougeLsum': (0.0, 1.0),
-    'ter': (0.0, None),
+    'ter': (0.0, float('inf')),
```
j-chim, line 77:
```suggestion
    'ter': (0.0, float('inf')),
    'word_perplexity': (1.0, float('inf')),
    'byte_perplexity': (1.0, float('inf')),
    'perplexity': (1.0, float('inf')),
    'bits_per_byte': (0.0, float('inf')),
```

every_eval_ever/adapters/hal/provenance.py
```diff
@@ -0,0 +1,58 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+EXTERNALLY_MANAGED = 'externally_managed'
+UNKNOWN = 'unknown'
+OPEN_WEIGHTS = 'open_weights'
+CLOSED_WEIGHTS = 'closed_weights'
+
+OPEN_MODEL_IDS = frozenset(
```
j-chim, line 12: Nit: internal inconsistency as this misses meta-llama/llama-4-maverick, which exists in `utils/hal/adapter.py`

j-chim: (I dont think there's actually maverick data upstream, but if there is this can cause an error)

utils/llm_stats/adapter.py
```diff
@@ -762,20 +906,178 @@ def make_model_details(model: dict[str, Any]) -> dict[str, str]:
...
+        'ai.google.dev': 'google',
+        'x.ai': 'xai',
+        'artificialanalysis.ai': 'artificial-analysis',
+    }
+    if host in domain_overrides:
+        return domain_overrides[host]
+
+    parts = host.split('.')
+    if len(parts) >= 2:
+        return normalize_slug(parts[-2])
```
j-chim, line 944: `source_organization_from_url()` returns the wrong organization for some URL shapes, e.g. github.com/openai/simple-evals → returns "github"; huggingface.co/deepseek-ai/DeepSeek-R1 → returns "huggingface". On these hosts the organization is the first path segment, not the domain.

Since this value feeds evaluator_relationship (for now!), a wrong organization can flip a score between first-party and third-party. 

Suggest expanding `domain_overrides` + adding path-based branches for GH and HF.

utils/scripts/run_adapters.py
```diff
@@ -0,0 +1,1126 @@
...
+    logs: list[dict[str, Any]] = []
+    for spec in contract.commands:
+        command = command_for(contract, spec, workspace)
+        result = subprocess.run(
+            command,
+            cwd=workspace,
+            env=env,
+            capture_output=True,
+            text=True,
+            timeout=contract.timeout_s,
```
j-chim, line 529: Is TimeoutExpired intentionally left uncaught, so a timeout aborts the whole run? Other failure types will fail the specific adapter and continue with the rest.

utils/scripts/run_adapters.py
```diff
@@ -0,0 +1,1126 @@
...
+    api: HfApi,
+    manifest: dict[str, Any],
+    prs: dict[str, Any],
+) -> dict[str, Any]:
+    augmented = json.loads(json.dumps(manifest))
+    manifest_files = augmented['files']
+    for collection, discussion in sorted(prs.items()):
+        revision = f'refs/pr/{discussion.num}'
+        prefix = f'data/{collection}/'
+        repo_files = api.list_repo_files(
```
j-chim, line 699: `list_repo_files(revision=refs/pr/N)`  returns every file at that ref, aka the whole datastore rather than files added by the PR

utils/scripts/run_adapters.py
```diff
@@ -0,0 +1,1126 @@
...
+            repo_id=DATASET_REPO_ID,
+            repo_type=DATASET_REPO_TYPE,
+            revision=revision,
+        )
+        pending = [
+            path
+            for path in repo_files
+            if path.startswith(prefix)
+            and path.endswith('.json')
+            and path not in manifest_files
```
j-chim, line 709: (Following comment above on `list_repo_files()`)

The `path not in manifest_files` filter is what's supposed to remove files already in the datastore. This works in enforced mode, where the manifest lists the full datastore. But in deferred mode the manifest starts empty, so nothing is removed. In effect every `data/<collection>/*.json` at the PR ref gets downloaded and fingerprinted, one file at a time, on every run.

Implications: 
1. Cost: a cron run with an open PR re-downloads the whole collection each time
2. Fingerprinting runs on old base records too rather than just freshly fetched ones. If there were changes in fingerprinting requirements in between old vs fresh runs, one such mismatched base file will fail the run + subsequent runs until the PR is closed / rebased (as a fix on main doesn't change what's at `refs/pr/N`). 

Suggest restricting the loop to files the PR actually added (diff against its base commit). If there are issues with fingerprinting legacy records we should still surface them, but not here.

j-chim (changes requested):
Overall looks good, left some inline comments.

Would be great to address especially the ones on `run_adapters.py` and `validation_core.py` - thanks!

General questions/comments more for future PRs:
1. Adapter failure policy can be more consistent.
  For example there are ones that (a) raise at ingestion and abort the run (helm, mercor, exgentic, swe_polybench, hle, hal), (b) fall back to explicit unknowns (llm_stats, openeval), and (c) infer a value from the benchmark's documented setup (rewardbench defaults to open_weights since RewardBench runs models locally). Suggest picking one contract (e.g. fail fast everywhere) or documenting explicitly why each adapter is setup that way.
2. Dedup - I presume things like what goes into fingerprinting will be revisited after schema changes. 
3. Feels like there are some redundancies in variables (e.g. defining UNKNOWN, OPEN_WEIGHTS, CLOSED_WEIGHTS) but I presume this is for future PR implementing a base adapter class? Also I think we'll move some model logic and metrics data logic into registry when we get there?

utils/llm_stats/adapter.py
j-chim: (See https://github.com/evaleval/every_eval_ever/pull/136 for suggested implementation)

nelaturuharsha:
Closing this in favour of other changes, before modifications move to the validator.

PR #204, opened 2026-07-20
Add Open Medical-LLM Leaderboard adapter

borgr:
## What / source

Converts the [Open Medical-LLM Leaderboard](https://huggingface.co/spaces/openlifescienceai/open_medical_llm_leaderboard)
into aggregate EEE records under `data/open-medical-llm/`.

Source is the HF dataset `openlifescienceai/results`, laid out as
`<developer>/<model>/results_*.json` in lm-evaluation-harness output format. One
record per model, one result per medical benchmark (MedMCQA, MedQA USMLE 4-options,
PubMedQA, and six MMLU medical subjects). Aggregates only.

Adapter lives at [`every_eval_ever/adapters/open_medical_llm/`](every_eval_ever/adapters/open_medical_llm/)
(rebased onto the post-#218 layout; `utils/` no longer exists).

## Review lane

- [ ] **Fast**
- [x] **Needs a human** — `metric_id` moves from nine per-benchmark ids
      (`open-medical-llm-leaderboard.<task>.accuracy`) to the global `accuracy`, and
      a path/config model-name disagreement is now reconciled through HF's alias
      redirect instead of taking the path. Both change the identifiers a maintainer
      joins on, even though no score value changes. A file that converts to nothing
      also changed from a silent skip to a non-zero exit.

Design agreed in: n/a — the mapping calls are answered in the review thread below.

## Checklist
- [x] `python -m every_eval_ever validate` clean at
      `data/open-medical-llm/<dev>/<model>/` — 180 files, **0 invalid, 0 errors,
      0 warnings**, exit 0
- [x] every unconvertible source file lands in
      `adapter_reports/open-medical-llm_failures.json` with a non-zero exit
      (0 on the current snapshot). The report is written **before** publication, so
      a publication error cannot take it down with it, and on **every** run — a
      clean one included — swapped in atomically, so no earlier run's report can be
      read as this one's and no interrupted write can truncate a complete one.
- [x] offline unit tests (24, in
      [`tests/test_open_medical_llm_adapter.py`](tests/test_open_medical_llm_adapter.py));
      full `pytest tests` green (435 passed, 20 skipped)
- [x] `ruff check` clean
- [ ] model ids resolve in the registry — **143 of 180 do** (132 `exact` at
      confidence 1.0, 11 `normalized` at 0.95). 37 need registry attention and the
      run names them; see below.
- [x] content spot-checked: `evaluation_id` stable across re-runs and keyed on the
      result file's own path and timestamp — not `now`, and not the reconciled repo
      or the registry id, either of which can move under a fixed source file;
      benchmark `source_data` points at each
      benchmark's own HF dataset repo rather than the results repo; scores are
      leaderboard aggregates, so no answer leak and nothing double-counted

## Decisions & coverage

- Decision / where: model identity, `evaluated_model_repo`.
  Chose: use the dataset path when it agrees with the run config; when they
  disagree, resolve **both** through HuggingFace's alias redirect and use the repo
  they agree on. If they resolve to different repos, report the row as
  unconvertible.
  Instead of: preferring the config (the reviewer's suggestion) or the path
  (the previous behaviour).
  Why: on the live snapshot exactly one of 180 models diverges —
  `aaditya/OpenBioLLM-Llama3-70B` (path) vs `aaditya/OpenBioLLMLlama-70B`
  (`model_args.pretrained`) — and **both are HF aliases of
  `aaditya/Llama3-OpenBioLLM-70B`**. So neither is a typo for a different model,
  and a fixed preference would have been right by luck. Resolving the alias gives
  a better id than either spelling, and it happens to be the one the registry
  matches exactly. The reconciled repo drives routing, `model_info.id` and the model
  metadata, but **not** `evaluation_id`: HF alias metadata is upstream state that can
  change, and an id that moves with it would re-ingest one source file as a second
  evaluation. That id keys on the `developer/model` path the file sits under, so on
  this snapshot the divergent record routes to
  `data/open-medical-llm/aaditya/Llama3-OpenBioLLM-70B/` with
  `evaluation_id: open-medical-llm-leaderboard/aaditya_OpenBioLLM-Llama3-70B/1714089773.142451`,
  and both original spellings stay in provenance.
  Confidence: high. General? **yes** — any adapter deriving identity from a repo
  path has this, and the alias redirect is a cheap general answer.

- Decision / where: `metric_id`, `make_result`.
  Chose: `accuracy`.
  Instead of: `open-medical-llm-leaderboard.<task>.accuracy`.
  Why: the skill is explicit — "minting `<src>.accuracy` fragments the one join
  that makes the datastore worth querying", and keeping two benchmarks apart is
  `evaluation_name`'s job. Bounds `[0,1]` already match the registry's accuracy
  family.
  Confidence: high. General? no (the rule is already written down).

- Decision / where: file accounting, `convert`.
  Chose: a selected result file that yields no record is a **failure**; the five
  root-level hand-curated baselines are **exclusions**.
  Why: the file was chosen as a model's latest leaderboard result, so an empty
  conversion means the export is incomplete and must not exit 0. The baselines are
  a documented policy skip — paper numbers for closed models (bare `acc`, no
  `acc_stderr`/`model_args`), not harness runs — so they are recorded but do not
  fail the run.
  Confidence: high. General? yes (the failure/exclusion split is already the shared
  helper's shape).

- Decision / where: when the accounting report is written, `write_conversion_report`.
  Chose: unconditionally, atomically, and **before** publication.
  Instead of: after successful publication (this round's suggestion), or only when
  there are failures (the previous behaviour).
  Why: the report accounts for the *conversion*, so a publication error is exactly
  when it is most worth having — an earlier round asked for that ordering and
  `test_failure_report_survives_a_publication_error` pins it. What was actually
  broken is orthogonal to the ordering: the write was conditional on failures, so a
  run with only exclusions, or a clean one, silently left the previous run's report
  in place to be read as its own. Writing every run fixes that, and the atomic swap
  gives the "without losing the last complete ledger" property the fix path asked
  for regardless of when it happens.
  Confidence: high. General? yes — a stale ledger reads as current in any adapter
  that writes one.

- Decision / where: republication, `existing_records`.
  Chose: refuse to write into a directory that already holds records for these
  models unless `--replace-existing` is passed, and with that flag remove the prior
  records only once this run's are on disk.
  Why: filenames are fresh uuid4s, so a second run added a duplicate of every
  `evaluation_id` rather than replacing it. This is the hazard `lexam`'s README
  already documents. Unlinking first would trade the duplicate for a worse failure —
  a validation or write error would leave the target holding neither copy — and
  `save_evaluation_logs` already preflights the whole batch and rolls back what it
  created, so publish-then-remove leaves one complete set at every moment.
  Confidence: high. General? **yes** — it belongs in `save_evaluation_logs` for
  every adapter, but that is a cross-package change, so it is adapter-local here
  and proposed as a follow-up rather than slipped in.

- Decision / where: `deployment_type` / `model_availability`.
  Chose: `self_deployed` / `open_weights` when `model_args.pretrained` is present
  (164 of 180), placeholders otherwise.
  Why: `pretrained=<hf repo>` is the harness loading a checkpoint locally, which
  evidences both axes. The 16 rows with an empty `model_args` are also open-weights
  HF repos, but the file does not say so, so no claim is made.
  Confidence: high for the 164 · deliberately silent for the 16. General? yes for
  any lm-eval-harness-derived source.

**Coverage:** 209 result files → 180 models selected (latest file per model) + 5
excluded baselines → **180 records**, 0 failed. The 24 non-latest files for models
with several runs are superseded by design, not dropped silently — `latest_per_model`
picks the newest and the chosen file is recorded in
`source_metadata.additional_details.source_result_file`. No caps, no sampling.
Re-verified live on this head: 180 records, `validate` **all 180 passed**, 0
failures, the 5 baselines itemized as exclusions in the report on a clean run.

**Operator asked about policy calls?** Yes: the model ids that need registry
attention are printed by the run rather than auto-created, so minting canonical ids
stays a maintainer decision. On this snapshot that is 37 — 36 `no_match` at
confidence 0.0, plus 1 the resolver could not be reached for (transient, so that
last count varies per run). Nothing else needed a policy call — no large data drop and no re-hosting.


Erotemic:
GPT 5.6 Review (3m 24s)

## Findings

### High — Resolve model identity from the result payload, not the directory path

`make_log()` always sets `ModelInfo.id` from `model_repo`, while independently setting `ModelInfo.name` from `config.model_name`.

Those values are not reliably equivalent upstream. For example:

* Result path: `aaditya/OpenBioLLM-Llama3-70B/...`
* `config.model_name`: `aaditya/OpenBioLLMLlama-70B`
* `model_args`: `pretrained=aaditya/OpenBioLLMLlama-70B,...`

The adapter would therefore attach this evaluation to one Hugging Face ID while naming a different model. ([[Hugging Face](https://huggingface.co/datasets/openlifescienceai/results/commit/41787b5fe9adb6d587b7136cce21be0dcab19300?utm_source=chatgpt.com)][1])

This is silent model misattribution. Prefer the explicit evaluated model from `config.model_name` or parsed `model_args`; compare it against the directory path and either reject mismatches or handle them through an audited override. Add this real record as a regression fixture.

---

### High — These records are evaluation runs, not re-hosted documentation

The adapter marks every retained record as:

```python
source_type="documentation"
source_role="aggregator"
```

However, the upstream leaderboard explicitly says submitted models are automatically evaluated and that it evaluates them using the EleutherAI LM Evaluation Harness. It separately identifies GPT-4, Med-PaLM-2, and Gemini as paper-derived baselines—which this adapter already skips. ([[Hugging Face](https://huggingface.co/spaces/openlifescienceai/open_medical_llm_leaderboard/blob/main/src/about.py?utm_source=chatgpt.com)][2])

For the remaining records, use `source_type="evaluation_run"` and describe Open Life Science AI as the evaluator. `evaluator_relationship=third_party` remains appropriate. The harness is also directly documented upstream, so `eval_library` does not need to claim that it was merely inferred.

The synthetic test currently hard-codes the incorrect provenance and would prevent the correction.

---

### Medium — Follow pagination when listing the result repository

`list_result_files()` performs one raw request to the recursive tree endpoint and assumes the returned JSON is the entire repository.

Hugging Face’s tree endpoint is paginated, with 1,000 entries per page by default. Pagination is transparent only when using `huggingface_hub.list_repo_tree`; it is not handled by this direct `urlopen()` call. ([[Hugging Face](https://huggingface.co/docs/huggingface_hub/package_reference/hf_api)][3])

Once the repository crosses that boundary, the adapter will silently omit models. Use `list_repo_tree(REPO, repo_type="dataset", recursive=True)` or explicitly follow the API’s pagination links.

---

### Medium — A malformed result aborts the entire import

The worker catches exceptions only around `fetch_json()`. `make_log()` runs outside that `try`, so a bad score, invalid record shape, malformed model path, or Pydantic validation error escapes through `Executor.map()` and terminates the complete run instead of incrementing `errors`.

Wrap both fetching and conversion in the per-model exception boundary. Add a test containing a malformed numeric metric and assert that subsequent models are still processed. The existing tests only exercise valid synthetic records and a missing metric.

---

### Medium — Preserve fractional seconds in run timestamps

The upstream filenames contain microsecond timestamps, but `TS_RE` parses only whole seconds. Distinct evaluations produced during the same second therefore receive the same timestamp token and potentially the same `evaluation_id`.

The test even supplies `.123` but expects it to be discarded.

Parse the entire fractional component, ideally with `datetime.strptime()` using both supported filename separators, and derive the ID from the complete timestamp or the complete source path.

## Status

The current head is mergeable, and both the test workflow and documentation deployment passed. There are no existing PR comments.

**Recommendation: request changes**, primarily for the model-attribution and provenance defects. I did not submit a GitHub review.

[1]: https://huggingface.co/datasets/openlifescienceai/results/commit/41787b5fe9adb6d587b7136cce21be0dcab19300?utm_source=chatgpt.com "Create aaditya/OpenBioLLM-Llama3-70B/results_2024-04-26 00:02:53.142451.json · openlifescienceai/results at 41787b5"
[2]: https://huggingface.co/spaces/openlifescienceai/open_medical_llm_leaderboard/blob/main/src/about.py?utm_source=chatgpt.com "src/about.py · openlifescienceai/open_medical_llm_leaderboard at main"
[3]: https://huggingface.co/docs/huggingface_hub/package_reference/hf_api "HfApi Client · Hugging Face"

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The adapter is clear and focused, and three confirmed behavior edges would benefit from follow-up.
**Result:** One high-priority identity issue and two medium-priority error-path issues remain.
**Scope:** This PR maps Open Medical-LLM Leaderboard aggregate results into EEE logs.
**Checks:** 6 concerns checked; 2 confirmed, 1 narrowed, 0 ruled out, 3 lowered in priority.
**Second check:** targeted, 2 reviewers; 5 kept, 3 fix paths clarified, 0 set aside, 1 removed.

## Earlier review comments

None.

## What I noticed

- **[high] model identity** `utils/open_medical_llm/adapter.py:219` — A real leaderboard record has different path and run-config model IDs. The adapter uses the path for `ModelInfo.id`, evaluation identity, developer, and output grouping, so the scores can be attributed to a model other than the one evaluated. (`RAVF001`; inline)
- **[medium] partial failures** `utils/open_medical_llm/adapter.py:308` — Per-model fetch or JSON errors still allow a successful process exit. Automation can accept an incomplete export even though the adapter prints an error. (`RAVF003`; inline)
- **[medium] record isolation** `utils/open_medical_llm/adapter.py:286` — `make_log()` runs outside the worker error boundary. One malformed metric stops iteration before later valid records are saved. (`RAVF005`; inline)

<details>

- **[low] pagination** — The raw tree request could use Hugging Face's paginated client before the repository grows past one API page. Current truncation was not reproduced. (`RAVF004`)
- **[low] timestamps** — The parser could retain filename microseconds. Precision loss is confirmed, but no current same-model collision was found. (`RAVF006`)

</details>

## Things to try

- **[high] model identity** — A small identity helper can prefer a valid `pretrained` value, then `model_name`, then the path. The selected identity can drive every grouping field while the mismatch stays in metadata. (`RAVF001`)
- **[medium] error handling** — Fetch, conversion, and save failures can share one per-record accounting path. The module entry point can then return nonzero when any selected record fails. (`RAVF003`, `RAVF005`)

<details>

- Target: PR #204 (`add-open-medical-llm-adapter`, 5 files, +420/-0)
- Rounds: 1/1 completed; adaptive off; two reviewers returned empty output
- Mix: 3 codex-exec; 1 succeeded and 2 failed empty-output checks
- Focus: correctness, maintainability, simplicity, production impact, and constructive suggestion-oriented language
- Finding counts: 0 critical, 1 high, 2 medium, 2 low, 0 nit
- Checks: concerns=6; confirmed=2/narrowed=1/lowered=3
- Second check: targeted; reviewers=2; kept=5/clarified=3/set-aside=0/removed=1; approval changed no
- Set aside: 0 items

</details>

---

utils/open_medical_llm/adapter.py
```diff
@@ -0,0 +1,308 @@
...
+    else:
+        # No parseable timestamp in the filename: derive a STABLE token from the
+        # result file path (never `now`), so evaluation_id stays idempotent across
+        # re-runs. evaluation_timestamp is left None (the run time is unknown).
+        ts_token = re.sub(r"[^0-9A-Za-z._-]+", "-", path.rsplit("/", 1)[-1].removesuffix(".json"))
+    model_id = model_repo  # already an HF repo id developer/model (canonical per registry S1)
+
+    model_info = ModelInfo(
+        name=config.get("model_name") or model_repo,
+        id=model_id,
```
mrshu: **[high] model identity** — The path can name a different model than the run config

A real upstream result is stored under `aaditya/OpenBioLLM-Llama3-70B`, while `model_name` and `pretrained` name `aaditya/OpenBioLLMLlama-70B`. This line uses the path identity for `ModelInfo.id`; the same value also drives the evaluation ID and output grouping.

A small helper can select a valid `pretrained` value, then `model_name`, then the path. Recording any mismatch in metadata would keep the source traceable.

utils/open_medical_llm/adapter.py
```diff
@@ -0,0 +1,308 @@
...
+                skipped += 1
+            else:
+                errors += 1
+                print(f"  ERROR {model_repo}: {payload}")
+    print(f"Wrote {written} logs; skipped {skipped}; errors {errors}. -> {args.output_dir}")
+    return written
+
+
+if __name__ == "__main__":
+    main()
```
mrshu: **[medium] partial failures** — Per-model errors still leave the command at exit status zero

The worker turns fetch and JSON failures into `ERR`, and `main()` prints their count. The module entry point ignores `main()`'s return value, so automation can accept an incomplete export.

The entry point can propagate a status that is nonzero whenever any selected record fails. A partial-failure test would cover the command contract.

utils/open_medical_llm/adapter.py
```diff
@@ -0,0 +1,308 @@
...
+        f"Skipped {len(baselines)} hand-curated baseline entries (different provenance): "
+        + ", ".join(sorted(p.split('/')[0] for p in baselines))
+    )
+
+    def worker(model_repo: str):
+        try:
+            obj = fetch_json(chosen[model_repo])
+        except Exception as e:  # noqa: BLE001
+            return ("ERR", model_repo, str(e))
+        built = make_log(model_repo, obj, chosen[model_repo], retrieved_ts)
```
mrshu: **[medium] record isolation** — One malformed result stops later valid imports

`fetch_json()` is inside the worker error boundary, but `make_log()` is outside it. A bad metric value therefore escapes through `executor.map()` and prevents later valid records from being saved.

Fetch and conversion can share the per-record boundary. Write failures can use the same error count while iteration continues.

borgr:
Thanks for the review — pushed a commit addressing these:

- **`source_type` — kept `documentation` (a judgment call, flagging for a second opinion).** Reasoning: `source_type` is set by *the artifact you hold*, not who ran the compute. This adapter scrapes only-aggregate reported numbers (`acc,none` per benchmark, no per-item rows), so `documentation` fits even though a harness pipeline produced them; `evaluator_relationship=third_party` records who ran it. @nelaturuharsha — can you confirm this is the right call and consistent with how the datastore already treats leaderboard scrapes? **If that rule is *not* how we want it, then the `documentation` vs `evaluation_run` boundary is currently unclear and probably inconsistent across our existing data**, and we should pin it down. I've written the artifact-you-hold rule into the conversion skill, so whatever you confirm I'll align it there.
- **Model identity** — now resolves live against the eval-card-registry by default, using `canonical_id` for `model_info.id`. Added `--no-registry-resolve` opt-out (speed/offline); never fatal (falls back to the path id, marked unverified). Provenance recorded in `additional_details`; the run prints a review summary of auto-created-draft / low-confidence ids for a follow-up alias PR.
- **Idempotency** — `evaluation_id` keyed on the raw `developer/model` slug + eval time, deliberately *not* the resolved canonical id (which can move). Raw = record identity; canonical = join key.
- **Timestamps** — sub-second precision preserved, so same-second runs stay distinct.
- **Config vs path name** — record BOTH and flag `name_path_divergence` rather than silently trusting one.
- **Tree pagination** — follows the HF tree API `Link: rel="next"` header (a single GET truncated large repos).
- **Exit code** — non-zero on partial failure.
- **`eval_library.inferred_from`** — left as-is intentionally: a `documentation` source can still name a *known* harness from its result-key signature.

Erotemic:
LGTM, This is just one adapter, so even if there is a bug in it, it's going to be found and only apply to specific data. Merge risk here is low.

utils/open_medical_llm/adapter.py
```diff
@@ -0,0 +1,469 @@
...
+
+Data source (HuggingFace dataset): openlifescienceai/results
+  Layout: <developer>/<model>/results_*.json  (lm-evaluation-harness output format)
+  Backs the Space: openlifescienceai/open_medical_llm_leaderboard
+
+One EvaluationLog per model (developer/model), with one EvaluationResult per medical
+benchmark (accuracy, proportion 0-1, higher is better). Aggregates only; no per-item data.
+
+Run from the EEE repo dir:
+    uv run python -m utils.open_medical_llm.adapter --output-dir /tmp/eee-omll [--limit N]
```
mrshu: **RAV-RUN2-R1-F002 [medium] repository-integration** — This adapter remains under the retired `utils` layout. Installed packages include `every_eval_ever*`, so the adapter and its documented command are outside the active package.

Current main can be integrated before moving the adapter and references under `every_eval_ever/adapters/open_medical_llm`.

tests/test_open_medical_llm_adapter.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def test_make_log_is_valid_and_mapped():
+    # real result filenames are `results_<YYYY-MM-DD HH:MM:SS.micros>.json`
+    built = adapter.make_log("acme/med-x", _results_obj(),
+                             "acme/med-x/results_2024-05-01 00:00:00.123.json", "1700000000.0")
+    assert built is not None
+    log, developer, model = built
+    v = EvaluationLog.model_validate(log.model_dump())  # schema-valid
+    assert v.schema_version == "0.2.2"
```
mrshu: **RAV-RUN2-R1-F003 [medium] schema-compatibility** — This assertion requires schema `0.2.2`, while current main emits `0.3.0`. The test fails after integration.

The expected value can come from the imported schema version.

utils/open_medical_llm/adapter.py
```diff
@@ -0,0 +1,469 @@
...
+            dataset_name=display,
+            source_type="hf_dataset",
+            hf_repo=hf_repo,
+        ),
+        metric_config=MetricConfig(
+            evaluation_description=(
+                f"Accuracy on the {display} medical QA benchmark as reported by the "
+                "Open Medical-LLM Leaderboard."
+            ),
+            metric_id=f"{SRC}.{task}.accuracy",
```
mrshu: **RAV-RUN2-R1-F004 [medium] metric-identity** — Ordinary accuracy receives nine benchmark-specific metric IDs. This fragments the global accuracy join key.

`accuracy` can remain the metric ID while task-specific evaluation names preserve benchmark identity.

utils/open_medical_llm/adapter.py
```diff
@@ -0,0 +1,469 @@
...
+            return ("SKIP", model_repo, "no medical results", None)
+        return ("OK", model_repo, built, prov)
+
+    written = errors = skipped = 0
+    flagged: list[tuple[str, dict]] = []
+    with cf.ThreadPoolExecutor(max_workers=args.workers) as ex:
+        for status, model_repo, payload, prov in ex.map(worker, models):
+            if status == "OK":
+                log, developer, model = payload
+                save_evaluation_log(log, args.output_dir, developer, model)
```
mrshu: **RAV-RUN2-R1-F006 [medium] publication** — Repeat runs add a fresh UUID file for each stable evaluation ID. A populated target can hold duplicate logical records.

The collection target can reject existing output JSON before UUID generation or batch publication.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — A confirmed model-attribution defect remains when registry resolution is unavailable.
**Result:** One high and five medium concerns remain. Pagination, timestamp precision, and worker isolation are fixed.
**Scope:** This PR adds an Open Medical-LLM leaderboard adapter.
**Checks:** 6 concerns checked; 5 confirmed and 1 narrowed.
**Second check:** targeted, 2 reviewers; 6 kept, 4 fix paths clarified, 0 set aside, 0 removed.

## Earlier review comments

- **Still present:** model attribution can use the wrong path identity when the registry is unavailable. The existing thread remains open.
- **Partly fixed:** fetch and conversion failures now affect exit status. Empty results and publication failures remain outside full accounting.
- **Fixed:** malformed conversion isolation, pagination, and timestamp precision.

## What I noticed

- **RAV-RUN2-R1-F001 [high] model-identity** `utils/open_medical_llm/adapter.py:429` — When registry resolution is unavailable or disabled, a path/config mismatch makes `model_info.id` use the path identity instead of the evaluated config identity. (earlier comment remains open)
- **RAV-RUN2-R1-F002 [medium] repository-integration** `utils/open_medical_llm/adapter.py:12` — The adapter remains under the retired `utils` layout. Installed packages include `every_eval_ever*`, so this adapter and its documented command are outside the active package. (inline)
- **RAV-RUN2-R1-F003 [medium] schema-compatibility** `tests/test_open_medical_llm_adapter.py:26` — The test requires schema `0.2.2`, while current main emits `0.3.0`. The integrated test fails. (inline)
- **RAV-RUN2-R1-F004 [medium] metric-identity** `utils/open_medical_llm/adapter.py:274` — Ordinary accuracy receives nine benchmark-specific metric IDs. This fragments the global accuracy join key. (inline)
- **RAV-RUN2-R1-F005 [medium] error-handling** `utils/open_medical_llm/adapter.py:434` — A selected file with no usable score exits as a successful skip. A later publication failure can also bypass the failure report after earlier output exists. (earlier feedback remains partly open)
- **RAV-RUN2-R1-F006 [medium] publication** `utils/open_medical_llm/adapter.py:444` — Repeat runs add a fresh UUID file for each stable evaluation ID. A populated target can hold duplicate logical records. (inline)

<details>
<summary>Things to try (5 items)</summary>

- **[high] model-identity** — Existing lm-eval parsing rules can select one valid config identity. That value can drive model metadata, evaluation ID, and routing, while the dataset path and mismatches stay in provenance. (`RAV-RUN2-R1-P001`; covers `RAV-RUN2-R1-F001`)
- **[medium] integration** — Current main can be integrated before moving the adapter and references under `every_eval_ever/adapters/open_medical_llm`. Tests can use the imported schema version. (`RAV-RUN2-R1-P002`; covers `RAV-RUN2-R1-F002`, `RAV-RUN2-R1-F003`)
- **[medium] metric-identity** — `accuracy` can remain the metric ID while task-specific evaluation names preserve benchmark identity. (`RAV-RUN2-R1-P003`; covers `RAV-RUN2-R1-F004`)
- **[medium] error-handling** — Fetch, conversion, and empty-result failures can be recorded by selected source path. Valid records can publish once as a batch; publication failure can roll back and exit nonzero. (`RAV-RUN2-R1-P004`; covers `RAV-RUN2-R1-F005`)
- **[medium] publication** — The collection target can reject existing output JSON before UUID generation or batch publication. A repeat-run test can confirm that no file appears. (`RAV-RUN2-R1-P005`; covers `RAV-RUN2-R1-F006`)

</details>

<details>

- Target: PR #204 at `8d9be4041f42774a40d733f15c459e8b93f929fd` (5 files, +650/-0)
- Run ordinal: 2
- Mix: 3 codex-exec
- Earlier review comments: 10 ledger entries; fixed and carry-forward items revalidated
- Finding counts: 0 critical, 1 high, 5 medium, 0 low, 0 nit
- Checks: concerns=6; confirmed=5/ruled-out=0/set-aside=0/narrowed=1
- Second check: targeted; reviewers=2; kept=6/clarified=4/set-aside=0/removed=0; approval changed no

</details>

---

borgr:
Thanks — all six findings are addressed, two of them differently from the suggestion. Head is `e0c62afad`, on top of a merge of `main` that relocates the adapter to `every_eval_ever/adapters/open_medical_llm/` for the post-#218 layout.

Everything below was measured against a full live run: 209 result files → 180 models + 5 excluded baselines → 180 records, `validate` reporting 0 invalid / 0 errors / **0 warnings**, `pytest tests` 406 passed / 20 skipped, `ruff check` clean.

**F001 — model identity (path vs run config).** Fixed, but not by preferring the config.

I probed all 180 models first, and exactly **one** diverges: path `aaditya/OpenBioLLM-Llama3-70B` vs `model_args.pretrained=aaditya/OpenBioLLMLlama-70B`. Both are HuggingFace **aliases of the same repo**, `aaditya/Llama3-OpenBioLLM-70B`. So on this source no scores are actually misattributed, and a fixed preference for either side would have been right by luck rather than by reasoning — next time the typo could be on the config side.

`evaluated_model_repo` now uses the path when the two agree, and when they disagree resolves **both** through HF's alias redirect: if they land on one repo that repo is used (`model_identity_source=hf_alias`), and if they land on different repos the row is reported as unconvertible, because which model was evaluated is genuinely not recoverable from the export. Offline (`--no-registry-resolve`) or an unreachable API also refuses to guess. This is strictly better than either preference — the reconciled canonical is the id that resolves `exact` at confidence 1.0 with `review_status=reviewed`, where the raw path spelling did not.

Per P001, that one identity then drives model metadata, routing, `model_info.id` and `evaluation_id` together (`data/open-medical-llm/aaditya/Llama3-OpenBioLLM-70B/`, `evaluation_id: open-medical-llm-leaderboard/aaditya_Llama3-OpenBioLLM-70B/1714089773.142451`), and **both** original spellings stay in provenance as `model_identity_dataset_path` and `model_identity_run_config` alongside `model_identity_source`.

**F002 — retired `utils/` layout.** Fixed. `main` is merged in and the adapter now lives at `every_eval_ever/adapters/open_medical_llm/`, inside the installed package. The documented command is `uv run python -m every_eval_ever.adapters.open_medical_llm.adapter`, and the adapter table in `every_eval_ever/adapters/README.md` has a row for it.

**F003 — `SCHEMA_VERSION`.** Fixed. The test asserted the literal `"0.2.2"`; it now imports `SCHEMA_VERSION`, so it tracks main (`0.3.0` today) instead of pinning a version.

**F004 — `metric_id`.** Fixed to `accuracy`. The skill's field guide is explicit that minting `<src>.accuracy` "fragments the one join that makes the datastore worth querying", and that keeping two benchmarks apart is `evaluation_name`'s job. The nine `evaluation_name`s are unchanged, so nothing collapses; bounds `[0,1]` already match the registry accuracy family.

**F005 — every selected file accounted for.** Fixed, both clauses.

Conversion moved into `convert()`, which returns a `SourceConversionResult`: a selected file that yields no record is now a **failure** (it was chosen as that model's latest leaderboard result, so an empty conversion means the export is incomplete), and the 5 root-level hand-curated baselines are **exclusions** that do not fail the run. Failures land in `adapter_reports/open-medical-llm_failures.json` and `raise_if_incomplete()` makes the command exit non-zero. `make_log` is now inside the worker's `try`, so a malformed record can no longer escape `ex.map()` and abort the whole run.

On the second clause — you're right, and my first pass still had it: I'd written the report *after* `save_evaluation_logs`, so a publication error would have taken the record of what failed to convert down with it. The report accounts for the conversion, not the publication, so it is now written before publication is attempted, and `test_failure_report_survives_a_publication_error` pins that by making `save_evaluation_logs` raise and asserting the report is on disk anyway.

**F006 — republication.** Fixed adapter-locally, deliberately not in the shared helper.

Record filenames are fresh uuid4s, so a second run added a duplicate of every `evaluation_id` rather than replacing it (the hazard `lexam`'s README already documents). `existing_records()` + `--replace-existing` now refuse to write into a populated target for these models. I agree the right home is `save_evaluation_logs` / `_prepare_evaluation_logs`, so that *every* adapter gets it — but that is a cross-package behaviour change and AGENTS.md asks for agreement before something structural, so I'm offering it as a follow-up PR rather than slipping it into an adapter PR. Say the word and I'll open it.

**One thing the review didn't raise.** `deployment_type` and `model_availability` were inheriting the validator's `unknown` placeholders, which the datastore gate treats as a hard error even though the validator emits no warning. `model_args.pretrained=<hf repo>` is lm-evaluation-harness loading a checkpoint locally, which evidences both axes, so they are now set from it for 164 of 180 records. The 16 rows with an empty `model_args` are also open-weights HF repos, but their file doesn't say so, so they keep the placeholders rather than have the claim asserted from a repo name.

Registry status for the maintainers: 143 of 180 ids resolve (132 `exact` at 1.0, 11 `normalized` at 0.95). 37 need attention — 36 `no_match`, plus 1 the resolver couldn't be reached for. The run prints them by name; nothing is auto-created.

every_eval_ever/adapters/open_medical_llm/adapter.py
```diff
@@ -0,0 +1,635 @@
...
+        args.output_dir, [(developer, model) for _, developer, model in result.records]
+    )
+    if stale and not args.replace_existing:
+        raise SystemExit(
+            f'{len(stale)} record(s) are already published under '
+            f'{args.output_dir} for these models, e.g. {stale[0]}. Record '
+            'filenames are fresh uuid4s, so writing now would add a second copy '
+            'of each evaluation_id. Pass --replace-existing to replace them.'
+        )
+    for path in stale:
```
mrshu: **RAV-RUN3-R1-F001 [high] publication** — `--replace-existing` deletes prior records before replacement preflight

Lines 601–602 remove the old files before `save_evaluation_logs()` validates and writes the new batch. Its rollback covers only files created by that call, so a later failure can leave the target without its prior records.

The replacement path could retain recoverable old files until every new write and route transition succeeds. A mid-write failure test across several routes would cover this path.

every_eval_ever/adapters/open_medical_llm/adapter.py
```diff
@@ -0,0 +1,635 @@
...
+        # keep sub-second precision so two runs in the same whole second differ
+        ts_token = f"{base}.{eval_dt.microsecond:06d}" if eval_dt.microsecond else str(base)
+    else:
+        # No parseable timestamp in the filename: derive a STABLE token from the
+        # result file path (never `now`), so evaluation_id stays idempotent across
+        # re-runs. evaluation_timestamp is left None (the run time is unknown).
+        ts_token = re.sub(r"[^0-9A-Za-z._-]+", "-", path.rsplit("/", 1)[-1].removesuffix(".json"))
+
+    resolved_id = model_id or model_repo  # join key; raw_slug keys evaluation_id
+    raw_slug = model_repo.replace("/", "_")
```
mrshu: **RAV-RUN3-R1-F002 [high] evaluation-identity** — Alias metadata can change the identity of the same source file

`evaluated_model_repo()` returns the current Hugging Face canonical repository. `make_log()` then includes that value in `evaluation_id`. A later redirect change can give the same source result a second logical identity.

The raw dataset model path can remain the evaluation identity input. The reconciled repository can still drive model metadata and routing.

every_eval_ever/adapters/open_medical_llm/adapter.py
```diff
@@ -0,0 +1,635 @@
...
+        print("  NOTE: --no-registry-resolve set; model_info.id is path-derived and NOT registry-verified.")
+
+    result, flagged = convert(
+        chosen, baselines, retrieved_ts,
+        resolve_enabled=resolve_enabled, workers=args.workers,
+    )
+
+    # Written before publication: it accounts for the conversion, so a publication
+    # that raises must not take the record of what failed to convert with it.
+    if result.failures:
```
mrshu: **RAV-RUN3-R1-F004 [medium] source-accounting** — Exclusions-only runs do not persist their accounting report

`failure_report()` includes exclusions, but this condition writes it only when failures exist. The five documented baseline exclusions therefore have no durable record, and an older failure report can remain current.

An atomic report replacement after successful publication can record failures, exclusions, and fully clean runs without losing the last complete ledger.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The new adapter is well structured, and five focused suggestions can make identity and publication safer.
**Result:** Two high-priority and two medium-priority concerns remain. One CLI edge case is a low-priority suggestion.
**Scope:** Add the Open Medical-LLM adapter with stable identities, complete source accounting, isolated failures, and safe publication.
**Checks:** 5 concerns checked; 4 confirmed and 1 narrowed.
**Second check:** 2 reviewers checked 5 findings; 4 kept, 1 lowered in priority, and 3 fix paths clarified.

## Earlier review comments

<details>
<summary>Earlier review comments (22 ledger entries, grouped into 10 topics)</summary>

- The model-attribution mismatch is fixed. Alias reconciliation now rejects unresolved conflicts. The stable evaluation identity concern below is a separate edge of that approach.
- Partial-failure exit handling, repository placement, schema compatibility, metric identity, pagination, and timestamp precision are fixed.
- Duplicate detection and empty-result accounting are fixed. The replacement and exclusions-only paths below are separate follow-up suggestions.
- Record isolation is still incomplete for non-finite numbers. The current path now fails during batch serialization instead of per-source conversion.

</details>

## What I noticed

|---|---|---|---|---|
| RAV-RUN3-R1-F001 | high | publication | `every_eval_ever/adapters/open_medical_llm/adapter.py:601` | `--replace-existing` deletes prior records before replacement preflight. A later validation or write failure can leave the target without its prior records. |
| RAV-RUN3-R1-F002 | high | evaluation identity | `every_eval_ever/adapters/open_medical_llm/adapter.py:397` | Alias reconciliation uses the current Hugging Face canonical repository in `evaluation_id`. The same source file can receive a different identity after alias metadata changes. |
| RAV-RUN3-R1-F003 | medium | record isolation | `every_eval_ever/adapters/open_medical_llm/adapter.py:301` | Non-finite scores pass conversion but fail strict batch serialization after worker accounting. One malformed record can block all valid records. |
| RAV-RUN3-R1-F004 | medium | source accounting | `every_eval_ever/adapters/open_medical_llm/adapter.py:585` | A run with exclusions but no failures does not persist its accounting report. A report from an earlier failed run also remains unchanged. |

<details>

- **RAV-RUN3-R1-F005 [low] CLI** `every_eval_ever/adapters/open_medical_llm/adapter.py:568` — `--limit 0` processes all models. Negative limits use Python negative slicing.

</details>

## Things to try

<details>
<summary>Suggested changes (5 items)</summary>

- **[high] publication** — The replacement path could preflight all new records and retain recoverable old files until every new write succeeds. Failure tests across several model routes would cover the rollback. (`RAV-RUN3-R1-P001`; covers `RAV-RUN3-R1-F001`)
- **[high] evaluation identity** — The original dataset model path can remain the evaluation identity input. The reconciled repository can continue to drive model metadata and routing. (`RAV-RUN3-R1-P002`; covers `RAV-RUN3-R1-F002`)
- **[medium] record isolation** — The existing `require_finite_number` helper can validate accuracy and uncertainty inside `make_result`. The worker boundary will then account for the selected source file. (`RAV-RUN3-R1-P003`; covers `RAV-RUN3-R1-F003`)
- **[medium] source accounting** — An atomic report replacement after successful publication can record failures, exclusions, and fully clean runs without losing the last complete ledger. (`RAV-RUN3-R1-P004`; covers `RAV-RUN3-R1-F004`)
- **[low] CLI** — Negative limits can be rejected, and slicing can run whenever the value is not `None`. (`RAV-RUN3-R1-P005`; covers `RAV-RUN3-R1-F005`)

</details>

<details>

- Target: PR #204 at `d6e7288a56d1976a255b70c42269a48f804508f7` (5 files, +949/-0)
- Run ordinal: 3
- Mix: 3 codex-exec
- Earlier review comments: 22 ledger entries grouped into 10 topics; 1 still present and 9 fixed or replaced by distinct concerns
- Finding counts: 0 critical, 2 high, 2 medium, 1 low, 0 nit
- Checks: concerns=5; confirmed=4; narrowed=1; set-aside=0
- Second check: targeted; reviewers=2; kept=5; clarified=3 plans; lowered=1; removed=0; approval unchanged
- Set aside: 0 items

</details>

---

borgr:
All five findings are fixed in `e3e5630a7`, four of them exactly along the fix path
and one deliberately not (`F004`, below — the divergence is about *when*, and it
argues the ordering an earlier round of this review asked for).

Re-verified live on this head: 209 source files → 180 records, `validate` **all 180
passed**, 0 failures, 5 baselines itemized as exclusions. `pytest tests` **413 passed,
20 skipped** (this adapter's file 17 → 24 tests); `ruff check` clean.

### RAV-RUN3-R1-F001 · publication — replacement deleted before it replaced

Correct, and the worse of the two failures it trades against: without
`--replace-existing` a re-run duplicated every record, and with it a validation or
write error left the target holding *neither* copy.

The unlink loop moved after `save_evaluation_logs`, which is where the preflight your
fix path asks for already lives — it converts, validates and serializes the whole
batch before creating a single file, and removes the files it created if a later write
fails. So the prior records are the recoverable copy until the new ones exist, and the
tree holds one complete set at every moment: the previous publication if this run
dies, this one if it succeeds. `--replace-existing`'s help text no longer says
"delete … before writing", which is what it used to do.

Two tests, and the failure one covers several routes as asked:
`test_a_successful_replacement_does_not_accumulate_copies` (one record before, one
after, a new uuid4 filename, the same `evaluation_id`) and
`test_a_failed_replacement_leaves_the_previous_records_in_place` (two models across
two `developer/model` routes, publication made to raise mid-run; the file listing and
contents afterwards are identical to before it ran).

### RAV-RUN3-R1-F002 · evaluation identity — the id moved with upstream alias metadata

Also correct. The alias reconciliation I added last round fixed the misattribution but
made the record's identity depend on HF metadata that can change under a source file
that has not: `aaditya/OpenBioLLM-Llama3-70B` currently redirects to
`aaditya/Llama3-OpenBioLLM-70B`, and if that ever stops or resolves elsewhere, the
same file re-ingests as a second evaluation.

Fixed as P002 describes — the dataset path is the identity input, the reconciled repo
keeps everything else. `make_log` takes `dataset_repo` explicitly rather than
inferring it, so the split is visible at the call site. On the live snapshot the one
divergent record now shows all three values at once:

| | |
|---|---|
| route | `data/open-medical-llm/aaditya/Llama3-OpenBioLLM-70B/` |
| `model_info.id` | `aaditya/Llama3-OpenBioLLM-70B` |
| `evaluation_id` | `open-medical-llm-leaderboard/aaditya_OpenBioLLM-Llama3-70B/1714089773.142451` |

with `model_identity_dataset_path`, `model_identity_run_config` and
`model_identity_source=hf_alias` in provenance, so the reconciliation stays auditable
from the record. `test_evaluation_id_follows_the_source_path_not_the_evaluated_repo`
pins routing and id moving independently.

### RAV-RUN3-R1-F003 · record isolation — non-finite scores failed the batch

Fixed with `require_finite_number` inside `make_result`, on both `acc,none` and
`acc_stderr,none`. The carry-forward note in your ledger is what made this worth
doing: last round's fix moved `make_log` inside the worker's `try`, which isolated
*conversion* errors, but a `NaN` converted fine and only failed later in the batch
serializer — after every worker had reported, so it took all 180 valid records with
it and named no source file. Now the failure happens where the task and the file are
still in hand, so the worker boundary attributes it:
`test_a_non_finite_score_is_attributed_to_its_own_file` asserts the good model still
converts and the failure reason names `pubmedqa acc,none`.

None of the 180 live records carry a non-finite value, so this is a guard, not a
repair of observed corruption.

### RAV-RUN3-R1-F004 · source accounting — one fix, one deliberate divergence

The two defects you name are both real and both fixed: the write was conditional on
`result.failures`, so a run with only exclusions — or a fully clean one — wrote
nothing and left the previous run's report standing to be read as its own. It is now
written on every run, and swapped in with `os.replace`, so an interrupted write cannot
leave a truncated report where a complete one was.

The one place I did not follow P004 is "after successful publication". The report
accounts for the *conversion*, not the publication, so a publication error is exactly
when it is most worth having on disk — and an earlier round of this review asked for
that ordering specifically (`e0c62afad`, pinned by
`test_failure_report_survives_a_publication_error`, which makes
`save_evaluation_logs` raise and asserts the report is there anyway). The ordering was
never what was broken here; the conditional write was. Writing unconditionally and
atomically gives the "without losing the last complete ledger" property your fix path
is after, and it holds wherever in the run it happens.
`test_the_accounting_report_replaces_the_previous_run_s_copy` seeds a stale report,
runs clean, and asserts the stale entry is gone and no partial file is left in the
directory. Happy to move it if you'd rather have publication as the gate.

### RAV-RUN3-R1-F005 · CLI

Both halves. Negative values are rejected *before* the source listing, so a mistyped
limit costs no requests, and the message says what a negative slice actually did
(dropped models off the end rather than taking the first N). Slicing now runs whenever
the value is given, so `--limit 0` selects nothing instead of being falsy and
converting all 180 — and an empty selection stops the run rather than reporting a
refresh that wrote nothing, which is the same call as the `--models`-matched-nothing
guard in #203. `test_a_negative_limit_is_rejected` and
`test_limit_zero_selects_nothing_rather_than_everything`; both also assert no output
directory was created.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This revision makes substantial progress and its
CI is green; the remaining items are focused suggestions around identity,
replacement, accounting, and source completeness.
**Result:** Most earlier concerns are fixed. One high-priority edge case and
five medium-priority follow-ups remain on the current head.
**Scope:** Adds the Open Medical-LLM Leaderboard aggregate adapter, its source
accounting and publication flow, offline tests, and contributor documentation.
**Checks:** 2 concerns checked; 2 confirmed, 0 ruled out, 0 set aside, 0 lowered
in priority.
**Second check:** targeted, 2 Codex reviewers; 6 findings kept, 4 fix paths
clarified, 0 set aside, 0 removed.

## Earlier review comments

<details>
<summary>Earlier review comments (12 unique items)</summary>

- Fixed: the path/config identity conflict, partial-failure exit status,
  per-record isolation, package placement, schema-version assertion, canonical
  accuracy metric, ordinary repeat-run duplicate guard, and limit boundaries.
- `RAV-RUN3-R1-F001` is narrower now: replacement no longer removes old files
  before new-file preflight, but cleanup failure can still leave a mixed set.
- `RAV-RUN3-R1-F002` is narrower now: the raw source path keeps
  `evaluation_id` stable, but an alias route change can bypass prior-record
  discovery.
- `RAV-RUN3-R1-F003` is narrower now: non-finite values fail within the worker,
  but finite values outside the declared range still pass conversion.
- `RAV-RUN3-R1-F004` is narrower now: reports are replaced atomically on every
  run, but superseded source files are not included in the ledger.

</details>

## What I noticed

|---|---|---|---|---|
| RAV-RUN3-R1-F002 | high | alias replacement | `every_eval_ever/adapters/open_medical_llm/adapter.py:554` | The raw source path now keeps the evaluation ID stable, which is a useful improvement. If the Hugging Face alias target later changes, prior-record discovery checks only the new route and can miss the same stable ID under its old route. |
| RAV-RUN3-R1-F003 | medium | score bounds | `every_eval_ever/adapters/open_medical_llm/adapter.py:307` | Finite accuracy values outside the declared `[0, 1]` scale pass conversion and publication model validation, then fail only in final-path semantic validation. |
| RAV-RUN3-R1-F004 | medium | source accounting | `every_eval_ever/adapters/open_medical_llm/adapter.py:274` | `latest_per_model` selects one file and drops the superseded paths, so `total_records` cannot reconcile with the complete source snapshot. |
| RAV-RUN3-R1-F001 | medium | replacement cleanup | `every_eval_ever/adapters/open_medical_llm/adapter.py:654` | New files are safely preflighted before old files are removed. A later unlink failure can still leave the new set beside all or part of the old set, without a restore path. |
| RAV-RUN4-R1-F001 | medium | canonical routing | `every_eval_ever/adapters/open_medical_llm/adapter.py:378` | Registry resolution updates `model_info.id`, while the datastore route stays keyed to the reconciled Hugging Face repository. When those differ, canonical-ID consumers can miss the record. (inline) |
| RAV-RUN4-R1-F002 | medium | benchmark completeness | `every_eval_ever/adapters/open_medical_llm/adapter.py:383` | A selected file can publish fewer than the documented nine benchmark results because missing tasks and missing `acc,none` values are skipped whenever at least one valid result remains. (inline) |

## Things to try

<details>
<summary>Things to try (6 suggestions)</summary>

- **[high] alias replacement** — Before publication, build one immutable
  collection-wide snapshot of prior records that match the incoming stable
  evaluation IDs or the selected models' superseded source paths. That snapshot
  can drive the same recoverable transaction as cleanup, without rescanning the
  newly written files. (`RAV-RUN4-R1-P001`; covers `RAV-RUN3-R1-F002`)
- **[medium] score bounds** — Keep the helpful per-source isolation by rejecting
  finite accuracy outside `[0, 1]` inside `make_result`, with an out-of-range
  fixture beside a valid model. (`RAV-RUN4-R1-P002`; covers
  `RAV-RUN3-R1-F003`)
- **[medium] source accounting** — Keep superseded paths grouped by model,
  include only the selected model scope in the ledger, and name each chosen
  replacement in its exclusion. An older publication can remain intact when a
  newer source file fails conversion. (`RAV-RUN4-R1-P003`; covers
  `RAV-RUN3-R1-F004`)
- **[medium] replacement cleanup** — Publish the preflighted batch, then move
  every snapshotted old file into a same-filesystem quarantine. If a move fails,
  the new paths can be removed and the moved originals restored; quarantine
  deletion can remain recoverable post-commit cleanup. (`RAV-RUN4-R1-P004`;
  covers `RAV-RUN3-R1-F001`)
- **[medium] canonical routing** — Derive the final route with the shared
  datastore identity helpers from the resolved canonical model ID, while
  keeping the source path and reconciled Hugging Face repository in provenance.
  (`RAV-RUN4-R1-P005`; covers `RAV-RUN4-R1-F001`)
- **[medium] benchmark completeness** — Use an all-or-nothing source-file rule:
  require all nine task dictionaries and their non-null `acc,none` values at the
  worker boundary, report the complete missing-task list once, and cover a
  partial file beside a valid file. (`RAV-RUN4-R1-P006`; covers
  `RAV-RUN4-R1-F002`)

</details>

<details>

- Target: PR #204 at `c6cec44a9d0835872578a261aed7fb65fef9014f`
  (5 files, +1139/-0)
- Run ordinal: 4
- Mix: 3 Codex reviewers requested; 2 completed, and 1 failed the output
  protocol after its allowed retry
- Focus: technical accuracy with constructive, positive suggestions
- Verification: exact-head GitHub Actions passed all four test matrices
- Earlier review comments: 12 unique concerns; 8 fixed and 4 narrower but still
  present
- Finding counts: 0 critical, 1 high, 5 medium, 0 low, 0 nit
- Checks: concerns=2; confirmed=2/ruled-out=0/set-aside=0/lowered=0
- Second check: targeted; reviewers=2; kept=6/clarified=4/set-aside=0/removed=0;
  approval changed no
- Set aside: 0 items

</details>

---

every_eval_ever/adapters/open_medical_llm/adapter.py
```diff
@@ -0,0 +1,683 @@
...
+    ``model_repo`` is the evaluated model as established by evaluated_model_repo, and
+    it drives the developer/model routing and the model metadata. ``model_id`` is the
+    registry-canonical id for ``model_info.id`` (the join key); pass ``None`` for
+    path-mode (id == source repo). ``dataset_repo`` is the ``developer/model`` path the
+    result file sits under, and it alone keys ``evaluation_id``: an HF alias redirect
+    can move ``model_repo`` and the registry can re-map a draft ``model_id``, so either
+    would hand the same source file a second identity on re-ingest. Offline unit tests
+    call this directly without a resolver, so it never touches the network.
+    """
+    developer, model = model_repo.split("/", 1)
```
mrshu, line 378: **RAV-RUN4-R1-F001 [medium] canonical routing** — Registry resolution updates `model_info.id`, while this route still comes from the reconciled Hugging Face repository. When those identities differ, canonical-ID consumers can miss the published record. The final route can use the shared datastore identity helpers, with both raw identities retained as provenance.

every_eval_ever/adapters/open_medical_llm/adapter.py
```diff
@@ -0,0 +1,683 @@
...
+    config = obj.get("config", {}) or {}
+    results = obj.get("results", {}) or {}
+
+    ev_results = []
+    for task in TASKS:
```
mrshu, line 383: **RAV-RUN4-R1-F002 [medium] benchmark completeness** — This loop skips missing tasks and tasks without `acc,none`, while the record succeeds if any one result remains. A partial source file can therefore publish fewer than the documented nine benchmarks without an accounting entry. Validation at the worker boundary, plus a partial-file fixture, would keep that omission visible.

borgr:
Reviewed all 12 unresolved automated review threads against the current head. Ten are dismissed — seven predate the `utils/open_medical_llm/` → `every_eval_ever/adapters/open_medical_llm/` move, and the other three are already handled (publication safety, evaluation-identity, source accounting). Resolving those.

Two threads I'm leaving open as author decisions: datastore directory routing by HF-repo vs. registry-canonical id (`adapter.py:379`), and whether a partial benchmark run (<9 tasks) should be flagged/accounted or allowed as a subset (`adapter.py:384`). Both are design calls; I'll settle them separately.

every_eval_ever/adapters/open_medical_llm/adapter.py
borgr: Fixed in bd7143dbc. `make_log` now derives the datastore route from the registry-canonical `model_info.id` via `datastore_path_components` (as mmlu_pro / openeval / hal do), instead of the raw HF repo path. When the resolver remapped the id, the on-disk route and `model_info.id` diverged and a canonical-id consumer missed the record. The raw repo stays as `source_model_repo` provenance and `evaluation_id` still keys on it. Test added asserting the route follows the canonical id after a remap.

every_eval_ever/adapters/open_medical_llm/adapter.py
borgr: Fixed in bd7143dbc. This is the error case, not an expected subset: every model's latest file is expected to carry all nine benchmarks (verified 180/180 upstream), so a shortfall means a task failed/was not reported upstream. The converted record is now kept, but a `SourceRecordFailure` is recorded for the shortfall, so the run reports it and exits non-zero rather than silently publishing a partial. A complete nine-benchmark file publishes clean. Tests added for both the partial (keep + raise) and complete (clean) cases.

PR #205, opened 2026-07-20
docs(utils): add 'Submitting a new adapter' checklist

borgr:
Adds a **Submitting a new adapter** section to `utils/README.md`.

Adapter output can be schema-valid and still misrepresent the data — several issues only surface in review, not in `validate`. This section gives adapter authors a pre-submit routine that mirrors CI (`ruff check`, `pytest tests`, `validate` on the output, plus an offline shape test using `llm_stats` as the reference), and a short self-review list of the mistakes `validate` can't catch:

- `input.raw` must not contain the model's answer (prompt = user/system turns only);
- `output.raw` is the full generation, not the extracted answer;
- `metric_name` (the metric) vs `evaluation_name` (the benchmark); no `metric_type`;
- `eval_library` is the harness, not the platform;
- `source_type` `evaluation_run` vs `documentation`;
- `continuous` score_type needs `min_score`/`max_score` (no unbounded; `inf` serializes to null);
- `additional_details` values must be strings;
- `model_info.id` carries no effort/mode/quant suffix.

Docs only; no code changes.

### DIFF
--- a/utils/README.md
+++ b/utils/README.md
@@ -6,6 +6,46 @@ One-off adapter scripts that fetch leaderboard data from external sources and co
 
 Each adapter is run with `uv run python -m utils.<name>.adapter`.
 
+## Submitting a new adapter
+
+Adapter output can be schema-valid and still misrepresent the data — review
+repeatedly catches records that pass `validate` but are wrong. Before opening the
+PR, mirror CI and self-check the record shape:
+
+```bash
+uv run ruff check                                   # E/F/I, line-length 80, py312
+uv run --locked pytest tests                        # add an offline test for your adapter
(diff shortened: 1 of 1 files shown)


borgr:
Withdrawing this one. On review the per-field self-review rules here are low-level converter-authoring knowledge, not something that belongs in the human-oriented adapter README — and this repo already has docs/eval-converters/, docs/contributing/, and a per-field docs/data-structure/schema.md for public guidance. We'll keep this field-level guidance in our converter's own workflow notes instead. Sorry for the noise; happy to propose a targeted docs/ change if maintainers want one.

PR #206, opened 2026-07-21
fix(lm_eval): don't build continuous metrics with a null bound

borgr:
### The bug

The `lm_eval` converter crashes on results containing `ter` — or any metric absent from its 15-entry bounds table.

`KNOWN_METRIC_BOUNDS` declares `'ter': (0.0, None)`, and for any metric not in the table `KNOWN_METRIC_BOUNDS.get(...)` returns `None`. In `adapter.py` the bounds were then used to build a `continuous` `MetricConfig` **unconditionally**:

```python
min_score = bounds[0] if bounds else None
max_score = bounds[1] if bounds else None
metric_config = MetricConfig(score_type=ScoreType.continuous, min_score=min_score, max_score=max_score, ...)
```

`continuous` requires numeric `min_score`/`max_score` (enforced by `validate_score_type_requirements`), so with a `None` bound the `MetricConfig(...)` constructor raises `score_type 'continuous' requires max_score` — the whole conversion fails. EEE has no unbounded `score_type`, and `float('inf')` is not a fix (pydantic serializes it to `null`, which fails the same validator on read-back, and `Infinity` is non-standard JSON).

### The fix

- **`ter` → `(0.0, 100.0)`**: TER (sacrebleu translation edit rate) is reported on a 0–100 scale, like `bleu` in the same table. It can exceed 100 in pathological cases, but 0–100 is the reported range and EEE does not range-check scores, so a finite nominal bound is correct here.
- **Skip metrics with no finite bound** (e.g. `perplexity`, `word_perplexity`) with a warning, instead of constructing an invalid `continuous` config — since EEE cannot represent an unbounded metric.
- **Drop a task whose every metric was skipped**, rather than emit a contentless `EvaluationLog`.
- **Regression test** (`test_unbounded_metrics_do_not_produce_invalid_records`): `ter` yields a valid 0–100 continuous record; a genuinely-unbounded metric is skipped; a perplexity-only task is dropped.

`uv run pytest tests` → 207 passed, 20 skipped. `ruff check` clean.

> Note: the deeper gap is that EEE has no `score_type` for genuinely unbounded metrics (elo, perplexity, dollars) — `continuous` mandates finite bounds. This PR stops the crash within the current schema; a first-class unbounded representation would be a separate schema discussion.

### DIFF
--- a/every_eval_ever/converters/lm_eval/adapter.py
+++ b/every_eval_ever/converters/lm_eval/adapter.py
@@ -1,6 +1,7 @@
 """Adapter for converting lm-evaluation-harness output to every_eval_ever format."""
 
 import json
+import logging
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 
@@ -37,6 +38,8 @@ from .utils import (
     MODEL_TYPE_TO_INFERENCE_PLATFORM,
     parse_model_args,
 )
+
+logger = logging.getLogger(__name__)
 
 
 class LMEvalAdapter(BaseEvaluationAdapter):
@@ -251,8 +254,19 @@ class LMEvalAdapter(BaseEvaluationAdapter):
             is_higher_better = higher_is_better.get(metric_name, True)
 
             bounds = KNOWN_METRIC_BOUNDS.get(metric_name)
-            min_score = bounds[0] if bounds else None
-            max_score = bounds[1] if bounds else None
+            if bounds is None or bounds[0] is None or bounds[1] is None:
+                # `continuous` requires numeric min/max and EEE has no unbounded
+                # score_type, so a metric with no finite bound (e.g. perplexity)
+                # cannot be represented. Skip it rather than emit an invalid
+                # record with a null bound.
+                logger.warning(
+                    'lm_eval: skipping metric %r in task %r — no finite bound '
+                    'for continuous score_type',
+                    metric_name,
+                    task_name,
+                )
+                continue
+            min_score, max_score = bounds
 
             description = metric_name
             if filter_name != 'none':
@@ -386,6 +400,14 @@ class LMEvalAdapter(BaseEvaluationAdapter):
         for task_name in tasks:
             task_metadata = {**metadata_args, 'task_name': task_name}
             log = self._transform_single(raw_data, task_metadata)
+            if not log.evaluation_results:
+                # Every metric for this task was skipped (no finite bound); an
+                # EvaluationLog with no results conveys nothing, so drop it.
+                logger.warning(
+                    'lm_eval: skipping task %r — no representable metrics',
+                    task_name,
+                )
+                continue
             results.append(log)
 
         return results
--- a/every_eval_ever/converters/lm_eval/utils.py
+++ b/every_eval_ever/converters/lm_eval/utils.py
@@ -71,6 +71,10 @@ KNOWN_METRIC_BOUNDS = {
     'rouge2': (0.0, 1.0),
     'rougeL': (0.0, 1.0),
     'rougeLsum': (0.0, 1.0),
-    'ter': (0.0, None),
+    # TER (sacrebleu translation edit rate) is reported on a 0-100 scale, like
+    # `bleu`. It can exceed 100 in pathological cases, but 0-100 is the reported
+    # range; EEE has no unbounded score_type and does not range-check scores, so
+    # a finite nominal bound is used rather than None (None -> invalid record).
+    'ter': (0.0, 100.0),
     'brier_score': (0.0, 1.0),
 }
--- a/tests/test_lm_eval_adapter.py
+++ b/tests/test_lm_eval_adapter.py
@@ -1,3 +1,4 @@
+import json
 import tempfile
 from pathlib import Path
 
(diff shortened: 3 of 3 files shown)


Erotemic:
GPT 5.6 Review (3m 32s):

## Review: request changes

### High — Missing bounds are being mistaken for unbounded metrics

In the new condition, `bounds is None` means only that the metric is absent from the local lookup table. It does **not** establish that the metric is unbounded. The branch consequently discards every metric outside the current 15-entry table.

This loses ordinary bounded `lm_eval` metrics. Upstream defines, for example:

* `acc_mutual_info` and `acc_bytes`, both accuracy metrics.
* `chrf`, which is returned as a SacreBLEU score.
* WMT configurations explicitly report `bleu`, `ter`, and `chrf`; this PR would silently remove `chrf` from every such conversion.

A task containing only an omitted-but-bounded metric is then removed entirely.

The schema already provides a non-lossy fallback: `MetricConfig.score_type` is optional, and the validator requires bounds only when `score_type` is explicitly `continuous`.

**Fix:** preserve metrics with unknown or unbounded ranges by leaving `score_type`, `min_score`, and `max_score` unset. Use `continuous` only when valid finite bounds are known. Also distinguish “explicitly known unbounded” from “not yet classified,” rather than encoding both as a missing dictionary entry.

Add regression cases for at least `chrf`, `acc_bytes`, and an arbitrary custom metric.

---

### High — `100` is not a valid maximum possible TER score

The schema defines `max_score` as the **maximum possible score**, but the patch explicitly installs a merely nominal maximum of 100.

SacreBLEU computes TER as:

```text
100 * total_edits / reference_length
```

There is no restriction requiring `total_edits <= reference_length`; insertion-heavy hypotheses can therefore produce TER greater than 100.

The absence of range validation does not make the metadata correct. A valid record could claim a score of 150 with `max_score=100`, which can break normalization or any consumer relying on the advertised range.

**Fix:** represent TER without `score_type` and bounds under the current schema, or extend the schema with a lower-bounded/unbounded continuous representation. Do not claim 100 is its maximum. The regression should include a TER score above 100 and verify serialization round-trips without contradictory bounds.

---

### Medium — Dropping the parent log prevents instance-level export

The new code removes a task as soon as all aggregate metrics are skipped.

However, `--include_samples` processes samples only by iterating over the logs returned from `transform_from_file`. A dropped task never reaches `LMEvalInstanceLevelAdapter`, so its detailed sample JSONL is also lost.

Those detailed records can still contain the per-sample score and complete `lm_eval_metrics` metadata, including metrics that the aggregate schema cannot classify.

**Fix:** do not discard the parent `EvaluationLog` in the aggregate adapter. Preserve the metric with an unspecified score type, which also eliminates the empty-log case. Add a CLI-level test showing that `--include_samples` still exports samples for a perplexity-only task.

## Status

Both GitHub Actions workflows—tests and documentation deployment—currently pass, and the PR has no prior review comments or submitted reviews.

I would block this PR primarily because it replaces a conversion crash with silent, potentially extensive data loss.

mrshu:
# ⚒️ review-anvil report

**Review decision:** COMMENT — This is comment-only, and two high-priority carry-forward findings remain.
**Result:** The crash guard is useful, but the current fallback can lose metric and sample data.
**Scope:** This PR avoids building continuous LM-Eval metrics when a bound is unavailable.

**Checks:** skipped; all findings had direct multi-reviewer code and caller evidence.
**Second check:** 2 reviewers checked 3 findings; all 3 kept and all fix paths clarified.

## Earlier review comments

All three items in the [earlier review](https://github.com/evaleval/every_eval_ever/pull/206#issuecomment-5121680754) remain present:

- **[high] metric conversion** — Missing table entries are treated as unbounded metrics, so bounded and custom results can disappear. (`RAVF001`)
- **[high] TER bounds** — TER still declares `100` as its maximum even though valid scores can exceed it. (`RAVF002`)
- **[medium] sample export** — Dropping the parent aggregate log also prevents instance-level sample export. (`RAVF003`)

## What I noticed

These carry-forward findings remain valid. No distinct new problems were confirmed.

## Things to try

- **[high] metric conversion** — The published JSON Schema can first accept an absent `score_type`. Unknown-range metrics can then remain present with type and bounds omitted. Dictionary membership can distinguish unknown names from explicit null bounds. (`RAVF001`)
- **[high] TER bounds** — TER can use the same schema-safe unknown-range form. A score above 100 can pass both Pydantic and published-schema validation. (`RAVF002`)
- **[medium] sample export** — Preserving each aggregate metric removes the empty-log cause. Both existing `include_samples` entry points can cover this path. (`RAVF003`)

<details>

- Target: PR #206 (`fix/lm-eval-unbounded-metric-bounds`, 3 files, +94/-3)
- Mix: 3 codex-exec
- Earlier review comments: 3 still present
- Finding counts: 0 critical, 2 high, 1 medium, 0 low, 0 nit
- Checks: skipped; multi-reviewer consensus with direct evidence
- Second check: targeted; reviewers=2; kept=3, clarified=3, set-aside=0, removed=0; approval changed=no

</details>

---

Erotemic:
> since EEE cannot represent an unbounded metric

Would a better fix for this to merge the "infinity" branch and allow EEE to represent and unbounded metric?

This PR seems like more of a hack than a solution.

Erotemic:
The answer is yes, the infinity branch is strictly better than this. I'm confident enough in this to make an executive decision to close this branch.

PR #207, opened 2026-07-21
fix: represent unbounded metric bounds as the JSON string "Infinity"

borgr:
## Problem
An unbounded continuous bound (PSNR, perplexity, TER, …) is `float('inf')`/`-inf`, but pydantic's default serialization wrote it to JSON `null` — which then failed the existing `continuous requires min/max_score` validator. So `inf` could be **read but not written**: unbounded metrics could only be dropped, faked, or made invalid.

## Fix
**Core** — a field serializer on `MetricConfig.min_score`/`max_score` emits `inf`/`-inf` as the JSON **string** `"Infinity"`/`"-Infinity"`:
- valid RFC-8259 JSON (parses in JS / Go / jq / orjson, not just Python) that pydantic reads back to `float('inf')`;
- chosen over model-level `ser_json_inf_nan` because that only affects `model_dump_json()` — the codebase also writes via `json.dump(model_dump(mode='json'))` (`cli.py`, instance writers), which would emit the bare, **invalid** `Infinity` token. `field_serializer(when_used='json')` covers **both** paths; `mode='python'` keeps the native float;
- the field stays typed `float | None`, so "convertible to a number" is enforced by the type; the schema pins the string form to exactly `{"Infinity","-Infinity"}`.

**Semantics (kept strict on purpose)**
- `null` stays **invalid** for a continuous metric: *missing ≠ unbounded*, so `null` cannot masquerade as `inf`; unbounded uses `inf`, genuinely-missing bounds stay invalid.
- `NaN` bounds are rejected (never meaningful; `NaN != NaN` breaks comparisons).

**Convention applied in-repo** (`converters/lm_eval`) so nothing is left emitting null-for-unbounded: `ter` bound `(0.0, None)` → `(0.0, inf)`, and a metric with no entry in `KNOWN_METRIC_BOUNDS` is now **skipped with a warning** instead of building an invalid null-bounded continuous config. **Supersedes #206.**

**Regen-safe:** `post_codegen.py` re-applies the validator + serializer to `MetricConfig`.

## Tests
`tests/test_inf_serialization.py` (both dump paths incl. the `mode='json'` regression, `mode='python'` float, strict-JSON validity, float typing/round-trip, null-still-invalid, NaN-rejected) + a new `lm_eval` case. Full suite green (207 passed / 20 skipped), ruff clean.

## Portability
`"Infinity"`/`"-Infinity"` is valid JSON everywhere. A statically-typed consumer (Go `float64`) needs a small custom unmarshaller for the `number | "Infinity"` union — the schema `enum` documents exactly what to expect.

## Follow-ups (separate — not required for correctness here)
- **`eval-card-registry` (separate repo):** reclassify the ~9 canonical metrics whose `max_score: null` means *unbounded* (`elo`, `cost-per-task`, `latency-*`, `rank`, …) to `inf`; leave the genuinely-vague ones (`score`, `*.mean` placeholders) `null`.
- **`EEE_datastore`:** re-run / patch affected records in parallel.

### DIFF
--- a/every_eval_ever/converters/lm_eval/adapter.py
+++ b/every_eval_ever/converters/lm_eval/adapter.py
@@ -1,6 +1,7 @@
 """Adapter for converting lm-evaluation-harness output to every_eval_ever format."""
 
 import json
+import warnings
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 
@@ -251,8 +252,18 @@ class LMEvalAdapter(BaseEvaluationAdapter):
             is_higher_better = higher_is_better.get(metric_name, True)
 
             bounds = KNOWN_METRIC_BOUNDS.get(metric_name)
-            min_score = bounds[0] if bounds else None
-            max_score = bounds[1] if bounds else None
+            if bounds is None:
+                # Unknown bounds: a continuous metric_config requires numeric
+                # min/max, so this would be invalid. Skip rather than emit a bad
+                # record. To include the metric, add it to KNOWN_METRIC_BOUNDS
+                # (use float('inf') for an unbounded side).
+                warnings.warn(
+                    f'lm_eval: skipping metric {metric_name!r} on task '
+                    f'{task_name!r} -- no known bounds in KNOWN_METRIC_BOUNDS',
+                    stacklevel=2,
+                )
+                continue
+            min_score, max_score = bounds
 
             description = metric_name
             if filter_name != 'none':
--- a/every_eval_ever/converters/lm_eval/adapter.py
+++ b/every_eval_ever/converters/lm_eval/adapter.py
@@ -48,10 +48,23 @@ class LMEvalAdapter(BaseEvaluationAdapter):
         # Stores per-log metadata so callers can find sample files after transform.
         # Keyed by evaluation_id -> {"parent_dir": str, "task_name": str}
         self._eval_metadata = {}
+        # (task_name, metric_name) pairs dropped for lack of known bounds, so a
+        # caller can report the loss instead of it being silent. Accumulates over
+        # the adapter's lifetime (one instance per convert run); see get_skipped_metrics.
+        self._skipped_metrics: List[tuple] = []
 
     def get_eval_metadata(self, evaluation_id: str) -> Dict[str, Any]:
         """Return stored metadata for a given evaluation_id."""
         return self._eval_metadata.get(evaluation_id, {})
+
+    def get_skipped_metrics(self) -> List[tuple]:
+        """(task_name, metric_name) pairs skipped for lack of known bounds.
+
+        Each entry is a metric that was silently dropped from the output because
+        KNOWN_METRIC_BOUNDS had no range for it (a continuous MetricConfig needs
+        finite/typed min+max). Callers surface this so the drop is visible.
+        """
+        return list(self._skipped_metrics)
 
     @property
     def metadata(self) -> AdapterMetadata:
@@ -257,6 +270,7 @@ class LMEvalAdapter(BaseEvaluationAdapter):
                 # min/max, so this would be invalid. Skip rather than emit a bad
                 # record. To include the metric, add it to KNOWN_METRIC_BOUNDS
                 # (use float('inf') for an unbounded side).
+                self._skipped_metrics.append((task_name, metric_name))
                 warnings.warn(
                     f'lm_eval: skipping metric {metric_name!r} on task '
                     f'{task_name!r} -- no known bounds in KNOWN_METRIC_BOUNDS',
--- a/every_eval_ever/eval_types.py
+++ b/every_eval_ever/eval_types.py
@@ -14,6 +14,7 @@ from pydantic import (
     Field,
     confloat,
     conint,
+    field_serializer,
     model_validator,
 )
 
@@ -428,10 +429,12 @@ class MetricConfig(BaseModel):
         description='Indicates whether there is an Unknown Level - if True, then a score of -1 will be treated as Unknown',
     )
     min_score: float | None = Field(
-        None, description='Minimum possible score for continuous metric'
+        None,
+        description='Minimum possible score for a continuous metric. Use -inf if unbounded below (serialized as the string "-Infinity"). null means "not provided", NOT unbounded.',
     )
     max_score: float | None = Field(
-        None, description='Maximum possible score for continuous metric'
+        None,
+        description='Maximum possible score for a continuous metric. Use inf if unbounded above, e.g. PSNR or perplexity (serialized as the string "Infinity"). null means "not provided", NOT unbounded.',
     )
     llm_scoring: LlmScoring | None = Field(
         None, description='Configuration when LLM is used as scorer/judge'
@@ -457,7 +460,27 @@ class MetricConfig(BaseModel):
                 raise ValueError("score_type 'continuous' requires min_score")
             if self.max_score is None:
                 raise ValueError("score_type 'continuous' requires max_score")
+        # A NaN bound is never meaningful (and NaN != NaN breaks comparisons).
+        # inf/-inf are allowed (unbounded); null is handled above per score_type.
+        for _name in ('min_score', 'max_score'):
+            _v = getattr(self, _name)
+            if _v is not None and _v != _v:
+                raise ValueError(f'{_name} must not be NaN')
         return self
+
+    @field_serializer('min_score', 'max_score', when_used='json')
+    def _serialize_bound(self, value):
+        # An unbounded bound is +/-inf; emit it as the JSON string
+        # "Infinity"/"-Infinity" (valid JSON that reads back to a float) rather
+        # than pydantic's default null or the non-standard bare Infinity token.
+        # when_used='json' covers BOTH model_dump_json() and
+        # model_dump(mode='json'); mode='python' keeps the native float. null
+        # stays null (reserved for "not provided", never unbounded).
+        if value == float('inf'):
+            return 'Infinity'
+        if value == float('-inf'):
+            return '-Infinity'
+        return value
 
 
 class EvaluationResult(BaseModel):
--- a/every_eval_ever/schemas/eval.schema.json
+++ b/every_eval_ever/schemas/eval.schema.json
@@ -219,13 +219,18 @@
                                 "description": "Indicates whether there is an Unknown Level - if True, then a score of -1 will be treated as Unknown"
                             },
                             "min_score": {
-                                "type": "number",
-                 
-                                "description": "Minimum possible score for continuous metric"
+                                "oneOf": [
+                                    { "type": "number" },
+                                    { "type": "string", "enum": ["Infinity", "-Infinity"] }
+                                ],
+                                "description": "Minimum possible score for a continuous metric. A number, or the string \"-Infinity\" if unbounded below (float('-inf'); reads back to a float). Absence/null means \"not provided\", NOT unbounded."
                             },
                             "max_score": {
-                                "type": "number",
-                                "description": "Maximum possible score for continuous metric"
+                                "oneOf": [
+                                    { "type": "number" },
+                                    { "type": "string", "enum": ["Infinity", "-Infinity"] }
+                                ],
+                                "description": "Maximum possible score for a continuous metric. A number, or the string \"Infinity\" if unbounded above, e.g. PSNR or perplexity (float('inf'); reads back to a float). Absence/null means \"not provided\", NOT unbounded."
                             },
                             "llm_scoring": {
                                 "type": "object",
--- a/post_codegen.py
+++ b/post_codegen.py
@@ -48,10 +48,10 @@ PATCHES = [
     },
     {
         'file': 'every_eval_ever/eval_types.py',
-        'import_add': 'model_validator',
+        'import_add': ['model_validator', 'field_serializer'],
         'class_name': 'MetricConfig',
         'validator': """
-    # --- validators (added by post_codegen.py) ---
+    # --- validators / serializers (added by post_codegen.py) ---
 
     @model_validator(mode="after")
     def validate_score_type_requirements(self):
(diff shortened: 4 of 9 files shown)


Erotemic:
GPT 5.6 Review (7m 11s)

## Review verdict: **Request changes**

CI is green, but the PR has several data-contract and converter correctness issues.

### High — The lm-eval adapter silently drops standard metrics

`_build_evaluation_results()` now skips every metric absent from `KNOWN_METRIC_BOUNDS`. That table omits several built-in lm-eval metrics, including `perplexity`, `word_perplexity`, `byte_perplexity`, `bits_per_byte`, `chrf`, `acc_mutual_info`, `acc_bytes`, and `acc_all`. As written, a normal lm-eval conversion can silently produce partial—or even empty—results, despite the PR specifically citing perplexity as a motivating case.

These are standard registered lm-eval metrics, not unknown third-party extensions.

**Fix:** add bounds for the built-in metric registry at minimum. For genuinely unknown metrics, fail explicitly or provide an opt-in skip policy rather than silently deleting evaluation data with a warning.

---

### High — The actual validator does not enforce the documented wire format

The schema permits numbers plus exactly `"Infinity"` and `"-Infinity"`, but repository validation does not validate against that schema. `validate_file()` calls `EvaluationLog.model_validate()`, and the fields remain coercive `float | None`. Pydantic therefore accepts other numeric strings such as `"inf"`, `"+inf"`, `"-inf"`, and `"1.5"`. Such a file passes the project validator but fails consumers validating against the published JSON Schema.

The new test only rejects a completely nonnumeric string, so it does not exercise this mismatch.

The bounds contract is also semantically under-enforced: both sentinel signs are allowed in both fields, and the model accepts `min_score=+inf`, `max_score=-inf`, or `min_score > max_score`.

**Fix:** add a `mode="before"` field validator that accepts only JSON numbers and the two exact sentinel strings, converts the sentinels to floats, and rejects all other strings and booleans. Add an after-validator enforcing `min_score <= max_score`, no positive-infinite minimum, and no negative-infinite maximum. Test the raw-file validation path, not only model serialization.

---

### High — This is a new wire contract still labeled schema `0.2.2`

Existing `0.2.2` consumers can reasonably generate `float64`/number-only models from the old schema. The PR begins emitting strings under that same schema version, and its own description acknowledges that typed consumers require a custom unmarshaller. Without a version change, consumers cannot determine which interpretation of `0.2.2` a record requires. The schema still declares version `0.2.2`.

**Fix:** coordinate a schema-version bump and make converters emit the new version before publishing string-valued bounds. Existing records can remain on `0.2.2`.

---

### Medium — Regeneration will not preserve the promised `float | None` annotation

The PR says the fields remain typed `float | None`, but the schema now declares a `oneOf` containing a number and a string enum. The repository’s generator processes `oneOf` as multiple combined data types, so the next regeneration will produce a union containing the generated sentinel enum rather than the checked-in `float | None` annotation.

`post_codegen.py` only appends validators and a serializer; it does not restore the annotation or otherwise normalize the generated public type.  The regeneration workflow runs after schema changes reach `main`, so current PR CI does not verify this claim.

**Fix:** run the exact regeneration pipeline before merging and commit or inspect its output. Either patch the generated annotation deliberately or document and test the resulting union type. Add a CI drift check that regenerates into a temporary location and compares the result.

I have not submitted this review to GitHub.

mrshu:
# ⚒️ review-anvil report

**Review decision:** COMMENT — The review could not produce a reliable result.
**Result:** No code judgment is included because all three reviewers timed out before returning complete findings.
**Scope:** This PR changes unbounded metric bounds to JSON strings and updates schema, code generation, converters, and tests.

## Failure

Round 1 stopped after all three Codex reviewers reached the 420-second limit. Each process returned no complete review response.

This report does not mean the change is clean or unsafe. A later run can retry the review with a larger reviewer timeout.

<details>

- Target: PR #207 (`fix/inf-serialization`, 7 files, +247/-15)
- Rounds: 0/1 completed; stopped in round 1
- Mix: 3 codex-exec
- Reviewer results: 3 timed out with empty final output
- Checks: not run
- Second check: not run

</details>

---

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This is comment-only, and several confirmed contract issues remain.
**Result:** The sentinel serializer works on the tested JSON paths. Schema, validation, regeneration, and converter contracts still need coordination.
**Scope:** This PR represents unbounded metric bounds as exact JSON strings. It updates schema, generated models, codegen hooks, LM-Eval conversion, and tests.

**Second check:** 2 reviewers checked 9 findings; 7 kept, 1 lowered to a suggestion, and 1 removed as pre-existing.

## Earlier review comments

Four items in the [earlier serialization review](https://github.com/evaleval/every_eval_ever/pull/207#issuecomment-5121718172) remain present:

- **[high] LM-Eval conversion** — Standard metrics absent from `KNOWN_METRIC_BOUNDS` can still disappear from output. (`RAVF001`)
- **[high] validation contract** — Raw-file validation still accepts numeric strings that the published schema rejects. Sign and ordering rules also remain unenforced. (`RAVF002`)
- **[high] schema versioning** — String-valued bounds still use schema version `0.2.2`. (`RAVF003`)
- **[medium] generated type** — The documented regeneration path does not preserve the checked-in `float | None` annotation. (`RAVF004`)

The Papers with Code review on this PR concerns stacked changes outside this serialization scope. The [earlier review-anvil timeout report](https://github.com/evaleval/every_eval_ever/pull/207#issuecomment-5122154889) made no code judgment and is superseded by this completed run.

## What I noticed

|---|---|---|---|---|
| RAVF005 | high | regeneration imports | `post_codegen.py:155` | The documented regeneration path can create an invalid parenthesized Pydantic import. |
| RAVF006 | medium | strict conversion | `every_eval_ever/converters/lm_eval/adapter.py:254` | Missing metric bounds warn and discard results even when `strict_validation=True`. |
| RAVF008 | medium | Pydantic schema | `every_eval_ever/eval_types.py:472` | Pydantic’s serialization schema excludes the Infinity strings that the model emits. |

<details>
<summary>Non-blocking low-priority suggestion</summary>

- **[low] assignment validation** `every_eval_ever/eval_types.py:465` — Post-construction NaN assignment can bypass validation and serialize as null. No reachable repository mutation path was found. (`RAVF007`)

</details>

## Things to try

- **[high] LM-Eval conversion** — A defined LM-Eval `0.4.*` metric baseline can cover supported built-ins. Unknown names can follow the strict or permissive policy below. (`RAVF001`)
- **[high] validation contract** — File validation can accept only JSON numbers and exact field-valid sentinels. Runtime checks can also enforce signs and ordering. Python-side coercion can remain an explicit compatibility choice. (`RAVF002`)
- **[high] schema versioning** — One new version can update the schema, converter constant, README, and every emitter together. Existing `0.2.2` records can remain unchanged. (`RAVF003`)
- **[medium] generated type** — The pinned regeneration workflow can select one stable Python contract and verify it with a drift check. (`RAVF004`)
- **[high] regeneration imports** — Import insertion can understand the generated parenthesized block. The exact regenerated output can then be compiled in a regression test. (`RAVF005`)
- **[medium] strict conversion** — Missing bounds can enter the existing transformation-error path. Strict mode can fail before skipping, while permissive mode can warn and skip. (`RAVF006`)
- **[medium] Pydantic schema** — Each bound can expose an accurate serialization contract for finite numbers, null, and its field-valid sentinel. (`RAVF008`)
- **[low] assignment validation** — Assignment checks can cover NaN and valid two-bound updates before becoming part of the generated model contract. (`RAVF007`)

<details>

- Target: PR #207 (`fix/inf-serialization`, 7 files, +247/-15)
- Mix: 3 codex-exec
- Reviewer timeout: 1200 seconds
- Earlier review comments: 4 still present, 1 stacked review outside scope, 1 stale timeout report
- Checks: concerns=4; confirmed=3, ruled-out=0, set-aside=0, lowered=0, narrowed=1
- Second check: targeted; reviewers=2; kept=7, lowered=1, removed=1; approval changed=no

</details>

---

post_codegen.py
```diff
@@ -131,7 +151,9 @@ def patch_file(patch: dict) -> None:
         print(f'  {patch["file"]}: already patched, skipping')
         return
 
-    content = add_import(content, patch['import_add'])
+    imports = patch['import_add']
+    for symbol in [imports] if isinstance(imports, str) else imports:
```
mrshu, line 155: **[high] regeneration imports** — The codegen hook can create invalid Python

The documented workflow produces a parenthesized Pydantic import. `add_import()` rewrites only its opening line, which can produce `from pydantic import (, ...`.

Import insertion can handle the generated multiline block. A regression test can run the exact workflow and compile its output.

every_eval_ever/converters/lm_eval/adapter.py
```diff
@@ -251,8 +252,18 @@ def _build_evaluation_results(
             is_higher_better = higher_is_better.get(metric_name, True)
 
             bounds = KNOWN_METRIC_BOUNDS.get(metric_name)
```
mrshu, line 267: **[medium] strict conversion** — Missing bounds bypass strict mode

This branch warns and discards the metric without using the adapter’s transformation-error handler. A runtime check returned zero results under both `strict_validation=True` and `False`.

Missing bounds can enter the existing error path. Strict mode can fail before skipping, while permissive mode can warn and skip.

every_eval_ever/eval_types.py
```diff
@@ -457,8 +460,28 @@ def validate_score_type_requirements(self):
...
+        # A NaN bound is never meaningful (and NaN != NaN breaks comparisons).
+        # inf/-inf are allowed (unbounded); null is handled above per score_type.
+        for _name in ('min_score', 'max_score'):
+            _v = getattr(self, _name)
+            if _v is not None and _v != _v:
+                raise ValueError(f'{_name} must not be NaN')
         return self
 
+    @field_serializer('min_score', 'max_score', when_used='json')
+    def _serialize_bound(self, value):
```
mrshu, line 472: **[medium] Pydantic schema** — The serialization schema excludes emitted sentinel strings

This serializer has no declared output union. Pydantic therefore describes the bound as number or null, although `model_dump_json()` can emit `"Infinity"` or `"-Infinity"`.

Each bound can expose an accurate serialization contract for finite numbers, null, and its field-valid sentinel.

every_eval_ever/eval_types.py
```diff
@@ -457,8 +460,28 @@ def validate_score_type_requirements(self):
...
+        # An unbounded bound is +/-inf; emit it as the JSON string
+        # "Infinity"/"-Infinity" (valid JSON that reads back to a float) rather
+        # than pydantic's default null or the non-standard bare Infinity token.
+        # when_used='json' covers BOTH model_dump_json() and
+        # model_dump(mode='json'); mode='python' keeps the native float. null
+        # stays null (reserved for "not provided", never unbounded).
+        if value == float('inf'):
```
Erotemic, line 479:
```suggestion
        if value == math.inf:
```

Use the constant value to avoid running a function call each time.

every_eval_ever/eval_types.py
```diff
@@ -457,8 +460,28 @@ def validate_score_type_requirements(self):
...
+            return 'Infinity'
+        if value == float('-inf'):
```
Erotemic, line 481:
```suggestion
        if value == -math.inf:
```

Erotemic: Could also do:

```
import math

POS_INF = math.inf
NEG_INF = -math.inf

if value == POS_INF:
    return 'Infinity'
if value == NEG_INF:
    return '-Infinity'
```

to avoid attribute lookup and the negation each function call

every_eval_ever/schemas/eval.schema.json
```diff
@@ -219,13 +219,18 @@
                                 "description": "Indicates whether there is an Unknown Level - if True, then a score of -1 will be treated as Unknown"
                             },
                             "min_score": {
-                                "type": "number",
-                 
-                                "description": "Minimum possible score for continuous metric"
+                                "oneOf": [
+                                    { "type": "number" },
+                                    { "type": "string", "enum": ["Infinity", "-Infinity"] }
```
Erotemic, line 224: God I hate it, but it's probably the right way to do it here.

every_eval_ever/converters/lm_eval/adapter.py
```diff
@@ -251,8 +252,18 @@ def _build_evaluation_results(
...
-            min_score = bounds[0] if bounds else None
-            max_score = bounds[1] if bounds else None
+            if bounds is None:
```
Erotemic, line 268: The GPT 5.6 review is right here. Anything absent from KNOWN_METRIC_BOUNDS will get skipped. So the question is: are valid bounds always required to be pre-registered or can min and max score live as None, None?

Erotemic (commented):
The top GPT 5.6 review seems important, the others are iffy, they're probably right, but you could live with them. 

My eye caught the `float('inf')` as a function call in a hot loop. Would it matter in reality? Maybe, would need to profile, but it's probably something I would defensively put in if I planed to deserialize millions of rows.

Overall this PR is important and @borgr and @nelaturuharsha you should also take a look and have an another agent rereview. Also use your domain knowledge to determine the answer to the metric bounds question. I don't know what the right call is there. It's a repo policy decision. My recommendation would be a metric not knowing what its bounds are should issue a warning, but shouldn't break or prevent itself from being recorded. Feels like something you could always put in later.

borgr:
Pushed `c343459` to address the silent-drop concern (**`RAVF001`**) on this branch, since it owns the skip behavior.

**Before:** a metric with no `KNOWN_METRIC_BOUNDS` entry was skipped with only a per-metric `warnings.warn` — easy to miss in an attended run, and invisible in an unattended one. The `Converted N log(s)` summary reported only successes.

**Now:** the adapter records each `(task, metric)` skip (`get_skipped_metrics()`), and the `lm_eval` convert command prints the tally to **stderr** after the count:

```
Converted 5 evaluation log(s).
Skipped 3 metric result(s) with no known bounds (bleu, ter); add them to KNOWN_METRIC_BOUNDS to include.
2 of 5 log(s) have no results after skips.
```

Success count stays on stdout; the loss goes to stderr, naming the metrics to add. No behavior change to what's emitted — this only makes the existing drop visible. Tests: adapter records the skip; a CLI test asserts the stderr summary. Full suite green.

every_eval_ever/converters/lm_eval/adapter.py
Erotemic: Man, this is a pure human comment, but it sounds like an LLM, 😄, at least the first sentence.

borgr:
Closing this as **superseded by #212**.

#212 independently lands the same core fix this PR was for — unbounded metric bounds now serialize as the JSON string `"Infinity"` (`field_serializer` on `min_score`/`max_score`, `when_used='json'`), matching this PR's approach — and it goes further, wiring strict-JSON parsing that *rejects* bare `Infinity`/`NaN` on the read side too. So the serialization half of this PR is fully covered there.

The other half here — reporting skipped/unknown-bound metrics instead of silently dropping them (`get_skipped_metrics()` + a stderr tally) — is handled differently in #212: the lm_eval adapter now **preserves** the score with no `score_type`/bounds rather than skipping it. That trades the silent drop for a bare, un-typed score. I've raised on #212 whether it's worth preserving *and flagging* (an `additional_details` note + a "N metrics without bounds" run summary) so the gap stays actionable — that's the better home for the discussion than keeping this PR open.

Nothing here is lost; closing to avoid a redundant/ conflicting serialization path.

PR #208, opened 2026-07-21
Add eee-dataset-conversion agent skill + a CI test that keeps it current

borgr:
## What

Adds an **Agent Skill** that guides a contributor (or a coding agent) through converting an eval dataset/leaderboard into the EEE schema — plus a CI test that keeps the guidance from going stale.

- **`.claude/skills/eee-dataset-conversion/`** — `SKILL.md` (tight entry point; the description is the invoke trigger) + `reference/` (loaded on demand) + runnable `templates/` + `scripts/validate.sh`.
- **`reference/datastore-gate.md`** — what rejects a record *after* the schema passes: the 5-part path and UUID4 filename rules, companion pairing, score-inside-declared-bounds, finite uncertainty, the deployment axes, the publish and partial-conversion contracts. Derived from `REGISTERED_CHECKS`, not from memory.
- **`reference/datastore-submission.md`** — the datastore-PR mechanics that caused most of the historical rework: collection naming, batching large uploads, iterating on the same PR ref, clearing the bot's *warnings* rather than just reading its verdict.
- **Registry-canonical metric ids and scales** — `fields.md`/`registry.md` now say a global metric takes the registry's canonical id (`accuracy`, not `<src>.accuracy`, which fragments the join the datastore exists for), only leaderboard-specific constructs are namespaced, and a metric's canonical bounds/direction come from its registry entry rather than from whatever scale the source printed. Four entity kinds resolve, not just models.
- **`AGENTS.md`** — agent entry point ([agents.md](https://agents.md) standard) indexing the skill.
- **`tests/test_skill_conversion.py`** — see below.
- **`.github/ISSUE_TEMPLATE/skill-gap.md`** — for reporting where the guidance was silent, ambiguous, or wrong. It names this skill and asks which reference file failed, so it lives with the skill rather than in the companion template PR.
- Discoverability pointers from `README.md` and `every_eval_ever/adapters/README.md`.
- **`README.md` instance-example fix**: the published `single_turn` example currently fails validation (`reference`/`output.raw` must be arrays, `sample_id` a string, `answer_attribution` items need `turn_idx`/`extraction_method`/`is_terminal`). Fixed here so the README teaches something that validates.

## Rebased onto the current `main`

The earlier revision of this PR predated #212/#215/#218, and three of its claims were wrong against `main`. Now corrected:

- paths moved `utils/<name>/adapter.py` → `every_eval_ever/adapters/<name>/adapter.py` (plus the run command, reference adapters, lint target, human README);
- `detailed_evaluation_results.file_path` is the **full repo-relative** path via `helpers.datastore_repo_file_path` — **not** the basename the old guidance taught, which is a hard error from both the publisher and the companion check;
- don't hand-roll the write order: publish through `converters.common.publication.publish_evaluation_logs` (staged samples, atomic, rollback, route-collision checks);
- account for unconvertible rows via `SourceConversionResult` + `save_failure_report` into `adapter_reports/` + a non-zero exit, instead of warn-and-continue (which can silently shrink an aggregate's denominator);
- schema marker `0.2.2` → `0.3.0`.

## The CI test — no agent runs in CI

`tests/test_skill_conversion.py` freezes the skill's *prescribed output* twice and re-checks both with the real validator CLI (semantic checks on, at a canonical `data/<collection>/<dev>/<model>/` path):

1. the **templates**, executed end to end — publish an aggregate, publish an aggregate + sidecar pair, and confirm an unconvertible row is reported and exits non-zero;
2. **`tests/data/skill_reference_conversion/`** — one committed conversion produced by those templates: the artifact a contributor following the skill would submit.

A third test asserts the two can't drift apart (the frozen records must still be byte-identical to what the templates produce), and a fourth pins `SKILL.md`'s schema marker to `SCHEMA_VERSION`, so a schema bump forces a re-read of the field claims.

**Warnings fail the test too**, not just errors — the datastore bot reports them and reviewers ask about them, so records built the way the skill prescribes must produce none. Every failure message names which skill file to fix and gives the regeneration command. Verified by tampering: rewriting the frozen record's `file_path` to a basename — the exact claim this PR corrects — fails with the validator's own message.

**Why warnings and not just errors** (kept here rather than in the code, per the docstring convention this PR adds to `AGENTS.md`): a warning means "a human should look at this". These files aren't a submission — they're the exemplar the skill tells contributors to copy, so a warning here is inherited by everyone who follows the guidance, and the datastore bot raises it on every one of their PRs. Real submissions keep the usual latitude to ship with an explained warning; the thing our own docs produce shouldn't need one.

The trade-off: adding a new warning class to the validator will turn this test red in an otherwise unrelated PR. That's intended — the guidance is part of such a change — and nothing here blocks a merge (`main` has no required checks), so the validator change can land first and the skill can follow. Say so if you'd rather it were errors-only.

## Earlier review findings

The three `high` findings from the [GPT-5.6 review](https://github.com/evaleval/every_eval_ever/pull/208#issuecomment-5121667240) lived in `converters/lm_eval/` on a stale base and landed via #212, not here. Of the template findings: `RAVF003` (dataset vs citation provenance), `RAVF005` (accuracy fields flagged as an example to replace), and `RAVF008` (source revision in `evaluation_id`) are fixed. `RAVF006` is resolved by #212's `"Infinity"` serialization. `RAVF007` (bare `python` in `validate.sh`) stays deliberate — it runs in the contributor's active env. `RAVF009` (sidecar rebuilding ids) is fixed properly: the sidecar template now reads `evaluation_result_id` off the aggregate's `evaluation_results` and raises on a sample that names a result the aggregate doesn't have. (Routing through the publisher does *not* cover this — it checks `evaluation_id` and `model_id`, never the result id.)

## Notes

- **Additive and portable** — no machine paths, no secrets; `SCHEMA_VERSION` is imported, never hardcoded.
- The repo-wide `.github/PULL_REQUEST_TEMPLATE.md` is **not** here — that's a repo-wide decision, so it's a companion PR (#225) that stands on its own, and this PR doesn't depend on it.
- The companion `registry-entity-aliases` skill for `eval-card-registry` is intentionally absent: that repo gitignores `.claude/`.
- Verified: full suite green (310 passed / 20 skipped locally; `tests/test_validation_scope.py` needs the `jsonschema` dependency, which CI installs), `ruff check` clean.


Erotemic:
GPT 5.6 Review (3 m 39s):

## Review: Request changes

### High — Unknown metric bounds now produce successful but incomplete conversions

`bounds is None` means only that the metric is absent from the small local lookup table; it does **not** mean the metric is unrepresentable. The new branch drops every unknown or unbounded metric, and then drops the entire task when none remain. The CLI still exits successfully, so custom and newly added lm-eval metrics can disappear from output behind warnings.

This also conflicts with the PR’s own guidance that non-trivial data drops require an operator decision. The repository already has a precedent for preserving metrics using explicitly recorded observed bounds rather than discarding them.

**Fix:** distinguish “unknown bound” from “known unbounded metric.” Unknown metrics should require an explicit override or fail with an actionable message rather than yielding partial output. For genuinely unbounded metrics, either use a documented observed-bound convention or make omission an explicit caller-selected policy with dropped-metric accounting.

---

### High — TER is assigned a maximum the code admits is false

The new mapping sets `ter` to `(0, 100)` while its own comment acknowledges that TER can exceed 100.

`MetricConfig.max_score` is defined as the **maximum possible score**, not a customary display scale. Because scores are not range-checked, a TER value above 100 would validate alongside metadata claiming that value is impossible.

**Fix:** do not encode 100 as TER’s maximum. Handle it through the same explicit unbounded/observed-bound policy used for other such metrics.

---

### Medium — The aggregate template conflates dataset provenance with citation provenance

The template accepts a parameter named `ref_url` and stores it in `source_data`, even commenting that it is the “DATASET / citation URL.”

The accompanying reference correctly says `source_data` must identify the dataset that was evaluated, while per-score citations have no typed field and belong in `additional_details`. The template therefore encourages exactly the semantically incorrect mapping the skill warns against.

**Fix:** rename this input to `dataset_url`, describe it only as dataset provenance, and demonstrate storing a paper or leaderboard citation separately in `additional_details`.

CI is green, but the newly added test explicitly blesses the first two semantic failures rather than detecting them.

every_eval_ever/converters/lm_eval/adapter.py
```diff
@@ -251,8 +254,19 @@ def _build_evaluation_results(
             is_higher_better = higher_is_better.get(metric_name, True)
 
             bounds = KNOWN_METRIC_BOUNDS.get(metric_name)
```
mrshu, line 257: **[high] mixed-metric sidecars** — Aggregate and instance metrics can disagree

This branch can retain `acc` and skip `perplexity`. The instance adapter independently selects the first numeric sample metric, so a mixed task can attach perplexity scores to an accuracy aggregate without a result-level link.

Deterministic aggregate result IDs and structured retained-metric descriptors can keep sidecar scores linked to the matching result.

.claude/skills/eee-dataset-conversion/templates/aggregate_adapter.py
```diff
@@ -0,0 +1,96 @@
...
+def _result(benchmark, score, ref_url):            # one benchmark -> one result
+    return EvaluationResult(
+        evaluation_result_id=f"{SRC}.{benchmark}",     # stable join key; instances point here
+        evaluation_name=f"{SRC}.{benchmark}",          # namespaced id, NOT a title
+        source_data=SourceDataUrl(dataset_name=benchmark, source_type="url",
+                                  url=[ref_url]),        # the DATASET / citation URL
+        metric_config=MetricConfig(
+            metric_name="accuracy", metric_kind="accuracy", metric_unit="proportion",
+            lower_is_better=False, score_type=ScoreType.continuous,  # never omit
+            min_score=0.0, max_score=1.0),               # continuous REQUIRES min/max
```
mrshu: **[high] metric semantics** — The generic template labels every result as 0–1 accuracy

The template accepts only benchmark, score, and URL values. Non-accuracy data can therefore pass validation with the wrong identity, unit, direction, and range.

A small explicit source-score record can require metric identity, unit, direction, score type, and branch-specific bounds or levels.

.claude/skills/eee-dataset-conversion/reference/gotchas.md
```diff
@@ -0,0 +1,53 @@
+# Gotchas that cost real time (with fixes)
+
+*Scope: deeper failure modes and mechanisms. For what each field **means**, see
+`fields.md` / `instance-level.md`; this file is the "why it bit us" layer.*
+
+- **`inf` serialization.** `model_dump_json` serializes `±inf` → `null`,
+  invalidating `continuous` bounds — even though the read path accepts `Infinity`.
+  If you truly need unbounded, serialize yourself with `json.dumps(allow_nan=True)`.
```
mrshu: **[medium] JSON serialization** — This advice can emit non-standard JSON

Python writes `Infinity` tokens when `allow_nan=True`. Strict JSON readers reject those files before schema validation.

The guidance can require strict JSON and finite declared bounds. Schema 0.2.2 can state that unbounded metrics need failure or explicit counted omission.

.claude/skills/eee-dataset-conversion/templates/aggregate_adapter.py
```diff
@@ -0,0 +1,96 @@
...
+        score_details=ScoreDetails(score=score))
+
+
+def make_log(model, developer, results, eval_ts, retrieved_ts):   # DEFAULT: 1 log / model
+    return EvaluationLog(
+        schema_version=SCHEMA_VERSION,
+        evaluation_id=f"{SRC}/{developer}_{model}/{eval_ts}",  # STABLE anchor -> idempotent (NOT now)
```
mrshu: **[medium] evaluation identity** — Mutable snapshots have no source-revision anchor

`evaluation_id` uses `eval_ts`, source, developer, and model. Two changed remote snapshots can keep the same evaluation time and reuse one logical ID.

Mutable remote sources can supply one stable anchor, such as a run ID or source revision. `evaluation_timestamp` can remain separate metadata.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This is comment-only, and confirmed data-integrity risks remain.
**Result:** The skill gives contributors a useful workflow. A few template defaults and cross-file links need tighter contracts.
**Scope:** This PR adds a dataset-conversion skill, runnable templates, validation guidance, contributor decision logs, and stacked LM-Eval fallback changes.

**Second check:** 2 reviewers checked 9 findings; 7 kept at material priority and 2 lowered to suggestions.

## Earlier review comments

All three items in the [earlier review](https://github.com/evaleval/every_eval_ever/pull/208#issuecomment-5121667240) remain present:

- **[high] LM-Eval data loss** — Unknown metrics and whole tasks can still disappear from successful conversions. (`RAVF001`)
- **[high] TER bounds** — TER still declares `100` as a maximum that valid values can exceed. (`RAVF002`)
- **[medium] provenance** — The aggregate template still mixes dataset and score-citation URLs. (`RAVF003`)

## What I noticed

|---|---|---|---|---|
| RAVF004 | high | mixed-metric sidecars | `every_eval_ever/converters/lm_eval/adapter.py:256` | A mixed task can retain `acc` in the aggregate but select skipped `perplexity` for instance scores. |
| RAVF005 | high | metric semantics | `.claude/skills/eee-dataset-conversion/templates/aggregate_adapter.py:38` | The copyable template labels every source result as 0–1 accuracy. |
| RAVF006 | medium | JSON serialization | `.claude/skills/eee-dataset-conversion/reference/gotchas.md:8` | `allow_nan=True` can emit the non-standard JSON token `Infinity`. |
| RAVF008 | medium | evaluation identity | `.claude/skills/eee-dataset-conversion/templates/aggregate_adapter.py:54` | Mutable remote snapshots have no stronger stable identity anchor than `eval_ts`. |

<details>

- **[low] validation environment** `.claude/skills/eee-dataset-conversion/scripts/validate.sh:6` — Bare `python` can use an unrelated checkout or environment. The helper can make interpreter selection explicit without excluding pip installs. (`RAVF007`)
- **[low] sidecar linkage** `.claude/skills/eee-dataset-conversion/templates/instance_sidecar.py:77` — Aggregate and sidecar formulas match today, but the sidecar rebuilds IDs. This can drift when metric-specific result IDs are added. (`RAVF009`)

</details>

## Things to try

- **[high] mixed-metric sidecars** — Aggregate results can receive deterministic IDs from task, filter, and metric. Sidecar conversion can use structured retained-result descriptors and emit only matching linked records. (`RAVF004`)
- **[high] metric semantics** — A small explicit source-score record can require metric identity, unit, direction, score type, and branch-specific bounds or levels. (`RAVF005`)
- **[medium] JSON serialization** — Guidance can require strict JSON and finite declared bounds. Schema 0.2.2 can state that unbounded metrics need failure or explicit counted omission. (`RAVF006`)
- **[medium] evaluation identity** — Mutable remote sources can supply one stable anchor, such as a run ID or source revision. `evaluation_timestamp` can remain separate metadata. (`RAVF008`)
- **[low] validation environment** — The helper can verify that its interpreter imports this checkout and meets the required version. (`RAVF007`)
- **[low] sidecar linkage** — ID and name threading can join the metric-metadata work when templates gain multiple result identities. (`RAVF009`)

<details>

- Target: PR #208 (`add-eee-conversion-skill`, 16 files, +805/-3)
- Mix: 3 codex-exec
- Earlier review comments: 3 still present
- Finding counts: 0 critical, 4 high, 3 medium, 2 low, 0 nit
- Checks: concerns=4; confirmed=3, ruled-out=0, set-aside=0, lowered=0, narrowed=1
- Second check: targeted; reviewers=2; material kept=7, lowered=2, set-aside=0, removed=0; approval changed=no

</details>

---

borgr:
### Update — this PR is now skill-only

**What changed:** the branch was accidentally cut on top of the **closed #206**, so its `lm_eval` converter commit was riding along. I rebased onto `main` and dropped it. This PR is now **purely the skill** (docs + templates).

**That's why the three `high` findings are gone** — all three lived in `converters/lm_eval/`, none in the skill. Their accepted home is **#207**.

| High finding | Now |
|---|---|
| `RAVF001` unknown-metric data loss | → #207 |
| `RAVF002` TER false max | → #207 (fixed there: `ter → (0.0, inf)`) |
| `RAVF004` mixed-metric sidecars | → #207 |

**Template fixes applied here** (`5509f05`):
- **`RAVF003`** — `ref_url` → `dataset_url`; citations go in `additional_details`, not `source_data`.
- **`RAVF005`** — accuracy fields flagged as an example to replace per metric.
- **`RAVF008`** — note to fold a source revision into `evaluation_id` for mutable sources.

**Left as-is, deliberately:**
- **`RAVF006`** — `gotchas.md` already steers to finite bounds; #207 fixes the write-path.
- **`RAVF007`** — bare `python` in `validate.sh` is intentional (runs in the contributor's active env).

borgr:
Pushed `1a44eb0` — aligns the skill with the validator hardening in #212 (which I reviewed [here](https://github.com/evaleval/every_eval_ever/pull/212)). No functional/adapter code, still docs + templates only.

**Heads-up on #207:** I closed it as **superseded by #212** — #212 lands the same `"Infinity"`-string bound serialization (via a `field_serializer`) and adds the strict read path. So the skill's inf guidance now points at that landed behavior, not #207.

What changed here:
- **inf bounds** (`gotchas.md`, `fields.md`): an unbounded `continuous` metric emits `±inf`; the library serializes it as the JSON string `"Infinity"` and it round-trips, so just save through `save_evaluation_log`. Dropped the now-wrong advice to hand-roll `json.dumps(allow_nan=True)` (a bare `Infinity` token is rejected by #212's strict parser) and the "no unbounded type" claim. `null` ≠ unbounded.
- **model deployment axes** (`gotchas.md`, `templates/aggregate_adapter.py`): document `deployment_type` ∈ `{self_deployed, externally_managed, unknown}` and `model_availability` ∈ `{open_weights, closed_weights, unknown}` in `model_info.additional_details`, and set them in the template. Noted that the library auto-fills `"unknown"` while the CLI/bot `validate` errors on a missing key or non-enum value — so set the real value, don't ship an unconsidered `"unknown"`.
- **validate invocation** (`SKILL.md` step 6, `verification.md`, `scripts/validate.sh`, template): #212's CLI rejects a directory arg, so `validate.sh` now expands a dir to explicit `.json`/`.jsonl` files, and the docs teach files-or-glob. Also flagged that the CLI runs the **semantic** checks the library `validate()` skips (path structure, companion pairing, deployment fields), which need the file at its final `data/<collection>/<dev>/<model>/` path.

Proved the skeleton after the edits: built a log from the updated template and it passes `validate`.

PR #209, opened 2026-07-21
Add Papers with Code adapter

borgr:
Converts Papers with Code results into aggregate EEE `EvaluationLog`s — one per canonical model id, `source_type=documentation`. The source is the nightly PostgreSQL backup of the PwC database, published to the HF bucket `huggingface/paperswithcode-backups` under `postgres/*.dump` and read with `pgdumplib`, so no PostgreSQL server is involved (`uv sync --extra paperswithcode`).

Code in `every_eval_ever/adapters/paperswithcode/`, with a maintainer runbook for the one judgment call the adapter cannot make for you (`METRIC_MAINTENANCE.md`).

## Scale reconciliation

Each metric has **one** canonical scale, and it comes from the registry, never from the number PwC printed. Per `(metric, dataset)` leaderboard the adapter infers how that board's numbers map onto it — in log space, mass-aware — and then:

- rescales a whole board percent↔proportion when it is uniformly off (raw value always kept);
- fixes a **lone** impossible value when exactly one ×100 / ÷100 resolves it;
- **flags, never squashes**, a board that genuinely does not fit — that flag is the signal the *registered* scale is wrong, which is how the bad-pixel family in registry#50 was caught.

Every record keeps `raw_value`, the rescale basis and factor, and the resolving registry revision, so each decision above is re-checkable from the data alone.

## Nothing invalid is published

Three ways a source row can fail to become a valid record, all handled the same way — omit the affected unit, record it in `adapter_reports/paperswithcode_failures.json` with its source reference and raw fragment, publish everything else, exit non-zero:

1. **No establishable identity.** A row with no `hf_model_url` whose developer the shared helper does not know is omitted, not filed under a placeholder org. This replaces the `developer/model_name` fallback @yananlong and @Erotemic flagged as unsafe — there is no fallback identity at all now.
2. **A score its bounds cannot contain.** After reconciliation, a score outside `[min_score, max_score]` is dropped per metric cell (a sound metric on the same row still publishes). Two cases reach it: a `scale_anomaly`, which keeps the raw value on purpose, and a boundary overrun inside the scale-classification tolerance.
3. **No usable metric, or a `dataset_id` absent from the dump.**

Neither `--best-effort` nor `--allow-unresolved` relaxes any of the three; they only decide whether an *imperfection* (unresolved metric, unknown direction, scale anomaly) aborts before publishing.

## One model per model

PwC spells the same model differently from row to row, and only some rows carry the HF url that settles it: `moonshotai/Kimi-K2.5` on one, `moonshotai/kimi-k2.5` on the next. Taken at face value each spelling became its own log, splitting one model's results across two records whose paths differ only in case — two directories on Linux, and on a case-insensitive filesystem one directory holding both, named after whichever row was read first.

**That artifact is already in the published set.** Discussion 177 holds no case-variant paths, and yet six of its directories contain two records each:

```
data/paperswithcode/microsoft/Phi-4-multimodal-instruct/
data/paperswithcode/moonshotai/Kimi-K2.5/
data/paperswithcode/moonshotai/Kimi-K2.6/
data/paperswithcode/moonshotai/Kimi-VL-A3B-Instruct/
data/paperswithcode/moonshotai/Kimi-VL-A3B-Thinking-2506/
data/paperswithcode/thinkingmachines/Inkling/
```

Rows are grouped case-insensitively, and the spelling from the HF url wins — recognisable without re-reading the row, because the fallback slug is always lowercased, so an uppercase letter means the url. Sorting settles urls that disagree in case, so the choice never depends on row order.

## The two deployment axes

`model_availability` is read from the source rather than guessed from a name. PwC records `is_open` per leaderboard row, and it is a claim about weights — in this dump GPT-5.5 Pro, Claude Opus 4.8, Gemini 2.0 Flash and GPT-4o are `f`; Gemma 3 27B IT, GLM-5.2, Inkling-Small and dots.llm1.inst are `t`. Because the flag is per row and the axis is per model, every row of a model is read, not the first: **1,249 `open_weights`, 202 `closed_weights`, 4 `unknown`** across 1,455 models.

The 4 are the models whose own metadata argues with itself, and `unknown` is the honest answer for them rather than whichever row was read first:

- `osunlp/UGround-V1-7B` is flagged not-open on a row that links the repo holding its five safetensors shards.
- `Aria-UI/Aria-UI-base`, `facebook/dinov3-vit7b16-pretrain-lvd1689m` and `xlangai/Aguvis-7B-720P` carry contradicting flags across their rows.

Surfaces checked, since `reference/model-deployment.md` asks for them: the flag is the pinned source snapshot's own field, and 738 of the 1,249 open calls independently carry a link to a weight repository. All 742 distinct `hf_model_url` values resolve against the HF model API — **727 (98%) hold downloadable weight files**, 7 are private, 7 hold none, and one points at a `collections/` page rather than a model. Those 15 are why a linked repo is not treated as proof on its own: it only puts a contradicting flag in doubt.

`deployment_type` is `unknown` for all 1,455, deliberately. PwC records what the model is, never who served it for the number on the board, and weight availability does not establish deployment.

## Verification

Reproducible on this branch, offline:

- 41 adapter tests, 452 in the suite (20 skipped); `ruff check` and `ruff format` clean; `uv lock --check` passes; CI green on all four jobs.
- Published fixture records validated through `validate_file(..., run_semantic_checks=True)` at their real datastore paths (`data/paperswithcode/Ruicheng/moge-2-vitl/…`): 3/3 valid, 0 errors, 0 warnings.
- The check that gate protects: the same record with a `1.0005` on a `[0,1]` metric fails semantic validation with `score 1.0005 is outside [min_score=0.0, max_score=1.0]`. That value used to be published.

Over the full pinned dump — `paperswithcode_hf_20260731_031511.dump`, sha256 `00c5f080a4f6…6762cbcb`, the version the published records' own `evaluation_id`s name:

- 13,897 selected rows → **1,455 model logs, 5,881 results, 549 publisher directories**, schema `0.3.0`, one record per model directory, no duplicate `evaluation_id`, nothing under a placeholder org.
- **All 1,455 validate at their real `data/paperswithcode/<developer>/<model>/` paths with semantic checks on: 0 invalid, 0 warnings.**
- 0 scale anomalies, 176 whole-board rescales, 1 per-row fix; direction audited across 940 leaderboards with 0 real errors, and all 91 singleton metrics checked — [details here](https://github.com/evaleval/every_eval_ever/pull/209#issuecomment-5147562117).
- The run exits non-zero by design: **8,350 of the 13,897 rows (60%) have no establishable developer**, across 2,654 distinct names that are method names rather than models — `2D Gaussian Splatting`, `DLinear`, `iTransformer`, `TSMixer`, `SegRNN`. Every one is listed in `adapter_reports/paperswithcode_failures.json` with its source reference. That is rule §1 above working, not a regression.

**What this does to the datastore submission** (discussion 177, currently 4,093 records from an older revision, on schema `0.2.2`). Reconciled against the ref itself:

| | on discussion 177 | regenerated |
|---|---|---|
| records | 4,093 | 1,455 |
| named `(developer, model)` directories | 1,455 | 1,455 |
| records in them | 1,461 | 1,455 |
| records under `unknown/` | 2,632 | 0 |

The named directories correspond **exactly** — 0 added, 0 dropped — and the six extra records are the duplicates listed above, one per colliding directory. Everything else that goes away is the **2,632 under `data/paperswithcode/unknown/`**, the placeholder org this adapter no longer emits: 64% of the submission is filed under an org that does not exist. It is regenerated from this adapter before the coordinated set merges.

## Registry pin

`registry_snapshot.json` vendors 400 canonical metric entries pinned to registry `8b83e9ca`, which is on registry `main`. It previously pinned `0052cd2c`, the branch commit of registry#50 — squash-merged, so the revision every record cited could not be fetched by anyone wanting to check a bound.

Re-pinning adds three canonicals (`average-rank`, `macro-accuracy`, `median-win-rate`), moves four aliases onto them and onto `macro-f1`, and changes **nothing** about the 397 entries that were already there — no bound, direction, score type or review status moves. Over the full dump it moves **54 of 5,881 results**: 52 `Average Rank` from `rank`, 2 `Macro Accuracy` from `accuracy`, onto ids that say which variant was measured. Bounds and direction are identical either way, so **no score changes**, and all 1,455 records keep their `evaluation_id` and directory.

One ambiguity is left and is not mine to fix here: the registry writes `null` for both "unbounded" and "bound unknown", and the adapter emits `±inf` for either (flagged `canonical_min/max: unbounded`, counted in a per-run NOTE). registry#49 separates them at the source.

373 of the 400 entries are `review_status: draft`, including the three new ones. The adapter records that per result as `canonical_review_status` rather than refusing draft entries, because refusing them would refuse nearly every metric — the review status travels with the bound instead of being dropped.

## Related

- Registry metric entries: evaleval/eval-card-registry#50 (merged), evaleval/eval-card-registry#49 (open, `null` → `inf`).
- Auto-downloading the newest dump needs `huggingface_hub>=1.0` for the bucket API, above the range this repo pins. The import is lazy and feature-detects, so only that path is affected and it explains itself; `--dump` works in the declared environment.


Erotemic:
GPT 5.6 Review (4m 14s)

## PR review: **Request changes**

I reviewed the Papers with Code adapter commit (`e3f8de25`) rather than re-reviewing the stacked #207 serialization work. CI is green, but the tests deliberately avoid the real dump and bucket paths, leaving several data-integrity problems uncovered.

### High — The documented bucket-download path cannot find the dumps

`latest_dump_remote_path()` calls `list_bucket_tree(bucket)` and filters the returned entries for `.dump`. The README says the dumps are nested under `postgres/*.dump`, while the Hugging Face API is non-recursive by default. The call will see the top-level `postgres` directory rather than its dump files and report that no dumps exist.   ([[Hugging Face](https://huggingface.co/docs/huggingface_hub/en/package_reference/hf_api?utm_source=chatgpt.com)][1])

There is also an unresolved dependency contradiction: the README says bucket access requires `huggingface_hub>=1.0`, but the project requires `<1.0`; the new extra only adds `pgdumplib`. Consequently, the documented `uv run python -m ...` environment cannot satisfy its own stated requirement.

**Fix:** reconcile the dependency range, call `list_bucket_tree(bucket, prefix="postgres", recursive=True)`, and add a mocked test exercising the default no-`--dump` workflow.

---

### High — “Fail closed” accepts draft or incomplete registry metadata as authoritative

The snapshot generator discards `review_status`, provenance commit, and any explicit distinction between “unknown bound” and “unbounded.” The resolver then treats every name match as resolved, defaults a missing score type to continuous, turns `lower_is_better=None` into `False`, and converts every null bound into infinity.

This is not hypothetical. The vendored snapshot contains entries whose direction and bounds are null, yet they will be emitted as higher-is-better and unbounded. It also directly includes the seven entries from still-open registry PR #50; that PR identifies them as `review_status: draft` and says the `null`→infinity reclassification is future work.

This defeats the stated guarantee that unvetted bounds cannot ship.

**Fix:** preserve and enforce `review_status`; reject entries with unknown direction, score type, or bounds; represent unboundedness explicitly rather than interpreting every null as infinity; and record the exact registry revision used to generate the snapshot.

---

### High — Re-running the same dump produces duplicate records

The README claims the output is stable and idempotent per dump, but every call to `save_evaluation_log()` creates a random UUID filename. The adapter never removes or replaces prior output, and it also regenerates `retrieved_timestamp` from `time.time()`. Running the same command twice therefore appends a second complete set of logically duplicate logs with the same stable `evaluation_id` but different filenames and timestamps.

**Fix:** write each bundle to a deterministic path derived from its `evaluation_id`, or explicitly replace the dump-version output directory. Add a test that executes the conversion twice and asserts the file set and logical records do not grow.

---

### High — Percent rescaling silently guesses and rewrites anomalous values

`parse_metric_value()` strips the `%` marker before returning the value. `reconcile_scale()` then divides every score in `(1, 100]` by 100 whenever the canonical bounds are `[0,1]`, without consulting the original marker or the PwC metric scale. Thus an unmarked anomalous value such as `1.2` is silently changed to `0.012` rather than being reported as out of range.

That is not the “unambiguous percent→proportion case” claimed by the README, and it can make the reported “zero scores out of bounds” result partly self-fulfilling.

**Fix:** preserve an explicit `had_percent_suffix` signal or use authoritative source-scale metadata. Without such evidence, retain the raw score and fail or flag the result instead of modifying it.

---

### Medium — A bare `±` value is incorrectly asserted to be a standard error

The parser describes the notation as `mean +/- sd`, but returns its second component as `std_err`. `score_details()` then always constructs `Uncertainty.standard_error`, even though `x ± y` alone does not establish whether `y` is a standard deviation, standard error, or confidence interval half-width. The schema explicitly distinguishes standard error from standard deviation.

**Fix:** only populate `standard_error` when the source explicitly identifies it as SE. Otherwise preserve the raw uncertainty text in `details`, or add an appropriately typed unknown/reported uncertainty representation.

---

### Medium — Decimal-comma parsing corrupts common three-decimal values

The parser recognizes decimal commas only when there are one or two digits after the comma. A valid value such as `0,991` falls through to the thousands-separator branch and becomes `991`; `97,345` becomes `97345`. These precision levels are common for evaluation metrics. `_to_float()` also accepts non-finite strings such as `NaN` and `Infinity` without filtering them.

**Fix:** detect actual grouped-thousands syntax rather than assuming every three-digit suffix is grouping, reject ambiguous input instead of silently rewriting it, and require `math.isfinite()` for emitted scores. Add regression tests for `0,991`, `97,345`, `NaN`, and `Infinity`.

---

### Medium — `metric_unit` is populated with a range declaration, not a unit

`build_metric_config()` assigns the PwC `scale` field directly to `metric_unit`. The fixtures show values such as `"0-1"` and `"unbounded"`, but the schema defines `metric_unit` as a physical or representational unit such as proportion, percent, points, milliseconds, or tokens. `"unbounded"` is not a unit, and PSNR should ordinarily identify its unit separately from its numeric range.

**Fix:** translate known PwC scale declarations into canonical units where justified, preserve the raw `scale` under `additional_details`, and otherwise leave `metric_unit` unset.

[1]: https://huggingface.co/docs/huggingface_hub/en/package_reference/hf_api?utm_source=chatgpt.com "HfApi Client · Hugging Face"

utils/paperswithcode/adapter.py
```diff
@@ -0,0 +1,968 @@
...
+        snapshot_path: str | Path = SNAPSHOT_PATH,
+    ) -> None:
+        data = json.loads(Path(snapshot_path).read_text(encoding='utf-8'))
+        self._by_id = {m['id']: m for m in data['metrics']}
+        self._index: dict[str, str] = {}
+        for m in data['metrics']:
+            keys = [m['id'], m.get('display_name'), *(m.get('aliases') or [])]
+            for k in keys:
+                if k:
+                    self._index.setdefault(_normalize_metric_key(k), m['id'])
```
mrshu: **[high] metric identity** — Normalized keys can select the wrong canonical metric

The first normalized spelling wins. `CLIP-IQA` can therefore resolve to `clipiqa-plus`, and `chamfer-distance` can resolve to `overall-chamfer`.

Matching can prefer exact case-folded IDs and unique exact aliases. Separator-stripped matching can stay limited to one canonical target.

utils/paperswithcode/adapter.py
```diff
@@ -0,0 +1,968 @@
...
+        source_name='Papers with Code',
+        source_type='documentation',
+        source_organization_name='Papers with Code',
+        source_organization_url=PWC_SITE,
+        # A leaderboard aggregating reported numbers is third_party wrt the model
+        # developer, even when a score was self-reported to it (see fields.md).
+        evaluator_relationship=EvaluatorRelationship.third_party,
+        additional_details={
+            'source_role': 'aggregator',
+            'source_bucket': DEFAULT_BUCKET,
```
mrshu: **[medium] source provenance** — Every run records the default bucket

Custom downloads use `args.bucket`, and local runs use `--dump`. This function receives neither value, so both paths can emit false bucket provenance.

The local-versus-download branch can pass an optional source bucket. Local dumps can omit unknown bucket provenance.

pyproject.toml
```diff
@@ -32,6 +32,7 @@ helm = [
     "crfm-helm>=0.5.14",
     "typer>=0.12,<1.0",
 ]
+paperswithcode = ["pgdumplib>=4.0.0"]
 all = [
```
mrshu, line 46: **[medium] aggregate extra** — `all` omits the Papers with Code dependency

The new adapter imports `pgdumplib` through its own extra, but `all` includes only `inspect` and `helm`. An `every-eval-ever[all]` install cannot run this adapter.

The aggregate extra can include `paperswithcode`. An isolated install check can verify that extra without relying on `--all-extras`.

every_eval_ever/adapters/paperswithcode/adapter.py
```diff
@@ -0,0 +1,968 @@
...
+                'uses_additional_data': coerce_bool(
+                    ev.get('uses_additional_data')
+                ),
+                'external': coerce_bool(ev.get('external')),
+                'external_source_url': ev.get('external_source_url'),
+                'source_url': ev.get('source_url'),
+                'paper_arxiv_url': f'https://arxiv.org/abs/{arxiv_id}'
+                if arxiv_id
+                else None,
+                'paper_title': paper.get('title'),
```
mrshu, line 896: **[medium] paper provenance** — Retained paper source URLs are not emitted

`read_papers_subset()` keeps `paper.source_url`, but this detail builder emits only the title and derived arXiv URL. Papers without arXiv IDs lose their retained direct source.

A distinct `paper_source_url` detail and a non-null fixture case can preserve it.

every_eval_ever/adapters/paperswithcode/adapter.py
```diff
@@ -0,0 +1,968 @@
...
+        slugs = set(args.dataset_slugs or SAMPLE_DATASET_SLUGS)
+        dataset_ids = {
+            d['id']
+            for d in datasets_by_id.values()
+            if d.get('slug') in slugs
+        }
+        missing = slugs - {
+            datasets_by_id[i].get('slug') for i in dataset_ids
+        }
+        if missing:
```
mrshu, line 1824: **[medium] CLI selection** — An all-missing explicit selection succeeds with no output

Missing slugs only print a warning. When every requested slug is absent, the adapter scans an empty selection, writes no logs, and the process still exits successfully.

An explicit non-empty selection can fail when it resolves to zero datasets. Partial matches can keep warning behavior.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This is comment-only, and confirmed high-priority mapping risks remain.
**Result:** The adapter has a solid testable shape. Five focused data-integrity and CLI issues remain.
**Scope:** This PR adds a Papers with Code dump adapter, metric registry refresh, tests, dependencies, and user documentation.

**Checks:** 4 concerns checked; all 4 confirmed.
**Second check:** 2 reviewers checked 5 findings; all 5 kept and 3 fix paths clarified.

<details>
<summary>Earlier review comments (7 items)</summary>

All seven items in the [earlier review](https://github.com/evaleval/every_eval_ever/pull/209#issuecomment-5121694523) remain present:

- **[high] bucket path and dependencies** — Default bucket discovery is non-recursive, and the documented dependency range conflicts with the project.
- **[high] registry metadata** — Draft or incomplete metadata is still accepted as authoritative.
- **[high] repeat runs** — Repeated conversions still create duplicate UUID files and timestamps.
- **[high] percent scaling** — Unmarked out-of-range values can still be rescaled as percentages.
- **[medium] uncertainty** — Bare `±` values are still labeled as standard errors.
- **[medium] numeric parsing** — Three-digit decimal commas and non-finite values remain unsafe.
- **[medium] metric units** — PwC range declarations still populate `metric_unit`.

</details>

## What I noticed

|---|---|---|---|---|
| RAVF001 | high | metric identity | `utils/paperswithcode/adapter.py:355` | Normalized key collisions can map exact metric names to the wrong canonical metric. |
| RAVF002 | medium | source provenance | `utils/paperswithcode/adapter.py:567` | Custom-bucket and local-dump runs are attributed to the default bucket. |
| RAVF003 | medium | aggregate extra | `pyproject.toml:36` | `every-eval-ever[all]` does not install the new `pgdumplib` runtime dependency. |
| RAVF004 | medium | paper provenance | `utils/paperswithcode/adapter.py:489` | The adapter retains `paper.source_url` but omits it from emitted details. |
| RAVF005 | medium | CLI selection | `utils/paperswithcode/adapter.py:915` | An explicit selection with no matching dataset exits successfully and writes no logs. |

## Things to try

- **[high] metric identity** — Matching can use three tiers: exact case-folded IDs, unique exact names or aliases, then unique normalized spellings. (`RAVF001`)
- **[medium] source provenance** — The local-versus-download branch can pass an optional source bucket. Local dumps can omit unknown bucket provenance. (`RAVF002`)
- **[medium] aggregate extra** — The `all` extra can include `paperswithcode`. An isolated metadata or install check can verify that aggregate alone. (`RAVF003`)
- **[medium] paper provenance** — A distinct `paper_source_url` detail and non-null fixture case can preserve this source. (`RAVF004`)
- **[medium] CLI selection** — An explicit non-empty slug selection can fail when it resolves to zero datasets. Partial matches can retain warning behavior. (`RAVF005`)

<details>

- Target: PR #209 (`pwc-adapter`, 15 files, +4158/-16)
- Mix: 3 codex-exec; 2 completed and 1 timed out
- Earlier review comments: 7 still present
- Finding counts: 0 critical, 1 high, 4 medium, 0 low, 0 nit
- Checks: concerns=4; confirmed=4, ruled-out=0, set-aside=0, lowered=0
- Second check: targeted; reviewers=2; kept=5, clarified=3, set-aside=0, removed=0; approval changed=no

</details>

---

borgr:
@Erotemic — fair hit on the percent rescaling. Here's the approach we're planning for it; want to confirm it matches your intent before implementing.

Instead of per-value guessing (÷100 on any value in (1,100]), decide scale per **leaderboard group** (metric × dataset) from the full-dump distribution:

- group values all ≤ 1.0 → proportion, leave as-is;
- group values in (1,100] → the whole leaderboard is on 0–100 → rescale all ÷100 **consistently**;
- any value > 100, or a **bimodal** group (values on both sides of 1.0) → **flag, don't rescale** (`scale_anomaly`), using the explicit `%` suffix as a tie-breaker where present (we'll preserve a `had_percent_suffix` signal from the parser).

Net: we rescale only when a metric's own evidence fixes the scale, never silently ÷100 an anomaly, and "0 out of bounds" stops being self-fulfilling (out-of-range values are surfaced, not rewritten).

Does that address the concern, or would you prefer we keep raw + flag even the confident groups?

Erotemic:
Human comment: Feels like all this auto-scaling is just a guessing game that's going to go wrong at somepoint. You group it one way now, but when there are more results, does that group change? I would prefer an authoritative source. In the case where non is available, but we must present results, flag it, and maybe there can be a concept of "TenativeScalingAuthority" or something. I.e. these results really look like they are all between 0 and 100, but we don't know that 100% and there no authority, so we are going to pretend there is an authority for presentation but flag it with a warning icon or something in a legend (or have the ability to do so). 

Keeping raw is probably the most robust thing to do, but it has a cost if this is a normalization step. Not sure if its possible to do the normalization, but then record what the normalization was as one concise field? That would let you reconstruct the original data without having duplicates on every row (that's if this is applying to instance level results). My review here is not through, I'm going from memory.

borgr:
But we will never have an authority, so what's the alternative, choose the
wrong scale and report them as sometimes 0.8 and sometimes 80 depending on
how it was scraped from the paper?

We also plan to keep raw results in general (like pre adapter results) so
we can reprocess things as the schema shifts and such. But giving others
results that are just wrong, I don't see it as safer just because they are
"raw", sure we supposedly not commit, but someone will eventually go and
use it and they won't clean the messy data, or won't use it if it is built
to be too messy to work with

Erotemic:
If we keep the raw results and can reprocess, then that makes a lot of my worry moot.

borgr:
Was Harsha's idea when we discussed CI, and it fixes so much. If we do it
we will be much less fearful to merge adapters and things and not block
everything to wait for everything else. a new PR is responsible to change
back, just like in code but also in data (changed something that all
adapters need, make a rerun). Ideally that won't happen often of course,
but I liked this solution a lot.

borgr:
## What this adds

A Papers with Code → EEE adapter (`utils/paperswithcode/`) plus a maintainer runbook (`METRIC_MAINTENANCE.md`). Companion registry entries are in eval-card-registry#50.

## Scale reconciliation (the main new piece)

Each metric has **one** canonical scale (from the registry). Per `(metric, dataset)` leaderboard, the adapter infers how that board's numbers map onto it — in log space, mass-aware — and then:
- rescales a whole board percent↔proportion when it's uniformly off (raw value always kept);
- fixes a **lone** impossible value (a per-row typo) when exactly one ×100 / ÷100 resolves it;
- **flags, never squashes**, a board that genuinely doesn't fit — that flag is the signal the *registered* scale is wrong (which is how the bad-pixel family in #50 was caught).

## Verification over the full dump

- **0 scale anomalies, 176 whole-board rescales, 1 per-row fix, 4093/4093 valid.**
- Spot-checked against papers: bad-pixel `[0,100]` (ETH3D "bad 1.0 [%]"), Word Accuracy typo `80.39 → 0.8039` (CVTG-2K reports 0.94 / 0.9116), MOTA 80.3 / MCC 60.5 / STS-B 86.5.
- **Direction** audited across 940 leaderboards → 0 real errors.
- **Singleton QC:** all 91 one-off metrics are real names with in-bounds values; a near-duplicate scan found no accidental splits.

Every emitted record keeps `raw_value`, the rescale basis/factor, and the resolving registry revision, so any decision above can be re-checked from the data alone.

## Follow-up suggestion (not in this PR)

Bare generic names — "Score", "Overall", "Average" — currently resolve to a single shared canonical id (e.g. `score`, which is unbounded). It's ~1% of rows and doesn't corrupt data, but it conflates unrelated benchmarks (an Atari game score vs an exam score both become `score`). A small **stoplist** of generic tokens that keeps them dataset-scoped (or unresolved) rather than collapsed into one id would prevent this on future dumps. Happy to add it if maintainers want.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — Recursive output deletion can remove prior or unrelated data before publication succeeds.
**Result:** One critical, two high, and three medium concerns remain. All 15 earlier feedback entries are fixed.
**Scope:** This PR adds a Papers with Code adapter with metric resolution, score scaling, and publication.
**Checks:** 6 concerns checked; all 6 confirmed.
**Second check:** targeted, 2 reviewers; 6 kept, 6 fix paths clarified, 0 set aside, 0 removed.

## Earlier review comments

- All 15 earlier feedback entries are fixed. Exact metric matching now fails closed on ambiguity.
- Source bucket provenance, aggregate dependencies, paper URLs, and explicit zero-match behavior now follow the earlier guidance.

## What I noticed

- **RAV-RUN2-R1-F001 [critical] publication-safety** `utils/paperswithcode/adapter.py:1693` — The adapter recursively deletes the selected output tree before publication. Empty batches and later write failures can destroy prior or unrelated data. (inline)
- **RAV-RUN2-R1-F002 [high] conversion-accounting** `utils/paperswithcode/adapter.py:915` — Malformed metrics, missing dataset references, and resultless rows are silently omitted. The command can exit successfully with incomplete output. (inline)
- **RAV-RUN2-R1-F003 [high] score-scaling** `utils/paperswithcode/adapter.py:469` — The adapter divides a group by 100 when its median is out of range and 90 percent then fit. This decision ignores source scale evidence. (inline)
- **RAV-RUN2-R1-F004 [medium] score-bounds** `utils/paperswithcode/adapter.py:357` — Scale tolerance accepts small boundary overruns but emits them unchanged. Current semantic validation rejects those scores. (inline)
- **RAV-RUN2-R1-F005 [medium] metric-metadata** `utils/paperswithcode/adapter.py:830` — `metric_unit` stays tied to the PwC source scale after score conversion. It can conflict with the emitted score and canonical bounds. (inline)
- **RAV-RUN2-R1-F006 [medium] runtime-dependencies** `utils/paperswithcode/adapter.py:1260` — The default download path needs `huggingface_hub` 1.x, but the project requires a version below 1.0. The declared environment exits before download. (inline)

<details>

- **[critical] publication-safety** — New or empty targets can use the batch publisher. Symlinks, roots, and populated targets can fail by default. Later replacement can use an explicit ownership manifest and recoverable prior directory. (`RAV-RUN2-R1-P001`; covers `RAV-RUN2-R1-F001`)
- **[high] conversion-accounting** — Metric cells can be the source unit. Each attempted or rejected cell can count once, while malformed objects and missing datasets retain row references. (`RAV-RUN2-R1-P002`; covers `RAV-RUN2-R1-F002`)
- **[high] score-scaling** — Rescaling can require an explicit recognized marker and compatible canonical definition. Other out-of-range cells can retain raw provenance in the failure report and stay out of publication. (`RAV-RUN2-R1-P003`; covers `RAV-RUN2-R1-F003`)
- **[medium] score-bounds** — Tolerance can support scale classification only. A strict pre-publication check can reject unproven overruns; documented rounding can retain raw and adjusted values. (`RAV-RUN2-R1-P004`; covers `RAV-RUN2-R1-F004`)
- **[medium] metric-metadata** — A unit can be set only when the canonical contract or an explicit conversion proves it. Otherwise it can stay unset while `pwc_scale` remains in details. (`RAV-RUN2-R1-P005`; covers `RAV-RUN2-R1-F005`)
- **[medium] runtime-dependencies** — `--dump` can be required in the current environment. Help, examples, and tests can avoid the broken default until the declared client supports bucket access. (`RAV-RUN2-R1-P006`; covers `RAV-RUN2-R1-F006`)

</details>

<details>

- Target: PR #209 at `eb4e5ef0334432499c97eb33346e1e03f229ff80` (16 files, +8479/-16)
- Run ordinal: 2
- Mix: 3 codex-exec
- Earlier review comments: 15 ledger entries; all fixed
- Finding counts: 1 critical, 2 high, 3 medium, 0 low, 0 nit
- Checks: concerns=6; confirmed=6/ruled-out=0/set-aside=0/lowered=0
- Second check: targeted; reviewers=2; kept=6/clarified=6/set-aside=0/removed=0; approval changed no

</details>

---

utils/paperswithcode/adapter.py
```diff
@@ -0,0 +1,1726 @@
...
+            'to tolerate only the unresolved class. See the report above.'
+        )
+    if fatal and args.best_effort:
+        print(
+            'best-effort: emitting despite ' + '; '.join(fatal)
+            + ' (all flagged in the output).',
+            file=sys.stderr,
+        )
+
+    # Replace, don't accumulate: uuid4 filenames mean a re-run would otherwise pile
```
mrshu: **RAV-RUN2-R1-F001 [critical] publication-safety** — The adapter recursively deletes the selected output tree before publication. Empty batches and later write failures can destroy prior or unrelated data.

New or empty targets can use the batch publisher. Symlinks, roots, and populated targets can fail by default until replacement has an explicit ownership manifest.

every_eval_ever/adapters/paperswithcode/adapter.py
```diff
@@ -0,0 +1,1726 @@
...
+    dataset: dict[str, Any],
+    task: dict[str, Any] | None,
+    resolver: MetricResolver,
+    metric_ranges: dict[str, tuple[float, float]],
+    metric_meta: dict[str, dict[str, Any]],
+    paper: dict[str, Any] | None,
+    group_scales: dict[tuple[Any, str], GroupScale] | None = None,
+) -> list[EvaluationResult]:
+    """Fan one evaluation row out to one EvaluationResult per jsonb metric."""
+    try:
```
mrshu, line 925: **RAV-RUN2-R1-F002 [high] conversion-accounting** — Malformed metrics, missing dataset references, and resultless rows are silently omitted. The command can exit successfully with incomplete output.

Metric cells can be the source unit. Each attempted or rejected cell can count once, while malformed objects and missing datasets retain row references.

every_eval_ever/adapters/paperswithcode/adapter.py
```diff
@@ -0,0 +1,1726 @@
...
+    # mistaken for a rescale (dividing anything small by 100 trivially "fits").
+    mixed = _detect_mixture(values, lo, hi, floor)
+    if mixed is not None:
+        return mixed
+
+    # (2) Uniform off-scale: the whole board reported on one wrong scale. The
+    # robust centre itself must be out of range at identity but rescued by a
+    # single factor (a factor that merely fits small values is not enough).
+    center = statistics.median(values)
+    if not _in_canonical_range(center, lo, hi):
```
mrshu, line 471: **RAV-RUN2-R1-F003 [high] score-scaling** — The adapter divides a group by 100 when its median is out of range and 90 percent then fit. This decision ignores source scale evidence.

Rescaling can require an explicit recognized marker and compatible canonical definition. Other out-of-range cells can stay out of publication and retain raw provenance in the failure report.

every_eval_ever/adapters/paperswithcode/adapter.py
```diff
@@ -0,0 +1,1726 @@
...
+# Relative tolerance for "out of range enough to matter" -- keeps rounding noise
+# at a boundary (a 1.001 on a [0,1] board) from being treated as a scale error.
+_SCALE_REL_TOL = 1e-3
+
+
+def _range_abstol(bound: float) -> float:
+    return max(abs(bound) * _SCALE_REL_TOL, 1e-9)
+
+
+def _in_canonical_range(v: float, lo: float, hi: float) -> bool:
```
mrshu, line 356: **RAV-RUN2-R1-F004 [medium] score-bounds** — Scale tolerance accepts small boundary overruns but emits them unchanged. Current semantic validation rejects those scores.

Tolerance can support scale classification only. A strict pre-publication check can reject unproven overruns.

utils/paperswithcode/adapter.py
```diff
@@ -0,0 +1,1726 @@
...
+    obs_range: tuple[float, float],
+    metric_meta: dict[str, Any] | None,
+) -> MetricConfig:
+    meta = metric_meta or {}
+    return MetricConfig(
+        evaluation_description=meta.get('description'),
+        metric_id=resolved.metric_id,
+        metric_name=metric_name,
+        metric_kind=resolved.metric_kind,
+        # A unit only when PwC's scale maps to a real one; a range/"unbounded"
```
mrshu: **RAV-RUN2-R1-F005 [medium] metric-metadata** — `metric_unit` stays tied to the PwC source scale after score conversion. It can conflict with the emitted score and canonical bounds.

A unit can be set only when the canonical contract or an explicit conversion proves it. Otherwise it can stay unset while `pwc_scale` remains in details.

every_eval_ever/adapters/paperswithcode/adapter.py
```diff
@@ -0,0 +1,1726 @@
...
+# requirement. `_require_bucket_api()` imports lazily and fails with an actionable
+# message ONLY when auto-download is actually triggered in an environment that
+# can't do it. See the adapter README ("Data source") and #209 review (Ero #1).
+def _require_bucket_api():
+    """Return an ``HfApi``, or exit with a clear remedy if the bucket API is absent.
+
+    The two bucket methods land together in ``huggingface_hub>=1.0``; feature-
+    detecting ``list_bucket_tree`` is more robust than parsing a version string
+    (and lets the test suite substitute a fake ``HfApi``).
+    """
```
mrshu, line 1415: **RAV-RUN2-R1-F006 [medium] runtime-dependencies** — The default download path needs `huggingface_hub` 1.x, but the project requires a version below 1.0. The declared environment exits before download.

`--dump` can be required in the current environment until the declared client supports bucket access.

every_eval_ever/adapters/paperswithcode/adapter.py
```diff
@@ -0,0 +1,1726 @@
...
+from every_eval_ever.helpers import (
+    SCHEMA_VERSION,
+    get_developer,
+    sanitize_filename,
+    save_evaluation_log,
+)
+
+SRC = 'paperswithcode'
+PWC_SITE = 'https://paperswithcode.com'
+DEFAULT_BUCKET = 'huggingface/paperswithcode-backups'
```
yananlong, line 80: In the description the storage bucket was given as `nielsr/paperswithcode-backups` but `https://huggingface.co/buckets/nielsr/paperswithcode-backups` returns 404.

every_eval_ever/adapters/paperswithcode/adapter.py
```diff
@@ -0,0 +1,1726 @@
...
+    display = str(model_name if model_name not in (None, '') else 'unknown')
+    display = display.strip() or 'unknown'
+    if hf_model_url:
+        m = HF_MODEL_RE.match(str(hf_model_url).strip())
+        if m:
+            dev, mdl = m.group(1), m.group(2)
+            return f'{dev}/{mdl}', dev, mdl, display
+    dev = slugify(get_developer(display) or 'unknown')
+    mdl = slugify(display, 'unknown')
+    return f'{dev}/{mdl}', dev, mdl, display
```
yananlong, line 247: When `hf_model_url` is available, the URL gives us a reasonably strong model identity. But when it is absent or doesn't match, we fall back to developer/model_name and use that as a global identity. That seems unsafe across the full PwC corpus: unrelated papers can both have models named Baseline, Ours, etc., and missing names all become `unknown/unknown`. Those rows will then be treated as the same model and their evaluation results can be merged.

Could the fallback identity be scoped to the source paper/evaluation instead, unless there is stronger evidence that two rows refer to the same model?

utils/paperswithcode/adapter.py
```diff
@@ -0,0 +1,1726 @@
...
+                'arxiv_id': row.get('arxiv_id'),
+                'title': row.get('title'),
+                'source_url': row.get('source_url'),
+            }
+    return out
+
+
+def dump_version_from_path(dump_path: str | Path) -> str:
+    m = re.search(r'(\d{8})(?:_\d+)?', Path(dump_path).name)
+    return m.group(1) if m else Path(dump_path).stem
```
yananlong: The source filenames already carry a more precise dump stamp such as `20260716_031511`, but this intentionally reduces that to `20260716`, so the claimed “idempotent per dump” identity is actually per model per calendar day. Two different same-day backups, retries, or locally supplied snapshots become indistinguishable at the evaluation_id/retrieval-time level even though they may contain different data.

Could `dump_version` retain the complete dump timestamp, or preferably an immutable source-object identifier/hash when available? The date can still be used for partitioning, but the durable record identity should identify the exact source snapshot rather than only its day.

every_eval_ever/adapters/paperswithcode/adapter.py
Erotemic: I agree this behavior is unsafe.

borgr:
Run-2 findings `RAV-RUN2-R1-F001`–`F006` are all addressed — the last two landed in 685405e6 — and the three comments from 2026-08-07 are answered at the end. Run-1 (`RAVF001`–`F005`) and @Erotemic's first-round comment were answered in the threads above and confirmed fixed by run 2, so I won't repeat them here.

**F001 · publication-safety — recursive delete before publication**

Gone; there is no `rmtree` of the output tree anywhere. The order is now: validate and write the whole new batch, then `remove_superseded_records` unlinks only the *files* the previous run wrote (skipping any path the new batch reused) and prunes just the directories that call emptied, never at or above the output root. So the output directory always holds one complete set — the previous run's if this one fails or is empty, this run's if it succeeds. `adapter.py:1430`, pinned by `test_superseded_records_are_removed_and_kept_ones_survive` and `test_emptied_directories_are_pruned_no_higher_than_the_output_root`.

**F002 · conversion-accounting — silent omissions, exit 0**

Every omission is now counted and reported, at the granularity that actually failed: the source row, or the row plus metric cell when only one cell is unusable. `build_logs` returns a `SourceConversionResult(total_records, records, failures, exclusions)`; each omission is a `SourceRecordFailure(source_ref, reason, source_record)` with `source_ref = evaluations.id=<id>` (plus ` metric=<name>` for a cell) and the raw fragment, written to `adapter_reports/paperswithcode_failures.json` outside `data/`. `raise_if_incomplete()` then makes the command exit non-zero. Malformed metric objects, a `dataset_id` absent from the dump, resultless rows, unestablished identity and out-of-range cells all go through that one path.

**F003 · score-scaling — ÷100 on a median plus a 90% fit**

Two halves. I disagree with the first and implemented the second.

The plan asks that rescaling require "an explicit recognized marker". There is no such marker in the dump. PwC's per-metric `scale` hint is absent or wrong far more often than it is right — `Accuracy` carries `scale: null` on boards that report `97,3` — so the numbers themselves are the only evidence available. The decision is therefore made from group evidence rather than one row's median: `analyze_group` works per `(metric, dataset)` board in log10 space with a mass floor of `max(3, 5% of N)` and returns *uniform* / *mixed* / *anomaly*. A whole board is rescaled only when one ×100/÷100 brings essentially all of it in range; two genuinely coexisting scales must show as two log-separated clusters that each clear the mass floor across an empty ~0.75-decade valley, and are rescaled per cluster; a substantial minority out of range with no consistent factor is **not** rescaled — it is flagged `group_scale_mismatch` and kept raw. That last branch is how the mis-bounded bad-pixel family in evaleval/eval-card-registry#50 was found, which is the argument for keeping it. `raw_value`, `canonical_rescale_factor`, `rescale_basis` and the resolving registry revision ride on every record, so each decision is re-checkable from the published data alone. Tests: `test_analyze_group_clean_and_uniform_percent`, `test_analyze_group_lone_stray_is_fixed_per_row`, `test_analyze_group_systematic_mismatch_is_flagged_not_fixed`, `test_analyze_group_true_mixture_rescales_per_cluster`.

The second half — "other out-of-range cells retain raw provenance in the failure report and stay out of publication" — is now exactly what happens; see F004.

**F004 · score-bounds — tolerated overruns emitted unchanged**

Confirmed, fixed in 685405e6. The `1e-3` tolerance now only classifies scales; publication uses exact bounds. `_partition_out_of_range` runs immediately before results are grouped into logs and drops any cell whose emitted score falls outside its own `[min_score, max_score]` — in **either** mode, `--best-effort` included — into the failure report with the metric it came from. A sound metric on the same row still publishes.

The check that gate protects, verified locally: a published fixture record carrying `1.0005` on a `[0,1]` metric fails `validate_file(..., run_semantic_checks=True)` with `score 1.0005 is outside [min_score=0.0, max_score=1.0]`. That value used to ship. Test: `test_scores_outside_canonical_bounds_are_reported_not_published`.

**F005 · metric-metadata — `metric_unit` tied to the source scale**

Confirmed, fixed in 685405e6. `metric_unit` now names the scale the emitted score is actually on: `proportion` for canonical `[0,1]`, `percent` for `[0,100]`, and unset for every other bound pair (`[0,inf)` errors, `[-1,1]` correlations, `[1,5]` MOS) as well as for an unresolved metric, which has no canonical contract to name. PwC's declaration is kept verbatim as `pwc_scale` in `additional_details`. On the fixtures this flips `Accuracy` (PwC `scale: null`, raw `97,3` → emitted `0.973`) and `SSIM` (no PwC metadata at all) to `proportion`, and leaves `PSNR`/`AbsRel` unset. Test: `test_metric_unit_comes_from_the_canonical_bounds_not_the_source_scale`.

**F006 · runtime-dependencies — bucket API needs `huggingface_hub>=1.0`**

Confirmed. The import is lazy and feature-detects `HfApi.list_bucket_tree` rather than parsing a version string, so only the auto-download path is affected and it fails with the installed version and the fix rather than an `AttributeError`. `--dump` — the documented path, and the one every number reported on this PR came from — has no such requirement. Raising the repo-wide `huggingface_hub` pin is a decision for the whole project rather than this adapter, so it is disclosed in the description instead of done here.

---

**@yananlong on the bucket ([adapter.py:80](https://github.com/evaleval/every_eval_ever/blob/pwc-adapter/every_eval_ever/adapters/paperswithcode/adapter.py#L80))** — you were right and the description is fixed. `DEFAULT_BUCKET` in code was always `huggingface/paperswithcode-backups`, which resolves; only the PR text said `nielsr/`. Sorry for the misdirection.

**@yananlong and @Erotemic on the fallback identity** — agreed, and there is no fallback identity any more. `model_identity` prefers the HF url; otherwise the developer must come from `helpers.get_developer`, and a name that helper does not know **raises**. The row is omitted, listed in the failure report with its raw fragment, and the run exits non-zero. So `unknown/unknown` can no longer be produced, and a bare `Baseline` or `Ours` with no establishable owner is never written at all, let alone merged with another paper's.

On scoping the id to the paper instead: I chose omit-and-report over a paper-scoped synthetic id because the datastore keys records by canonical model id, and `paper-1234/ours` is an id nothing can join on — it would also spawn a directory per paper under `data/paperswithcode/`. Canonicalising a bare method name is the registry's job; until an entry exists, omitting the row is more honest than inventing an owner for it.

Where identity *is* established, two rows do merge into one log. That is intended — several papers reporting the same released model — and it stays traceable: each result keeps its own `pwc_evaluation_id`, `paper_arxiv_url`/`paper_source_url`, dataset and metric, so the source rows remain separable inside the record. Test: `test_unestablished_developer_is_reported_not_dropped`.

The datastore submission is regenerated from this adapter before the coordinated set merges: the identity change and the new bounds gate can only remove records, but the data currently on that submission predates both.

every_eval_ever/adapters/paperswithcode/adapter.py
borgr: Good catch — the description was wrong, the code was not: `DEFAULT_BUCKET` is `huggingface/paperswithcode-backups`, which resolves. Description fixed.

every_eval_ever/adapters/paperswithcode/adapter.py
borgr: Agreed, and the fallback is gone: a row with no `hf_model_url` whose developer `helpers.get_developer` cannot establish now raises, so it is omitted, listed in `adapter_reports/paperswithcode_failures.json`, and the run exits non-zero. No `unknown/unknown`, and no bare `Baseline`/`Ours` written at all. Why omit rather than scope the id to the paper, and what keeps legitimately-merged rows separable: https://github.com/evaleval/every_eval_ever/pull/209#issuecomment-5224703264

borgr:
Three changes since the last pass, and the body is rewritten around them. Checks were absent earlier because GitHub had the branch marked `CONFLICTING` — stale, but a conflicting PR gets no merge ref, so `pull_request` CI never fired. Merging `main` (`2069b79`) cleared it and all four jobs now run green on `c1fb2bc`.

**`model_availability` is decided from the source, not left as a placeholder** (`c82a9c1`). PwC's `is_open` is a claim about weights, so it names that axis; it is recorded per leaderboard row while the axis is per model, so every row of a model is read rather than the first. **1,249 `open_weights`, 202 `closed_weights`, 4 `unknown`** across 1,455 models. The 4 are the models whose own metadata argues with itself — `osunlp/UGround-V1-7B` is flagged not-open on a row linking the repo that holds its five safetensors shards, and three others carry contradicting flags across rows. `unknown` there rather than whichever row was read first, per `reference/model-deployment.md`'s "preserve the conflict".

That rule is deliberately narrow, and the evidence is why: all 742 distinct `hf_model_url` values in the dump were resolved against the HF model API, and **727 (98%) hold downloadable weight files** — but 7 are private, 7 hold none, and one points at a `collections/` page. So a linked repo only puts a *contradicting* flag in doubt; it is not promoted to proof on its own. 738 of the 1,249 open calls carry such a link independently. `deployment_type` stays `unknown` for all 1,455 — PwC records what the model is, never who served it for the number on the board.

**The registry pin is now reachable** (`c1fb2bc`): 397 entries at `0052cd2c` — the squash-merged branch commit of registry#50, so nobody could fetch the revision every record cited — become 400 at `8b83e9ca`, on `main`. Measured over the full dump: three canonicals added, four aliases moved, **nothing** changed about the 397 already there, and **54 of 5,881 results** move metric id (52 `Average Rank` off `rank`, 2 `Macro Accuracy` off `accuracy`). Bounds and direction are identical either way, so **0 score changes**, and all 1,455 records keep their `evaluation_id` and directory.

**Full-dump figures are re-measured** on `paperswithcode_hf_20260731_031511.dump` — the version the published records' own `evaluation_id`s name, not the `20260716` the old body quoted. 13,897 rows → 1,455 logs, 5,881 results, 549 publisher directories; **all 1,455 validate at their real `data/paperswithcode/<developer>/<model>/` paths with semantic checks on, 0 invalid, 0 warnings.** The run still exits non-zero by design: 8,350 rows (60%) have no establishable developer, across 2,654 distinct names that are method names rather than models, each listed in the failure report.

**Consequence for the datastore** (discussion 177, 4,093 records, schema `0.2.2`): the named `(developer, model)` directories correspond exactly with the regenerated set — 1,455 in both, 0 added, 0 dropped — so what goes away is **2,632 records under `data/paperswithcode/unknown/`**, 64% of the submission filed under an org that does not exist, plus six duplicates. Those six are the case-collision artifact this branch fixes: 177 holds no case-variant paths, and yet six of its directories hold two records each, because `Kimi-K2.5` and `kimi-k2.5` are one directory on a case-insensitive filesystem. Regeneration is ready and lands before the coordinated set merges.

every_eval_ever/adapters/paperswithcode/adapter.py
```diff
@@ -0,0 +1,1968 @@
...
+#   re10k-2-view   -> url source, an unbounded metric (PSNR) and a lower-is-better
+#                     metric (LPIPS), multi-metric rows.
+SAMPLE_DATASET_SLUGS = ('eth3d-relative', 're10k-2-view')
+
+# Vendored snapshot of the eval-card-registry's canonical metrics (bounds +
+# direction + score_type), so bounds come from the registry without needing it
+# installed at runtime. Regenerate with refresh_registry_snapshot.py.
+SNAPSHOT_PATH = Path(__file__).with_name('registry_snapshot.json')
+
+HF_MODEL_RE = re.compile(
```
mrshu, line 95: **RAV-RUN3-R1-F001 [high] model-identity** — A Hugging Face collection URL becomes a model identity

The expression excludes `datasets/` and `spaces/`, but it accepts `collections/`. `model_identity()` then reads `collections/<owner>` as a model repository. The source snapshot includes one collection URL, so this path is reachable.

Local URL-shape checks can reject reserved namespaces without adding network dependence. An independent fallback identity can remain available and marked unverified.

every_eval_ever/adapters/paperswithcode/adapter.py
```diff
@@ -0,0 +1,1968 @@
...
+        # any scaffold is recorded in additional_details instead.
+        harness_set = harnesses.get(key, set())
+        eval_lib_details = (
+            {'pwc_harness': ', '.join(sorted(harness_set))}
+            if harness_set
+            else None
+        )
+        log = EvaluationLog(
+            schema_version=SCHEMA_VERSION,
+            # STABLE anchor: model id + dump version -> idempotent per dump, never `now`.
```
mrshu, line 1238: **RAV-RUN3-R1-F002 [high] stable-identity** — Same-day snapshots can share an ID while their bundled content differs

`dump_version_from_path()` discards the filename time. `evaluation_id` then uses only that date and the model ID. Two same-day backups can therefore produce different model bundles with one identity.

An immutable dump hash or object revision can identify the source. A stable digest of the bundle's included raw evaluation units can distinguish content without depending on selector syntax.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The adapter is much stronger, and five focused suggestions remain for accounting, identity, scaling, and runtime support.
**Result:** Four high-priority and one medium-priority findings remain actionable. Five uncertain items were set aside.
**Scope:** Add a Papers with Code adapter with stable identity, complete accounting, trustworthy scale conversion, supported runtime behavior, and safe publication.
**Checks:** 5 uncertain concerns were checked; all 5 were set aside because exact-head access failed during that check.
**Second check:** 3 reviewers kept 5 findings and clarified all 5 fix paths.

## Earlier review comments

<details>
<summary>Earlier review comments (33 ledger entries, grouped into 13 topics)</summary>

- Metric matching, source provenance, the `all` extra, paper provenance, empty explicit selection, score bounds, bucket location, and missing-developer handling are fixed.
- `RAV-RUN2-R1-F002` conversion accounting, `RAV-RUN2-R1-F003` score scaling, and `RAV-RUN2-R1-F006` runtime support remain present.
- `RAV-RUN2-R1-F001` publication cleanup and `RAV-RUN2-R1-F005` metric units need exact-head proof. They are set aside below.

</details>

## What I noticed

|---|---|---|---|---|
| RAV-RUN2-R1-F002 | high | conversion accounting | `every_eval_ever/adapters/paperswithcode/adapter.py:925` | A malformed metric cell remains invisible when a sibling cell converts. The command can publish a partial row without accounting for the rejected cell. |
| RAV-RUN2-R1-F003 | high | score scaling | `every_eval_ever/adapters/paperswithcode/adapter.py:471` | Board-wide scale conversion still comes from score distributions. Canonical bounds and neighboring values do not prove the source unit. |
| RAV-RUN3-R1-F001 | high | model identity | `every_eval_ever/adapters/paperswithcode/adapter.py:95` | A Hugging Face `collections/` URL matches the model-path expression and becomes a false `collections/<owner>` model identity. |
| RAV-RUN3-R1-F002 | high | stable identity | `every_eval_ever/adapters/paperswithcode/adapter.py:1238` | Same-day dump snapshots can produce different bundled content with the same `evaluation_id` because the dump time is discarded. |
| RAV-RUN2-R1-F006 | medium | runtime support | `pyproject.toml:16` | The default download path needs the Hugging Face Hub 1.x bucket API, while the declared environment pins the package below 1.0. |

## Things to try

<details>
<summary>Suggested changes (5 items)</summary>

- **[high] conversion accounting** — Metric-cell attempts can be the main conservation unit. Each rejected cell can produce one failure while valid siblings remain publishable. Structural row failures can stay separate. (`RAV-RUN3-R1-P001`; covers `RAV-RUN2-R1-F002`)
- **[high] score scaling** — A conversion factor can require a source marker bound to that exact metric. Raw scores, markers, and factors can remain auditable. Unproved mismatches can enter cell accounting. (`RAV-RUN3-R1-P002`; covers `RAV-RUN2-R1-F003`)
- **[high] model identity** — Local URL-shape checks can reject reserved Hugging Face namespaces without adding network dependence. An independent fallback identity can remain available and marked unverified. (`RAV-RUN3-R1-P003`; covers `RAV-RUN3-R1-F001`)
- **[high] stable identity** — Each model bundle can use an immutable dump hash or object revision plus a stable digest of its included raw evaluation units. Selector syntax and mutable join keys can stay outside the ID. (`RAV-RUN3-R1-P004`; covers `RAV-RUN3-R1-F002`)
- **[medium] runtime support** — This release can require `--dump` and align help, examples, and tests. Automatic download can follow when a compatible dependency path has isolated coverage. (`RAV-RUN3-R1-P005`; covers `RAV-RUN2-R1-F006`)

</details>

- **RAV-RUN2-R1-F001 [critical] publication safety** — We set this aside because the required exact-head deletion check could not complete.
- **RAV-RUN3-R1-F003 [high] unresolved metric identity** — We set this aside because fallback collision reachability could not be checked at the exact head.
- **RAV-RUN3-R1-F004 [high] error gating** — We set this aside because rejected-row effects on strict gating could not be checked at the exact head.
- **RAV-RUN3-R1-F005 [medium] failure report lifecycle** — We set this aside because clean-run report cleanup could not be confirmed.
- **RAV-RUN2-R1-F005 [medium] metric units** — We set this aside because the exact unit contract could not resolve the reviewer disagreement.

<details>

- Target: PR #209 at `c1fb2bcc214dd2c201b03c1d174bf53630e87ed8`
- Run ordinal: 3
- Mix: 3 codex-exec
- Earlier review comments: 33 ledger entries grouped into 13 topics; 8 fixed, 3 still present, and 2 set aside pending proof
- Finding counts: 0 actionable critical, 4 high, 1 medium, 0 low, 0 nit
- Checks: concerns=5; confirmed=0; set-aside=5; exact-head access failed during the verifier pass
- Second check: full; reviewers=3; kept=5; clarified=5 plans; removed=0; approval unchanged
- Set aside: 5 items; reason=exact-head verification unavailable during the required check

</details>

---

yananlong:
### [P1] Account for the new adapter in the current catalog

PR head `c1fb2bcc` is based on `0e4f5027`, before `every_eval_ever/adapters/catalog.py` was added. In a merge simulation with current `main` (`9bce4136`), all 41 Papers with Code adapter tests pass, but `uv run pytest tests/test_adapter_catalog.py::test_every_adapter_package_is_accounted_for` fails because `paperswithcode` is in neither `ADAPTERS` nor `LEGACY_ADAPTERS`.

Please update the branch and add an `AdapterSpec`, or explicitly classify the package as legacy, so the catalog can validate its parser arguments and scheduling status. The existing green checks ran against the older base and therefore do not cover this integration gate.

borgr:
Reviewed all 15 unresolved automated review threads against the current head. One was a real bug and is now **fixed in fa700082a**: a `huggingface.co/collections/<owner>/<slug>` URL matched `HF_MODEL_RE` and produced the bogus identity `collections/<owner>`; adding `collections/` to the negative lookahead routes it to the fail-closed developer resolver (regression test added).

Thirteen are dismissed — two predate the `utils/paperswithcode/` move, and the rest are already addressed (exact-first metric resolution with collision refusal, local vs. default bucket provenance, `[all]` extra now includes paperswithcode, paper source_url emitted, all-missing selection raises, publication-safe supersede, strict score bounds at publication, canonical metric_unit, fail-closed model identity, correct default bucket). The hf-hub finding is moot now that the base pin is `>=1.0`.

One thread I'm leaving open as an author decision — `dump_version_from_path` keys `evaluation_id` on the dump *date*, so two same-day snapshots would share an id (`adapter.py:1239`). PwC's nightly cadence makes this one-dump-per-day in practice; whether to keep the `_HHMMSS` suffix for finer identity vs. the current date-only idempotency contract is a design call. I'll settle it separately.

every_eval_ever/adapters/paperswithcode/adapter.py
borgr: Fixed in fa700082a — added `collections/` to the negative lookahead in `HF_MODEL_RE`, so a `huggingface.co/collections/<owner>/<slug>` URL no longer resolves to the bogus `collections/<owner>` identity but falls through to `get_developer` (omit + report), like any other unresolvable row. Regression test: `test_collection_url_is_not_read_as_a_model_identity`.

every_eval_ever/adapters/paperswithcode/adapter.py
borgr: Fixed in d7079f161. `dump_version_from_path` now retains any sub-day `_HHMMSS` suffix (capturing group rather than a non-capturing one), so two dumps produced on the same date get distinct dump_versions and thus distinct evaluation_ids instead of the second overwriting the first. A date-only filename still yields just the date, so the one-dump-per-day case is unchanged and idempotent. Regression test added.

borgr:
Pushed b635d25ba: the adapter now answers `--emit-source-version` — it prints the current dump's date stamp (resolved by listing the `paperswithcode-backups` bucket, no multi-GB download; a local `--dump` names that file's stamp) and exits without converting.

That is the freshness-probe contract from #261. Once both land, the weekly Papers with Code run skips automatically whenever the upstream dump has not rolled over since the last completed run (source + schema + adapter-code token unchanged). The two PRs compose in either merge order: this flag is inert on its own, and #261's `skip_if_unchanged` defaults on for every catalog entry, so no catalog change is needed here.

every_eval_ever/adapters/paperswithcode/adapter.py
```diff
@@ -0,0 +1,2011 @@
...
+            EvaluationLogOutput(
+                eval_log=bundle.log,
+                base_dir=args.output_dir,
+                developer=bundle.developer,
+                model_name=bundle.model,
+            )
+            for bundle in bundles
+        ]
+    )
+    removed = remove_superseded_records(stale, set(written), args.output_dir)
```
mrshu: The default run converts only two sample datasets, but cleanup recursively deletes every prior JSON not produced by that run. Running the documented command against an existing output directory could wipe most of the Papers with Code export. Scheduled runs use fresh staging, so this specifically affects direct/repeated CLI use.

I leave this to your call @borgr but it might make sense to fix.

mrshu (approved):
Thank you for plowing through this @borgr -- I only had one last thing that might blow up, everything else should be ready.

every_eval_ever/adapters/paperswithcode/adapter.py
borgr: Good catch, fixed in a7d15e94a. `remove_superseded_records` now deletes a stale record only when its `evaluation_id` is one this run actually re-emitted (`superseded_ids = {bundle.log.evaluation_id for bundle in bundles}`), so cleanup is scoped to the current selection instead of the whole tree. A default two-dataset run against an existing full export now removes only its own prior copies of those two datasets and leaves everything else in place. Two tests cover it — `test_supersession_only_removes_reemitted_ids_not_the_rest` and `test_supersession_leaves_unreadable_stale_files`.

PR #212, opened 2026-08-01
Update to validator

nelaturuharsha:
### Update to validator workflow

  Harden validator, converters, and adapter output handling

  ### PR description

  ## Summary

  This PR hardens local validation, converter publication, and adapter failure handling while adding the minimal new model metadata fields.

  It intentionally excludes registry-backed validation, global de-duplication, cron redesign, and the proposed raw-data ledger.

  ## Validator changes

  - Enforce the canonical data path:
    `data/<collection>/<developer>/<model>/<UUIDv4>.json`
  - Reject extra subdirectories and invalid repository paths.
  - Enforce exact aggregate and `_samples.jsonl` companion pairing.
  - Verify sample evaluation/model IDs and declared row counts.
  - Reject duplicate JSON keys and bare `NaN`/`Infinity` values.
  - Require finite scores.
  - Require bounds for continuous metrics and validate supplied bounds.
  - Do not require bounds for binary or other non-continuous metrics.
  - Support explicit files and user-provided globs without implicitly walking the datastore.
  - Use the same validation rules for CLI and library callers.

  ## Model metadata

  Add:

  - `deployment_type`: `self_deployed`, `externally_managed`, or `unknown`
  - `model_availability`: `open_weights`, `closed_weights`, or `unknown`

  Generated models default these fields to `unknown` for compatibility. `post_codegen.py` reapplies the custom validators after model generation, and CI now tests the generated result.

  ## Converter hardening

  - Centralize aggregate and sample publication.
  - Validate all output paths and records before writing.
  - Ensure aggregates cannot reference missing sample files.
  - Validate UUIDs, sample identities, row counts, and checksums.
  - Prevent silent overwrites and output collisions.
  - Roll back files created by a failed publication.
  - Raise when requested sample data is unavailable.
  - Treat zero usable converted logs as an error.
  - Remove duplicate or indirect UUID generation paths.
  - Propagate Inspect and HELM save failures correctly.

  ## Adapter hardening

  - Preserve and write valid records when other source records fail.
  - Record rejected source records and their failure reasons outside `data/`.
  - Exit nonzero after partial conversion failures so automation can detect dropped data.
  - Distinguish intentional exclusions from malformed source records.
  - Stop silently replacing critical model, developer, benchmark, and evaluation identities with `unknown`.
  - Use strict JSON serialization and canonical Windows-safe output paths.
  - Prevent collisions and partially published output batches.
  - Replace Terminal-Bench’s hardcoded snapshot with parsing of the current leaderboard, including offline replay and raw HTML capture.

  ## Out of scope

  This PR does not add:

  - Registry-backed model, dataset, or metric checks
  - Global datastore de-duplication
  - Manifest-based validation
  - Download-avoidant processing
  - A shared raw-data ledger
  - Cron scheduling or publication changes
  - Alternative JSON backends

  ## Testing

  - 284 tests passed
  - 20 optional/environment-dependent tests skipped
  - Changed Python files pass lint and formatting checks


utils/swe_bench_verified/adapter.py
```diff
@@ -200,13 +236,64 @@ def convert_submission(submission_dir: Path, retrieved_timestamp: str, total_ins
...
+            )
+            model_id = require_identity(
+                eval_log.model_info.id,
+                "SWE-bench Verified model id",
+            )
+            if "/" not in model_id:
+                raise ValueError(
+                    f"model id must be developer/model: {model_id!r}"
+                )
+            developer, model_name = model_id.split("/", 1)
```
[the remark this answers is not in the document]
nelaturuharsha: Addressed in 300c7ec2b. EvaluationLogOutput now normalizes slash-separated model routing before batch publication, so nested model namespaces become one safe filesystem component while eval_log.model_info.id remains unchanged. Added SWE-bench Verified and shared path regression coverage, including colons.

utils/multi_swe_bench/adapter.py
```diff
@@ -152,17 +180,66 @@ def convert_submission(
...
+            model_id = require_identity(
+                eval_log.model_info.id,
+                "Multi-SWE-bench model id",
+            )
+            if "/" not in model_id:
+                raise ValueError(
+                    f"model id must be developer/model: {model_id!r}"
+                )
+            developer, model_name = model_id.split("/", 1)
+            outputs.append(
```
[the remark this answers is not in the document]
nelaturuharsha: Addressed in 300c7ec2b through the shared EvaluationLogOutput boundary. Multi-SWE-bench nested model IDs are normalized before publication, so they cannot fail and roll back the batch. Added a regression with a valid sibling record and a nested model identity.

utils/helm/adapter.py
```diff
@@ -321,52 +330,67 @@ def convert(
...
-        if model_info.developer == 'unknown':
-            developer = model_id
-            model = model_id
-        else:
             if '/' in model_id:
                 developer, model = model_id.split('/', 1)
             else:
-                developer = model_info.developer
+                developer = require_identity(
+                    model_info.developer, 'HELM model developer'
```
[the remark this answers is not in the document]
nelaturuharsha: Addressed in 300c7ec2b through the shared EvaluationLogOutput boundary. HELM IDs such as org/family/model retain that identity in JSON while routing to org/family_model; publication no longer fails the full batch. Added adapter-level regression coverage.

evijit:
One note: the dataset we generated for the ICML "who evaluates" paper has codes for open/closed models, which could help with some of this backfill. 

In general, almost all open weights models will have a match on HF hub stats, so that could also be another way to backfill.

nelaturuharsha:
@evijit Sounds good, will make a different PR for backfill. I have some local backfills completed for all the adapters in the datastore - will include both.

borgr:
Did a close read across the validator core, the publication boundary, the new model fields, and the converters. The strict-JSON core (rejecting bare `NaN`/`Infinity` and duplicate keys, per-JSONL-row too), the `"Infinity"`-string bound handling, and the file-level rollback are solid. A few things worth weighing before merge, roughly by impact:

### 1. Breaking schema changes without a `SCHEMA_VERSION` bump (compat)
`detailed_evaluation_results` now requires `format`+`file_path`, the `format` enum drops `"json"` (jsonl-only), and continuous metrics require bounds — all reasonable individually, but `eval.schema.json` stays `"0.2.2"`. Any previously-valid log using `format:"json"` or a partial detail block is now rejected, and a consumer keying on `schema_version` gets no signal the contract changed. Suggest (a) a version bump, and (b) confirming the committed `data/` tree has zero legacy `format:"json"` / partial-detail files (or shipping a migration), since `validate.py main()` runs semantic checks on every file.

### 2. New model fields: enforcement is split, and they're absent from the JSON schema
`deployment_type` / `model_availability` are a good addition, and requiring an *explicit* value (including `unknown`) is the right call. Two wrinkles:
- The requirement is enforced when **validating a file** (`check_model_deployment` treats a missing key as an `error`), but `ModelInfo.default_model_metadata`'s `setdefault(..., 'unknown')` silently fills them when **building** a log. So adapter authors who never think about it ship an unconsidered `unknown` that validates, while legacy files hard-fail. If the intent is to force a conscious decision, the `setdefault` undercuts it — either drop the default so builders must set it too, or make the check a warning. Right now the two paths disagree.
- The fields exist only in pydantic + `validation_core.py`; they're not in `eval.schema.json`. An external `jsonschema` validator therefore neither requires them nor rejects `deployment_type:"banana"`. Adding them to the schema (the codegen source of truth) keeps the two contracts in sync.

### 3. CLI converters are all-or-nothing — reuse the partial-failure machinery you already built
The active `utils/*` adapters do the right thing on a bad record: preserve the good ones, write a report to `adapter_reports/`, and signal partial failure via `SourceConversionResult.raise_if_incomplete()`. The CLI converters don't. AlpacaEval, for instance, raises on the first unmapped developer and emits **nothing for the whole leaderboard**, where it previously converted the row — one new upstream model → zero output. Routing them through the same boundary preserves the good data. Roughly:

```python
# adapter.fetch_leaderboard: collect instead of raising
return SourceConversionResult(
    source_name=benchmark_key, total_records=len(rows), records=logs,
    failures=[SourceRecordFailure(source_ref=f'CSV row {n}', reason=msg)
              for (n, msg) in failed_rows],
)

# cli: write the good logs first, then report + signal non-zero exit
result = adapter.fetch_leaderboard(version)
publish_evaluation_logs(result.records, output_dir, ...)
if result.failures:
    save_failure_report(result, default_failure_report_path(output_dir))
    result.raise_if_incomplete()   # good data already on disk; exits non-zero
```

A `--strict` flag could restore abort-on-any-failure for anyone who wants it, but partial-success feels like the right default. (Note: under the new `require_identity`, a genuinely unmappable developer can't be published as `unknown` anymore, so for those rows the choice really is drop-and-report vs abort-all — and drop-and-report keeps every mappable model.)

### 4. Path flattening is non-injective (collision warning)
The multi-slash fix maps `org/family/model` → dir `org/family_model` (and `:` → `_`). That's many-to-one: a model literally named `org/family_model` flattens to the **same** directory. Files don't overwrite (unique UUID names), so no data loss — but the directory stops uniquely identifying a model, and path→id is no longer recoverable (the true id survives only inside each JSON's `model_info.id`). Low likelihood, but cheap to guard: at publish time, detect when two distinct `model_info.id`s resolve to the same directory and error, rather than silently co-locating them.

### 5. Rollback leaves empty directories (minor)
`publish_evaluation_logs` (and `save_evaluation_logs`) unlink created *files* on failure but never the directories they `mkdir`ed, so a failed publish leaves empty `data/<coll>/<dev>/<model>/` trees. Minimal, safe fix — track the dirs you create and `rmdir` them (empty-only, so it's a no-op on anything pre-existing) during rollback:

```python
created: list[Path] = []
created_dirs: list[Path] = []
try:
    for artifact in prepared:
        p = artifact.path.parent
        while not p.exists():                 # remember dirs we're about to create
            created_dirs.append(p); p = p.parent
        artifact.path.parent.mkdir(parents=True, exist_ok=True)
        with artifact.path.open('xb') as h:
            created.append(artifact.path); h.write(artifact.content)
except Exception:
    for path in reversed(created):
        path.unlink(missing_ok=True)
    for d in sorted(set(created_dirs), key=lambda x: len(x.parts), reverse=True):
        try: d.rmdir()                         # empty-only; safe
        except OSError: pass
    raise
```

### 6. Design question — drop vs preserve for unknown-bound metrics
For genuinely-unknown metrics the lm_eval adapter now preserves the score with no `score_type`/bounds, which avoids the old silent drop (good). But a bare score with no type or range is hard for a consumer to interpret. Have you considered preserving **and flagging** — an `additional_details` note that bounds are unknown, plus a run-summary count ("N metrics emitted without bounds; add to `KNOWN_METRIC_BOUNDS`")? Keeps the data and makes the gap actionable. Also worth confirming a `MetricConfig` with no `score_type` passes both pydantic *and* the JSON-schema `oneOf` (which otherwise routes to the `levels` branch requiring `level_names`).

---

Minor nits: the dead exported write-paths (`cli._write_log`, the `inspect`/`helm` `__main__` savers) still bypass every new guard — worth removing so a future caller can't sidestep them. And `_UUID_RE` (path check) enforces the RFC-4122 variant nibble while `require_uuid4` only checks `version == 4`, so a hand-built UUID could pass one and fail the other.

Nice work overall — the hardening is real and the test coverage is thorough; the above is mostly about compat signaling and making the CLI converters as resilient as the active adapters already are.

nelaturuharsha:
Addressed the actionable review points in 5923e0411: model metadata is now required and enum-constrained in JSON Schema while retaining the agreed temporary builder defaults to unknown; schema regeneration now reapplies and formats the intentional runtime types; AlpacaEval preserves valid rows and writes raw failure/exclusion provenance before signaling partial failure; normalized route collisions are rejected before writing; failed publication removes only empty directories it created; unknown lm-eval bounds are marked with a per-log count; dead CLI/Inspect/HELM write routes are removed; and the RFC-4122 UUID variant check is explicit. I also expanded aggregate/sample tests so a full-looking path in another collection, developer, model folder, or UUID is rejected in both directions and at publication. I did not bump SCHEMA_VERSION: that is a compatibility/migration decision outside the agreed minimal-field scope, and this code repository has no tracked data tree to migrate. Verification: 358 tests pass; changed-file lint, workflow YAML parsing, and the datamodel-codegen plus post-codegen check path pass under locked uv environments.

borgr:
when are small schema versions change? If the schema is exactly the same but allowed fields or optional recommended change (e.g. validators) this doesn't even bump a minor?

evijit:
We should still make it 0.2.3 or something I think just to signal that field optionality has changed?

every_eval_ever/cli.py
```diff
@@ -76,35 +94,124 @@ def _cmd_convert_lm_eval(args: argparse.Namespace) -> int:
...
+        sample_successes = []
+        sample_failures = []
+        for log, eval_uuid in zip(logs, eval_uuids, strict=True):
+            meta: dict[str, Any] = {}
+            try:
+                if not args.include_samples:
+                    publish_logs.append(log)
+                    publish_uuids.append(eval_uuid)
+                    continue
+                meta = adapter.get_eval_metadata(log.evaluation_id)
```
j-chim, line 127: **Fix before merge: samples can be published under the wrong model.**

In directory mode, line 94 sets `parent_eval_output_dir` to the top-level directory for every results file, and the adapter keeps that value (`converters/lm_eval/adapter.py:401-404` only sets the per-file parent when the key is absent). `find_samples_file` then searches `**/samples_<task>_*.jsonl` from the top directory and returns the lexicographically last match.

With two model subdirectories that share a task name, both aggregates resolve to the same samples file. `transform_samples` re-stamps every row with each aggregate's `evaluation_id` and `model_id`, and the checksum is computed after re-stamping, so every publication check passes.

Reproduced: `out/modelA/` and `out/modelB/`, each with its own `results_*.json` and a `samples_math_perturbed_full_*.jsonl` containing different rows. Converting `out` with `--include-samples` published modelB's rows under both models; the two published `_samples.jsonl` files differed only in the stamped IDs. One model's instance data is published as another model's, silently, and the new checksum machinery certifies it.

Fix: resolve samples from each results file's own directory — either stop overriding `parent_eval_output_dir` in directory mode, or make `find_samples_file` search only the results file's directory. The standalone `python -m every_eval_ever.converters.lm_eval` entry point used the per-file parent before this PR, so this is also a regression for that path.

Suggested regression test: two model subdirectories, same task name, different sample rows, `--include-samples`; assert each published `_samples.jsonl` matches its own upstream file.

---
_Generated by [Claude Code](https://claude.ai/code)_

every_eval_ever/cli.py
```diff
@@ -198,13 +379,20 @@ def _cmd_convert_alpaca_eval(args: argparse.Namespace) -> int:
...
 
-    total = 0
+    logs_to_publish = []
+    eval_uuids = []
+    conversion_results = []
     for version in versions:
         cfg_name = LEADERBOARDS[version]['source_name']
         print(f'\n=== {cfg_name} ===')
-        logs = adapter.fetch_leaderboard(version)
+        conversion_result = adapter.fetch_leaderboard_result(version)
```
j-chim: A fetch failure for one version discards the other version's converted rows. This loop has no per-version error boundary, and `publish_evaluation_logs` runs only after every version converts. If the second version's download raises, the first version's rows are lost and no report is written. The standalone CLI previously caught per-version errors and continued with the remaining version.

Suggestion: catch fetch errors per version, publish what converted, record the failed version in the failure report, and exit nonzero.

---
_Generated by [Claude Code](https://claude.ai/code)_

utils/openeval/adapter.py
```diff
@@ -1247,16 +1606,23 @@ def run(args: argparse.Namespace) -> int:
...
-    bundles = make_logs(
+    result = make_logs_result(
         payload,
         revision=args.revision,
         limit_responses=args.limit_responses,
         allow_unknown_benchmark=args.allow_unknown_benchmark,
         include_instances=args.include_instances,
     )
-    paths = export_logs(bundles, args.output_dir)
+    paths = export_logs(result.records, args.output_dir)
```
j-chim, line 1486: An unrecognized model name takes down the whole batch with no report. `normalize_model_info` (line 657) uses `get_developer(name)` without `require_identity`, and `model_name()` (line 583) falls back to `'unknown'` for a missing name. Both produce bundles with `developer='unknown'` and zero recorded failures. `export_logs` on this line then raises `model developer must be known` for the entire batch — before `save_failure_report` runs below. Nothing publishes, and the failure report is lost. For community-submitted payloads, unrecognized model names are the normal case, not the edge case.

Suggestion: require identity during conversion, inside the per-record boundary, so the bad response becomes a recorded failure and the valid records still publish.

---
_Generated by [Claude Code](https://claude.ai/code)_

utils/openeval/adapter.py
```diff
@@ -1200,38 +1459,138 @@ def save_instance_logs(
...
-            bundle.log.evaluation_id,
-            bundle.log.model_info.id,
-            bundle.binary_result_ids,
-        )
-        if detailed is not None:
-            bundle.log.detailed_evaluation_results = detailed
-            path.write_text(
-                bundle.log.model_dump_json(indent=2, exclude_none=True),
-                encoding='utf-8',
+    prepared = []
```
j-chim: This re-implements `publish_evaluation_logs` and has drifted from the shared boundary in three ways:

1. No route-collision guard: two model identities that normalize to the same directory are silently co-located (the shared publisher rejects this within a batch).
2. `mkdir(parents=True, exist_ok=True)` with no directory rollback: a failed publish leaves empty `data/...` trees behind, which the shared publisher now cleans up.
3. The samples checksum is computed over `sample_text.encode('utf-8')` but the file is written in text mode, so newline translation on Windows makes the on-disk bytes differ from the recorded checksum.

Suggestion: add sample-sidecar support to `converters/common/publication.py` and delete this copy. `save_instance_logs` and `aggregate_scores` in this file are already dead code duplicating live logic and are worth removing at the same time.

---
_Generated by [Claude Code](https://claude.ai/code)_

utils/mt_bench/adapter.py
```diff
@@ -452,29 +481,68 @@ def make_log(
...
+) -> SourceConversionResult[tuple[EvaluationLog, str, str]]:
+    rows = list(rows)
     timestamp = retrieved_timestamp or str(time.time())
-    by_model = aggregate(rows)
+    by_model, source_rows, failures = _aggregate(rows)
     bundles = []
     for model in sorted(by_model):
-        result = make_log(by_model[model], timestamp)
+        try:
+            result = make_log(by_model[model], timestamp)
```
j-chim, line 498: An unrecognized model name takes down the whole batch. `make_log` resolves the developer through `normalize_developer_and_slug` (line 268), which returns whatever `get_developer` gives — including `'unknown'`. That is not a `TypeError` or `ValueError`, so this boundary does not catch it. The bundle then makes `save_evaluation_logs` reject the entire batch (`model developer must be known`): valid models go unpublished and no failure report is written.

`utils/mmlu_pro/adapter.py:301` wraps the same call in `require_identity` inside its per-record boundary. The same pattern works here.

---
_Generated by [Claude Code](https://claude.ai/code)_

utils/hle/adapter.py
```diff
@@ -465,18 +478,54 @@ def make_log(
...
+        if row.raw.get('score') is None:
+            failures.append(
+                SourceRecordFailure(
+                    source_ref=f'leaderboard row {index}',
+                    reason='missing score',
+                    source_record=raw_row,
+                )
+            )
+            continue
+        if normalize_developer(row.company) == 'unknown':
```
j-chim, line 505: This developer check is good, but the model-name slug has no equivalent. `model_slug` (line 289) falls back to `'unknown'` when `sanitize_filename` empties the name. Such a row passes these checks with zero failures, and `EvaluationLogOutput` later raises `model name must be known` during export — the whole batch aborts before the failure report is written.

Suggestion: reject rows whose sanitized model name is empty here, next to the developer check, and remove the `or 'unknown'` fallback at line 289.

---
_Generated by [Claude Code](https://claude.ai/code)_

utils/llm_stats/adapter.py
```diff
@@ -531,16 +609,46 @@ def scores_from_benchmark_detail(
...
+            )
             continue
         score_value = first_present(
             entry, ('score', 'benchmark_score', 'normalized_score')
         )
         if score_value in (None, ''):
+            failures.append(
+                SourceRecordFailure(
+                    source_ref=source_ref,
+                    reason='score entry is missing a score value',
```
j-chim, line 650: Missing scores are caught here, but non-finite ones are not. `parse_float` accepts `'nan'` and `'inf'`, and this module has no `math.isfinite` check. A single NaN score converts with zero failures and then fails the entire batch at serialization time (`allow_nan=False`): valid records go unpublished and no failure report is written.

openeval (`adapter.py:554`), vals_ai (`:356`) and terminal_bench_2 (`:293`) check finiteness per record. The same gap exists in artificial_analysis (`:427`), bfcl (`:442`), helm (`:301`), hfopenllm_v2 (`:146`), cocoabench (`:278`) and hle (`:121`).

---
_Generated by [Claude Code](https://claude.ai/code)_

utils/terminal_bench_2/adapter.py
```diff
@@ -39,278 +39,497 @@
...
+                SourceRecordFailure(
+                    source_ref=f'leaderboard row {index}',
+                    reason=str(e),
+                    source_record=entry,
+                )
+            )
+            continue
+        bundles.append((eval_log, org_slug, model_slug))
+    if not bundles and not failures:
+        raise ValueError('Terminal-Bench 2.0: converted 0 source records')
```
j-chim: When every parsed row is malformed, this raise fires before `main` writes the merged failure report, so all the recorded rejection reasons are lost — in exactly the case where the report matters most. The exit is still nonzero.

Suggestion: skip this raise when `failures` is non-empty (`raise_if_incomplete` still signals the error), or move the zero-record check into `main` after the report is written.

---
_Generated by [Claude Code](https://claude.ai/code)_

utils/swe_bench_verified/adapter.py
```diff
@@ -234,21 +322,26 @@ def main():
...
+                result,
+                default_failure_report_path(OUTPUT_DIR),
+            )
+            print(f"Failure report: {report_path}")
+
+    print(
+        f"\nGenerated {len(paths)} files, {len(result.failures)} errors "
+        f"→ {OUTPUT_DIR}/"
+    )
+    result.raise_if_incomplete()
```
j-chim, line 353: `raise_if_incomplete` does not fire when records and failures are both empty, so a run against an empty or wrong directory prints `Found 0 submission directories` and exits 0. The PR description says zero usable converted logs is an error; terminal_bench_2, openeval, llm_stats, mt_bench and mmlu_pro have an explicit guard for this. The same gap exists in multi_swe_bench, swe_polybench and rewardbench.

---
_Generated by [Claude Code](https://claude.ai/code)_

every_eval_ever/eval_types.py
```diff
@@ -312,6 +316,17 @@ class ModelInfo(BaseModel):
         description='Additional parameters (key-value pairs, all values must be strings)',
     )
 
+    # --- validator (added by post_codegen.py) ---
+
+    @model_validator(mode='after')
+    def default_model_metadata(self):
+        """Emit compatibility placeholders for the new model metadata axes."""
+        details = dict(self.additional_details or {})
+        details.setdefault('deployment_type', 'unknown')
```
j-chim, line 326: These defaults keep `additional_details` a plain `dict[str, str]`, so any value passes: `deployment_type='banana'` validates here and again at publication (`helpers/io.py` re-validates with pydantic only). The value is rejected only later, by `check_model_deployment` at validation time. An adapter with a typo therefore publishes locally and fails only when the data PR is validated.

Suggestion: validate the two enum values in this validator so mistakes fail at build time, matching the JSON schema.

---
_Generated by [Claude Code](https://claude.ai/code)_

every_eval_ever/validation_core.py
```diff
@@ -0,0 +1,942 @@
...
+from every_eval_ever.schema import get_schema_version as get_schema_version
+
+DEFAULT_MAX_ERRORS = 50
+
+_EXPECTED_PATH_PARTS = 5  # data / benchmark / developer / model / filename
+_UUID_RE = (
+    r'^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}'
+)
+_AGGREGATE_FILE_RE = re.compile(rf'{_UUID_RE}\.json$', re.IGNORECASE)
+_INSTANCE_FILE_RE = re.compile(rf'{_UUID_RE}_samples\.jsonl$', re.IGNORECASE)
```
j-chim: `re.IGNORECASE` here is wider than the pairing logic, which is case-sensitive (`endswith('_samples.jsonl')` in `_aggregate_repo_path_for_samples`, and the sibling probe builds a lowercase path). A file named `<uuid>_Samples.jsonl` passes path validation but skips every companion rule: it needs no sibling aggregate, and no aggregate has to declare it. Verified: `check_path_structure` accepts the mixed-case path and `check_instance_companion` returns no errors for it.

Publishers only emit lowercase names, so dropping `re.IGNORECASE` closes the hole with no loss. (A mixed-case `.JSON` suffix is already handled safely — it fails later with an `unsupported_extension` error — but lowercasing the rule set makes the two paths consistent.)

---
_Generated by [Claude Code](https://claude.ai/code)_

every_eval_ever/validation_core.py
```diff
@@ -0,0 +1,942 @@
...
+    if (
+        isinstance(value, (int, float))
+        and not isinstance(value, bool)
+        and not math.isnan(value)
+    ):
+        return float(value)
+    return None
+
+
+def check_model_deployment(data: dict[str, Any]) -> list[str]:
```
j-chim, line 577: This check reads only the top-level `model_info`. The JSON schema also requires `deployment_type` and `model_availability` for every judge's `model_info` (via `$defs/judge_config`, which references `$defs/model_info`), and pydantic fills them silently when building. A file whose judge `model_info` lacks these fields therefore passes this validator but fails the published JSON schema — external validators and this tool disagree.

Suggestion: apply the same check to each judge's `model_info` as well.

---
_Generated by [Claude Code](https://claude.ai/code)_

utils/hal/adapter.py
```diff
@@ -508,23 +547,62 @@ def parse_table(html: str, benchmark: BenchmarkDef) -> list[LeaderboardRow]:
...
         cost_usd = _parse_cost(cost_raw)
+        if (
+            cost_raw
+            and cost_raw.lower() not in {"n/a", "na", "-", "—"}
+            and cost_usd is None
+        ):
+            failures.append(
+                SourceRecordFailure(
+                    source_ref=f"{row_ref} cost",
+                    reason=f"could not parse cost from {cost_raw!r}",
```
j-chim: This branch never fires for comma-formatted values, because a value was parsed — just the wrong one. `_parse_cost`'s pattern `\$([\d.]+)` stops at the first comma, so `$1,234.56` parses as `1.0` and is published as `total_cost_usd` with no recorded failure.

`parse_value` in `utils/bfcl/adapter.py` strips commas before parsing; the same is needed here. Related: in `_parse_percent`, a malformed match like `1.2.3` raises an uncaught `ValueError` that voids the whole benchmark through the broad `except ValueError` in `collect_benchmark`.

---
_Generated by [Claude Code](https://claude.ai/code)_

utils/helm/adapter.py
```diff
@@ -321,52 +330,67 @@ def convert(
...
-            ),
-            eval_library=EvalLibrary(
-                name=eval_library_name,
-                version=eval_library_version,
-            ),
-            model_info=model_info,
-            evaluation_results=list(results_by_metric.values()),
-        )
+        try:
+            model_info = model_infos[model_name]
```
j-chim, line 448: Two gaps in this boundary's coverage:

1. It starts too late. The tab/row/cell parsing above has no error boundary, so one malformed header cell (for example a column name without ` - `) raises at the `split(' - ', 1)` unpack and ends the run with nothing published and no report. Short rows also silently drop trailing metrics through `zip` truncation.
2. Missing cell values still become score `-1` (around line 301) with no recorded failure. The score-bounds validator rejects the file downstream as a blocking error, so the value cannot enter the datastore silently — but one missing upstream cell then blocks the whole data PR instead of becoming one recorded failure. This predates the PR; noting it because this adapter was in scope for exactly this class of fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

every_eval_ever/helpers/io.py
```diff
@@ -72,13 +486,87 @@ def save_evaluation_log(
...
+        base_dir = base_dir.with_name(collection)
+        # Revalidate generated/dataclass-constructed values at the publication
+        # boundary so all schema validators run before the first write.
+        validated = EvaluationLog.model_validate(output.eval_log.model_dump())
+        dir_path = generate_output_path(
+            base_dir,
+            output.developer,
+            output.model_name,
+        )
+        route_owner = (collection_identity, validated.model_info.id)
```
j-chim, line 546: `route_owners` is rebuilt on every call, so this guard rejects collisions only within one batch. Across runs it is silent: publishing `org/family/model` in one run and `org/family_model` in a later run co-locates two different models in one directory, and validation does not cross-check directory names against `model_info.id`. No files are overwritten (UUID names, exclusive create) and each file keeps its true id, so nothing is lost — but the directory stops identifying one model, and the flattening cannot be reversed from the path alone.

A full fix needs to read existing files in the target directory at publish time, which this PR's scope deliberately avoids. Deferring is reasonable; suggest documenting the limitation and opening a follow-up issue so the guard is not over-trusted.

---
_Generated by [Claude Code](https://claude.ai/code)_

every_eval_ever/helpers/io.py
```diff
@@ -24,6 +173,271 @@ def sanitize_filename(name: str) -> str:
...
+    """Return a non-placeholder identity value or raise."""
+    if not isinstance(value, str) or not value.strip():
+        raise ValueError(f'{field_name} is required')
+    value = value.strip()
+    if value.lower() == 'unknown':
+        raise ValueError(f'{field_name} must be known')
+    return value
+
+
+def _required_path_component(value: str | None, field_name: str) -> str:
```
j-chim, line 209: Windows-safety here is partial. The checks cover `<>:"|?*`, empty parts, dot components, `data` and `unknown`, but these all pass: Windows reserved device names (`CON`, `NUL`, `COM1`, …), trailing dots, embedded control characters (tab, newline), and names that differ only by case — which collide on the case-insensitive filesystems of Windows and macOS checkouts.

If full Windows-safety is the goal, these need adding. Otherwise, narrowing the claim in the docs avoids over-promising.

---
_Generated by [Claude Code](https://claude.ai/code)_

j-chim (commented):
Apologies for the heavy-handed AI review, am AFK for a few more hours but hope this can help unblock other work today.

From a quick manual pass there seems to be one real bug with remaining nits/suggestions. Would be great to fix the issue if confirmed.



______




## Review summary

I read the full diff, ran the test suite, and checked each finding below by reading and running the code. All tests pass (314 passed, 20 skipped without the optional inspect/HELM extras; CI runs those).

Overall, this PR moves the project in the right direction. The validation core is cleanly layered, with no circular imports between the validator, helpers, and converters. The publication boundary is strong: exclusive file creation, rollback of created files and empty directories, and staged-sample checks with no gap between check and write. The per-record failure pattern in the adapters is the right design, and test coverage is broad.

I found one data-integrity bug that should be fixed before merge, a group of robustness gaps, some consistency issues, and a few limitations worth documenting. The line comments carry the details; this summary groups them.

### Fix before merge

- **Samples can be published under the wrong model.** In directory mode with `--include-samples`, the lm_eval converter can attach one model's samples file to another model's aggregate. ID re-stamping and the checksum (computed after re-stamping) make the wrong data pass every check. Reproduced end to end. See the comment on `every_eval_ever/cli.py`.

### Robustness gaps: failures that erase valid records or their failure report

These all fail loudly (nonzero exit; nothing invalid is published), so they do not corrupt data. But they break the PR's core promise — one bad record must not take down the rest — and most of them lose the failure report:

- openeval and mt_bench: an unrecognized model or developer name passes conversion with zero failures, then makes publication reject the whole batch. Nothing publishes and no report is written. hle has the same pattern through its model-name slug fallback.
- Most utils adapters do not check that scores are finite. A single NaN from upstream aborts the whole batch at serialization time with no report. openeval, vals_ai, terminal_bench_2 and sciarena already check; the others need the same few lines.
- AlpacaEval: a fetch failure for one leaderboard version discards the other version's converted rows.
- terminal_bench_2: when every parsed row is bad, the error fires before the failure report is written, so all recorded reasons are lost.
- swe_bench_verified, multi_swe_bench, swe_polybench and rewardbench exit 0 when zero records convert. The PR description says zero converted logs is an error; these four lack the guard the other adapters have.

### Consistency issues

- `deployment_type` and `model_availability` are enforced only at validation time, not at build or publication time. A typo publishes locally and fails only when the data PR is validated.
- Judges' `model_info` is required by the JSON schema but never checked by the validator, so the two contracts disagree.
- A file named `<uuid>_Samples.jsonl` (capital S) passes path validation but skips every aggregate-pairing rule.
- HAL parses `$1,234.56` as `1.0` and publishes it with no recorded failure.
- openeval re-implements the shared publisher with three drifts: no collision guard, no directory rollback, and a checksum computed over bytes while writing in text mode.

### Limitations to document (no code change needed now)

- The route-collision guard works within one publication call only. Two separate runs can place two different models in the same directory with no error. Nothing is overwritten and each file keeps its true `model_info.id`, so no data is lost — but the directory no longer identifies one model. Detecting this needs reading existing datastore files at publish time, which this PR's scope deliberately avoids. A sentence in the docs plus a follow-up issue would keep the guard from being over-trusted.
- Windows path safety is partial: reserved device names, trailing dots, control characters, and case-only name collisions all pass.
- HELM still turns a missing cell value into score `-1`. The score-bounds validator blocks the file downstream, so it cannot enter the datastore silently, but one missing upstream value then blocks a whole data PR. This predates the PR.

### Schema version

The schema contract changed: `model_info.additional_details` is now required with two required keys, `detailed_evaluation_results` requires `format` and `file_path`, and the `format` enum drops `json`. Files that were valid before are now rejected while still declaring `schema_version: "0.2.2"`. A version bump is already under discussion in this thread; the changes support at least 0.2.3, and the breaking direction would justify 0.3.0.

### Minor notes

- `strict_json_loads` rejects the bare `Infinity` token but accepts the overflow literal `1e999`, which parses to infinity. Only `score` is checked for finiteness; `standard_deviation` and confidence-interval values are not. Low priority: the publication path cannot produce such files, only a hand-crafted data PR can.
- An empty companion `_samples.jsonl` passes validation when the aggregate does not declare `total_rows`.
- Pydantic accepts bound strings like `"inf"` and `"0.5"` that the JSON schema forbids; a present-but-null bound passes the schema but fails pydantic.
- Dead code worth deleting: `InspectEvalLogConverter`, `HELMEvalLogConverter`, openeval's `save_instance_logs` and `aggregate_scores`, cocoabench's `ensure_row_map_complete`.
- rewardbench hardcodes `SCHEMA_VERSION = '0.2.1'`; utils/helm and exgentic hardcode `'0.2.2'`. All three should import the shared constant.

### Tests

`tests/test_hfopenllm_v2_adapter.py` is a good template: it checks the published file, the report location outside the data tree, and the nonzero exit. Missing tests map to the findings above: a NaN score through any utils adapter, an unrecognized model name through openeval or mt_bench, a directory conversion with two model subdirectories and `--include-samples`, a mixed-case samples filename, and zero-record runs for the four adapters without the guard.

---
_Generated by [Claude Code](https://claude.ai/code)_

nelaturuharsha:
Addressed the latest review in ebe471ab4 (after merging main/#213):

- bumped aggregate and instance schemas and active converter output to 0.2.3
- bound lm-eval samples to each results file directory; removed recursive cross-model lookup
- preserved successful AlpacaEval versions when another fetch fails, with provenance + nonzero exit
- moved OpenEval/MT-Bench/HLE identity failures inside per-record boundaries
- rejected non-finite active-adapter scores before publication (HF Open LLM v2 remains legacy/out of scope)
- preserved reports for all-malformed Terminal-Bench and made empty SWE/RewardBench inputs fail
- enforced ModelInfo enums at construction and through post_codegen
- made filename pairing case-exact and applied deployment metadata checks to judge models
- fixed HAL grouped costs/malformed percentages and hardened HELM tab/row/cell boundaries
- replaced OpenEval's duplicate writer with the shared atomic byte publisher
- rejected 1e999/non-finite uncertainty fields, empty sample companions, and non-canonical bound strings
- added Windows reserved-name/trailing-dot/control-character checks

The cross-run route-collision and case-only collision limitations require repository-wide state; they are now explicitly documented rather than presented as guarantees of a single-file/local run. Registry, cron/raw ledger, dedup, and legacy HF Open LLM v2 remain separate scope.

-- written by codex --

PR #218, opened 2026-08-02
Refactor adapters, maintainer tools, and validator package

nelaturuharsha:
## Summary

- Move all adapter modules, shared helpers, migration scripts, READMEs, and the HELM helper under `every_eval_ever.adapters`.
- Move the HF Community Evals tool under `every_eval_ever.tools` and post-codegen under `every_eval_ever.post_codegen`.
- Add validated repository-root handling for post-codegen and package-resource schema loading.
- Organize validation implementation under `every_eval_ever.validator` while preserving root compatibility imports and existing CLI commands.
- Bump the canonical emitted schema version to `0.3.0` across packaged schemas, adapters, converters, documentation, and fixtures.

## Verification

- `uv run ruff check every_eval_ever tests`
- `uv run pytest -q` with all optional dependencies — 382 passed
- Wheel import and packaged-schema smoke test from outside the checkout
- Offline import smoke test across 31 adapter/converter producer modules
- HF parser help and repeatable temporary-tree post-codegen smoke test


borgr:
From Claude, anything worth addressing? 1 is on purpose 3. seems like copilot already said. (It also obssessed on 208 that didn't get merged yet and would need updating.)
Bottom line
The mechanical refactor is well-executed and worth landing. The schema bump should come out (or be split), the DATA_DIR regression fixed, and it must update AGENTS.md/CONTRIBUTING.md/the skill to the new paths before merge.

Criticisms (most → least important)
1. The 0.2.3 → 0.3.0 bump is unjustified — the schema didn't change. The diff proves it: the only edits to eval.schema.json / instance_level_eval.schema.json are the version string; the eval_types.py / instance_level_types.py diffs are pure comment-text (post_codegen.py → every_eval_ever.post_codegen). A 0.2.3 record and a 0.3.0 record are structurally byte-identical. This misuses the version field (consumers gating on it will think the format changed), silently splits the datastore corpus (existing data stamped 0.2.3, new data 0.3.0, nothing documents they're identical), and contradicts the repo's own stated bump policy — the 0.2.2→0.2.3 bump was a field-optionality change; "refactor" is explicitly not a reason to bump. Recommend: drop the bump from this PR (keep 0.2.3), or split it into its own PR with a note that 0.3.0 ≡ 0.2.3 structurally.

2. No shim for the adapter invocation path — and the docs/skill that describe it are not updated. The Python API got wrappers, but python -m utils.<name>.adapter → python -m every_eval_ever.adapters.<name>.adapter just breaks, and AGENTS.md, CONTRIBUTING.md, and the eee-dataset-conversion skill are untouched by this PR (confirmed — only README.md and in-tree docstrings/READMEs are updated). Those files still say utils/<name>/adapter.py, utils/README.md, utils/openeval, etc. This is a textbook violation of the repo's own "if you change the invocation path / layout, update the skill in the same PR" rule — the exact governance our #208 is institutionalizing. Ironic, and a concrete drift bug regardless.

3. livecodebenchpro DATA_DIR regression. Path(__file__).parent.parent.parent / 'data' / ... → Path('data') / 'livecodebenchpro' makes it CWD-relative — the adapter breaks unless run from repo root. Copilot flagged this and the author suppressed it. The subtlety: the old parents[2] anchor is now wrong (it resolves to every_eval_ever/, not repo root), so a fix was needed — but the right fix is Path(__file__).resolve().parents[3] / 'data' / 'livecodebenchpro', not a CWD-relative path. Worth un-suppressing.

4. The bare-vs-prefixed instance schema_version inconsistency is left half-migrated. After this PR: instance schema doc = instance_level_eval_0.3.0 (prefixed), but the README example, test fixtures, and all converters emit the bare 0.3.0. It moved README+fixtures to match the code (good direction) but left the schema doc's own version prefixed — so doc and data now disagree, and nothing enforces either (the record's schema_version is a free string, no const/pattern). Clean resolution: make the schema doc bare too, or export an INSTANCE_SCHEMA_VERSION and use the prefixed form everywhere + add a const so validation actually catches drift.

Minor: data/global-mmlu-lite/ (hyphen, datastore convention) now diverges from the adapter module global_mmlu_lite (underscore, required for import) — harmless but a doc footnote.

andrewtran117 (approved):
looks good! verified @borgr 's comment above and additionally my own code review pass at it.. no breaking changes.

PR #221, opened 2026-08-05
Show validator warnings on files that pass

borgr:
## Problem

`report.warnings` is invisible in the CLI's default output whenever the file
passes. `render_report_rich` returns early on a valid report, and
`render_summary_rich` counts only errors, so a file with warnings prints a
plain green `PASS` and the run ends with `All N file(s) passed validation`.

The datastore review bot does not work that way. It consumes
`--format github`, which emits `::warning` annotations regardless of
validity, and `--format json` carries `warnings` too. So the contributor
path is: validate locally, see all green, submit, and get a review comment
about something their own tooling already knew.

## Change

- `render_report_rich` renders warnings on passing files as well as failing
  ones, with a yellow border to distinguish "passed, with notes" from a
  clean pass.
- `render_summary_rich` reports the warning total and says explicitly that
  warnings are not fatal but the review bot reports them — so a green run
  that will still draw a comment does not read as "nothing to do".

Exit code semantics are untouched: `main` still returns 1 only when a
report is invalid.

## Why now

Every entry in `REGISTERED_CHECKS` on `main` has severity `error`, so
`report.warnings` is always empty today and this PR changes nothing
observable on its own. That is also the reason it is worth landing first:
until warnings are visible locally, adding a `warning`-severity check gives
contributors a review comment with no local way to reproduce it. Two such
checks follow in #222 and #223; each stands alone, and neither depends on
this one for the bot's output.

## Tests

`TestWarningVisibility` in `tests/test_validate.py` renders a report through
a `Console` writing to a `StringIO` and asserts the warning text and its
location appear on a `PASS`, that the summary says `1 warning(s)` and `not
fatal`, and that a clean pass mentions no warnings at all.

### DIFF
--- a/every_eval_ever/validator/validate.py
+++ b/every_eval_ever/validator/validate.py
@@ -146,7 +146,12 @@ def _truncate(value: object, max_len: int = 80) -> str:
 
 
 def render_report_rich(report: ValidationReport, console: Console) -> None:
-    """Render one validation report as a terminal panel."""
+    """Render one validation report as a terminal panel.
+
+    Warnings are shown whether or not the file passed. A warning on a passing
+    file is exactly what the datastore review bot reports on a submission, so
+    dropping it locally sends contributors into review with surprises.
+    """
     if report.valid:
         label = Text(' PASS ', style='bold white on green')
         kind = (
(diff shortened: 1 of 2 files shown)


mrshu:
# ⚒️ review-anvil report

**Review decision:** COMMENT — Approval is disabled for this run; no confirmed problems were found.
**Result:** The warning output change is focused, non-fatal, and covered by rendering tests.
**Scope:** This PR shows validator warnings in rich per-file and summary output while preserving exit behavior.
**Checks:** skipped (no findings needed checking)
**Second check:** off

## Earlier review comments

## What I noticed

No confirmed problems found.

No changes were made, and there are no suggested fixes.

<details>

- Target: PR #221 (`validator-warnings-visible`, 2 files, +121/-17)
- Run ordinal: 1
- Rounds: 1/1 completed; adaptive off; clean result
- Mix: 3 codex-exec
- Finding counts: 0 critical, 0 high, 0 medium, 0 low, 0 nit
- Checks: skipped
- Second check: off; auto selected off because the comment-only result was clean
- Set aside: 0 items

</details>

---

nelaturuharsha:
The warning-visibility fix is now included in #238, reconciled with the standalone local validator interface. The package CLI uses PASS/WARN/FAIL states, shows warning-only findings and counts, and keeps the existing non-fatal exit behavior. #238 also adds the grouped local-folder view and detailed JSON log.

PR #222, opened 2026-08-05
Warn when metric_id names no metric at all

borgr:
## Problem

`metric_id` is the join/dedup key for scores from different leaderboards, so
it only works if it means the same thing everywhere. Nothing enforced it, and
"use a real `metric_id`" has been one of the most repeated review comments on
incoming adapters.

## Change

`check_metric_identity`, registered as a **warning**-severity check, fires
when `metric_id` is:

1. **missing** — the records cannot be joined at all;
2. **a word that names no quantity** — `score`, `rank`, `elo`, `mean_score`,
   `overall`, `total`, `value`, `cost` and a few more, so whatever each source
   called its headline number joins under one id;
3. **identical to `evaluation_name`** — a task id standing in for a metric id.

Warning, not error: `metric_id` is optional in the schema and published
records predate the rule. Severity lives in `REGISTERED_CHECKS`, so the
datastore bot picks it up with no change on its side.

**Exit code, after #238.** `validate` now returns `2` on warnings-only, so
records this check flags no longer exit `0`. That also means the shared
`valid_aggregate()` / `VALID_AGGREGATE` fixtures had to gain a `metric_id`:
#238's `test_canonical_record_is_merge_ready` asserts `warnings == []` on that
fixture, and the fixture carried no `metric_id` at all. The three tests that
need the missing case `pop` it locally, so the shared fixture stays the clean
record it now claims to be.

## What this check deliberately does not do

It does not decide whether an id names a *real* metric. That needs a
vocabulary, and the one that exists is the eval-card-registry's — which this
check then has to agree with.

The registry namespaces **6 of its 1842 metric ids** (read-only
`/api/v1/metrics`), so a rule demanding a namespace flags the other 1836:

- it spells canonical metrics with dashes — `exact-match`, `rouge-l`,
  `pass-at-1`, `win-rate`;
- it qualifies a benchmark-specific metric with a prefix rather than a dot —
  `mteb-score`, `mmau-pro-open-ended-judge-score`.

Growing an in-repo whitelist does not fix it either, because no list holds
every field's metrics: `psnr`, `cider`, `iou`, `stoi` are all specific and none
would be in it. A check that flags the registry's own ids drives contributors
off them, which inverts the intent — and contradicts the conversion skill
merged in #208, which asks for a registry-canonical id for a global metric and
a namespaced one only for a leaderboard-specific construct.

So the rule is the narrow one that holds offline: warn only where two
sources' numbers really would collide.

Six registry ids do still warn — `average`, `elo`, `mean-score`, `overall`,
`rank`, `score`, four of them `reviewed`. That is the rule working rather
than missing: an Elo or a rank is only comparable inside one leaderboard's
pool, so the bare id is exactly the ambiguity being reported. **Those six
registry entries are worth revisiting too** — flagging them here is not a
claim that the registry is wrong to hold them, only that a record should not
publish under them bare.

## Measured against published data

Every record at canonical depth on `evaleval/EEE_datastore` `main`
(`f8a7acdb`) — 13,184 files and 112,298 results, not a sample:

| | |
| --- | --- |
| records warning | **3,132** of 13,184 (23.8%) |
| …missing `metric_id` | 1,722 |
| …`metric_id` repeats `evaluation_name` | 882 |
| …`metric_id` names no quantity | 528 |
| non-string `metric_id` (the schema-deferral guard) | **0** |

**Collapsing to one warning per finding kind.** A leaderboard file has one
entry per task and every entry comes from the same adapter code, so a
per-result warning repeats itself once per task. Over this corpus that is
**54,422 lines instead of 3,132** — 374 identical lines on the largest
published record alone, burying every other finding. The count and the first
location are what a contributor needs.

**Narrowing to generic words, rather than requiring a namespace.** A
"must be namespaced" rule would warn on **8,193 records (62%)** where this one
warns on 3,132. The 13 ids that stop being flagged are all specific
quantities, and the two biggest are the ones that matter:

| id | results affected |
| --- | --- |
| `accuracy` | 31,328 |
| `exact_match` | 4,574 |
| `cost_per_task`, `win_rate`, `mean_attempts`, `latency_mean`, `pass_at_k`, `standard_error`, `average_refusal_rate`, `cost_per_100_calls_usd`, `standard_deviation`, `average_full_score`, `average_score_non_refusals` | 2,036 combined |

`accuracy` is the registry's own canonical id. Telling 31,328 results to
namespace it would fragment the single id that already means one thing
everywhere.

**Separators.** A namespaced id is never generic, whichever separator it
uses. `.` is the one the schema documents, but
`every_eval_ever/tools/hf_community_evals.py` accepts `hle.accuracy` and `hle/accuracy`
interchangeably and the datastore is full of `mmlu_pro/overall` and
`mt_bench/turn_1`. Normalization unifies `-` and `_` so the registry's
`mean-score` matches an adapter's `mean_score`, and stops at `.` and `/` so
`mmlu_pro/overall` stays qualified. Finding 3 normalizes **both** sides, so a
`metric_id` of `Bench ` against an `evaluation_name` of `bench` is caught —
neither spelling is canonical, so neither can be the one that is compared raw.

## Left alone deliberately

Four in-tree adapters trip the warning (unchanged after #157 and #160 landed —
the Vectara and LEXam adapters both namespace or use registry ids).
`metric_id` is a join key, so rewriting one means republishing the data under
it — a per-adapter decision, not a drive-by:

- `arc_agi` — `score`, `cost`
- `sciarena` — `elo`, `rank`
- `mercor_eval` — `mean_score`
- `artificial_analysis` — `metric_id` is set to the already-namespaced
  `evaluation_name`, so it trips finding 3

## Tests

`tests/test_validation_scope.py`: each of the three findings; generic words in
either separator spelling and either case; registry-shaped ids and specific
quantities accepted (`exact-match`, `mteb-score`, `latency_mean`, `psnr`…); a
generic word behind a namespace accepted (`mmlu_pro/overall`, `lmarena.elo`);
finding 3 firing on either side's spelling of the same name; a non-string
`metric_id` left to the schema rather than reported as missing; one warning per
finding kind with the count and first location; and that the warning does not
fail validation.

## Note for whoever merges the validator warnings

#221 is no longer the dependency it was — it merged and was reverted by #239.
#238 landed the visible-warning rendering instead, so these warnings show in
the default `rich` output on current `main`; `--format github` and
`--format json` carried them all along.

#222, #223 and #230 each add an entry to `REGISTERED_CHECKS`, to the import
list in `validate.py`, and to the end of `tests/test_validation_scope.py`, so
whichever merges second conflicts on the adjacent lines; the resolution is to
keep all entries. Once this lands, the other two branches will need the `metric_id` added to the
shared fixtures here when they merge `main` — #238 asserts that fixture warns
about nothing.


mrshu:
# ⚒️ review-anvil report

**Review decision:** COMMENT — Approval is disabled; two small input-normalization suggestions remain.
**Result:** No confirmed blocker was found. Two material concerns were set aside because exact-head verification could not complete.
**Scope:** This PR adds a warning-only check for missing, generic, or evaluation-name-shaped metric IDs.
**Checks:** 2 concerns checked; 0 confirmed, 0 ruled out, 2 set aside, 0 lowered in priority.
**Second check:** off

## Earlier review comments

## What I noticed

<details>
<summary>Non-blocking low findings (2 items)</summary>

- **RAV-RUN1-R1-F003 [low] normalization** `every_eval_ever/validator/validation_core.py:677-683` — The comparison strips `metric_id` but not `evaluation_name`. Matching values with surrounding whitespace can miss this warning.
- **RAV-RUN1-R1-F004 [low] malformed-inputs** `every_eval_ever/validator/validation_core.py:673-675` — Non-string IDs enter the missing-ID branch. Schema-invalid numeric values can therefore receive both a type error and an inaccurate missing-ID warning.

</details>

- **[low] normalization** — The equality check can normalize both strings and cover surrounding whitespace. (`RAV-RUN1-R1-P001`; covers `RAV-RUN1-R1-F003`)
- **[low] malformed-inputs** — The missing warning can stay limited to absent, null, or blank values while schema validation handles other types. (`RAV-RUN1-R1-P002`; covers `RAV-RUN1-R1-F004`)

- **RAV-RUN1-R1-F001 [high] warning-integration** — We set this aside because the check's relationship to PR #221 could not be verified at the exact PR head.
- **RAV-RUN1-R1-F002 [medium] warning-collapse** — We set this aside because the grouped output and test expectations could not be inspected at the exact PR head.

<details>

- Target: PR #222 (`validator-metric-id-warning`, 3 files, +325/-0)
- Run ordinal: 1
- Rounds: 1/1 completed; adaptive off; low findings only after checks
- Mix: 3 codex-exec; 1 completed exact-head review, 2 failed because exact-head access was unavailable
- Finding counts: 0 critical, 0 high, 0 medium, 2 low, 0 nit
- Checks: concerns=2; confirmed=0/ruled-out=0/set-aside=2/lowered=0
- Second check: off; auto selected off because only low suggestions remained and approval was disabled
- Set aside: 2 items; reason=exact-head verification could not complete

</details>

---

borgr:
Both low findings are implemented, and the two set aside for exact-head access are answerable now.

**RAV-RUN1-R1-F003 / P001 — normalize both sides of the `evaluation_name` comparison.** Done. `_normalize_metric_id` is now applied to `evaluation_name` as well, so `' Bench '` against `'bench'` is caught. Neither side is canonical — a `metric_id` is no more authoritative than the task name it duplicates — so comparing one raw against one normalized was arbitrary in the direction it happened to pick. `test_metric_id_must_name_a_quantity` asserts the finding for `'bench'`, `' Bench '` and `'bench '` against an `evaluation_name` of `bench`.

**RAV-RUN1-R1-F004 / P002 — keep the missing branch to absent, null, and blank.** Done. A `metric_id` that is present but not a string now returns no finding: it is already a schema error, and "missing" would name the wrong fix for a field that is populated. Blank-after-strip stays in the missing branch, since that is schema-valid and nothing else would report it. Asserted inline in the same test with `metric_id = 12`.

**RAV-RUN1-R1-F001 — relationship to #221.** #221 was merged and then reverted by #239, so it is not a dependency any more. #238 landed the visible-warning rendering instead, and it also made `validate` return `2` on warnings-only. So this check now reports in the default `rich` output *and* changes the exit code of any record it flags — the description states that up front rather than leaving the old "exit code stays 0" claim standing.

That exit contract had one concrete consequence here: the shared `valid_aggregate()` / `VALID_AGGREGATE` fixtures carried no `metric_id` at all, while #238's `test_canonical_record_is_merge_ready` asserts `warnings == []` on that fixture. I gave the fixtures a `metric_id` and had the three tests that need the missing case `pop` it locally, rather than relaxing #238's assertion — the fixture is meant to be the merge-ready record, so it should not be one this check would warn about.

**RAV-RUN1-R1-F002 — the grouped output.** One warning per finding *kind*, not per result, with the count and the first location. I re-measured the reason on the whole published corpus rather than a sample: over all 13,184 canonical-depth records on `EEE_datastore` `main` (`f8a7acdb`), per-result warnings would be **54,422 lines** where the grouped form is **3,132** — and 374 of those lines would land on one record. The tests cover the grouping directly (`test_one_warning_per_finding_carries_the_count_and_first_location`).

**Blast radius, stated honestly.** 3,132 of 13,184 published records warn — 23.8%. That is a large number and it is the point of the check rather than a surprise: 1,722 have no `metric_id` to join on at all. The two alternatives I measured are both worse. Requiring a namespace instead would warn on 8,193 records (62%) and would flag `accuracy` on 31,328 results — the registry's own canonical id. Requiring membership in a vocabulary would flag 1,836 of the registry's 1,842 ids, since it namespaces only 6 of them.

Full suite green against current `main`: 416 passed / 20 skipped, `ruff` clean.

mrshu:
# ⚒️ review-anvil report

**Review decision:** COMMENT — This pass is comment-only; the remaining feedback is low-priority warning UX and guidance.
**Result:** The validator rule is correctly scoped and integrated. Three small suggestions can make contributor feedback more complete.
**Scope:** Warn when aggregate `metric_id` values do not name a metric, while preserving schema errors and validator output behavior.
**Checks:** skipped; the main finding had three-reviewer consensus.
**Second check:** 1 reviewer kept all 3 suggestions and lowered warning aggregation to low priority.

## Earlier review comments

- **Fixed:** warning registration, aggregate scope, CLI rendering, whitespace symmetry, and non-string schema-error handling are present now.
- **Still present at lower priority:** `RAV-RUN1-R1-F002` now has exact-head evidence for mixed generic IDs.

## What I noticed

No blocking problems were found.

<details>
<summary>Non-blocking low findings (3 items)</summary>

- **RAV-RUN1-R1-F002 [low] warning aggregation** — Mixed IDs such as `elo` and `rank` collapse to the first value plus a count. Contributors need another validation run to see the next fix.
- **RAV-RUN2-R1-F001 [low] normalization** — Internal whitespace does not normalize like hyphens and underscores, so `mean score` bypasses the intended heuristic.
- **RAV-RUN2-R1-F002 [low] contributor guidance** — The registered validator rule is not yet documented in the required `datastore-gate.md` reference.

</details>

## Things to try

- **[low] warning aggregation** — Distinct normalized IDs can retain first-seen order, first location, raw value, and per-ID count. (`RAV-RUN2-R1-P001`; covers `RAV-RUN1-R1-F002`)
- **[low] normalization** — Internal word spaces can normalize without changing `.` or `/` namespace separators.
- **[low] guidance** — A warning-marked gate section can name the three triggers and link to metric-ID field guidance.

<details>

- Target: PR #222 (`validator-metric-id-warning`, 4 files, +324/-0) at `db4bedc6d3f71fde244c71bb123f6e19cb4e1638`
- Run ordinal: 2
- Mix: 3 codex-exec
- Finding counts: 0 critical, 0 high, 0 medium, 3 low, 0 nit
- Checks: skipped; consensus finding only
- Second check: challenge; reviewers=1; kept=3, lowered=1, removed=0

</details>

---

borgr:
The two new low findings are implemented; the re-raised aggregation one is intentional, and I'll say why it stays.

**RAV-RUN2-R1-F001 — internal whitespace not normalized.** Fixed (`96a2d261d`). `_normalize_metric_id` now collapses internal whitespace to the same separator as `-` and `_`, so `mean score` normalizes to `mean_score` and is caught by the generic-id set. `.` and `/` are still left alone as namespace separators, so a qualified `mmlu_pro/overall` does not warn. `test_metric_id_must_name_a_quantity` now asserts the finding for `'mean score'` and `'Total Score'`.

**RAV-RUN2-R1-F002 — rule not documented in `datastore-gate.md`.** Fixed (`e6f26a4bf`). Added a `§metric-id` section that names the three triggers (missing/blank, generic, repeats `evaluation_name`), the separator/namespace matching rule, and the fact that it whitelists nothing (any specific id passes, only ids that name no quantity are flagged), with a pointer to the metric-id field guidance.

**RAV-RUN1-R1-F002 — mixed IDs collapse to first-plus-count.** Kept as is, deliberately. Grouping per finding *kind* is the design, not an oversight: over the whole published corpus (13,184 canonical-depth records on `EEE_datastore main`), per-result warnings are 54,422 lines against 3,132 grouped, with 374 of those lines landing on a single record — one line per finding kind is what keeps the output usable, and `test_one_warning_per_finding_carries_the_count_and_first_location` pins that shape. Listing every distinct bad id per kind (RUN2-P001) trades back toward the per-result volume the grouping exists to avoid; re-running `validate` after each fix already surfaces the next value without spending more warning budget. If the one-fix-per-run cadence turns out to be real friction in practice I'll revisit the balance, but I'd rather not widen the default output on the `elo`/`rank` example alone.

borgr:
Reverted on main in #266.

Not because the rule is wrong. #244 landed the merge-gate case that requires a converted record to validate with **no warnings**, and the lm-eval converter emits no `metric_config.metric_id` at all, so this warning turned that case red the moment it merged (`test_conversion_passes_the_merge_gate[lm_eval]`). The missing piece is the ids, not the check.

#246 is the PR that gives every converted metric a joinable id. Order is #246 first, then this re-lands unchanged — the branch `validator-metric-id-warning` is untouched, so it is a fresh PR off the same commits.

PR #223, opened 2026-08-05
Warn when a record's identity and its directory address different models

borgr:
## Review lane

**Needs a human.** It adds a registered validator rule, so the same input now produces a report it did not before — a material change in outcome even though the diff is one file plus tests.

Design agreed in: no prior issue. The rule follows the shape of the checks already in `REGISTERED_CHECKS`, so if the shape is wrong the objection is to all of them, not just this one.

## Checklist

- [x] `validate` clean at the final `data/<collection>/<dev>/<model>/` path on the reference conversion
- [x] offline unit tests added; full `pytest tests` green (419 passed / 20 skipped) against current `main`
- [x] `ruff check` clean

**Exit code, after #238.** `validate` now returns `2` on warnings-only, so a record this check flags no longer exits `0`. Two fixtures merged with #238 claim `model_info.id: 'org/test-model'` while filing under `data/benchmark/developer/model/` — exactly the disagreement this check reports — and so began exiting `2` against their `== 0` assertions. This branch fixes them by filing each record where its own identity says (`data/benchmark/org/test-model/`), not by weakening the assertion.

## Problem

`data/<collection>/<developer>/<model>/` is how the datastore is queried, so a
reader who looks a model up by path finds only the records filed under the
spelling they guessed. Contributors who place files by hand choose the developer
slug themselves, and the slug they choose often disagrees with `model_info.id` —
a record whose id says `dev-labs/model` filed under `dev/model/`. The same model
then exists twice, and neither listing is complete. This is one of the most
common corrections in submission review.

Nothing caught it. `check_path_structure` validates the *shape* of the path
(depth, reserved names, portable filenames) but never its agreement with the
record inside it, and the schema has no opinion about where a file lives.

## Change

A new `model identity path` aggregate check re-derives the directory the
identity addresses, using the publisher's own helper
(`helpers.io.datastore_path_components`), and compares it to the actual one:

```
model_info: id 'dev-labs/model' addresses bench/dev-labs/model/ but the file
is in bench/dev/model/. The datastore is queried by path, so one model under
two paths is two models to every reader — make the identity and the directory
agree on the spelling this collection already uses, rather than publishing
under both
```

No heuristics and no vocabulary of its own. A flat `model_info.id` falls back to
`model_info.developer` the same way the publisher does, so records that never
carried a slash-bearing id are unaffected.

Five deliberate non-behaviours, so each problem is reported once by whichever
check owns it:

- when the path is not the expected depth, the check returns nothing and
  lets `check_path_structure` explain the path — no duplicate complaints
  about one file;
- when `model_info.id` is missing or blank, the check returns nothing: the
  schema already requires it, and semantic checks run *even when schema
  validation has already failed*, so without this guard a contributor with an
  incomplete draft would get a warning about a field the errors above already
  name. No published record is affected (0 of 13,184) — this is purely
  contributor-facing friction, which is the audience that matters most here;
- likewise when `model_info.id` is present but not a string, and when a *flat*
  id's `model_info.developer` is present but not a string. Both are already
  schema errors, and this check would restate them as "`model_info.developer`
  is required" about a field that is populated. The developer guard is scoped to
  flat ids on purpose: a `dev/model` id never reads that field, so its drift
  stays reportable regardless of what `developer` holds (covered by a test);
- the collection component is never re-derived. `data/unknown/dev/model/`
  passes `check_path_structure` but `datastore_path_components` refuses the
  collection, and a `model_info:`-prefixed message about a directory two levels
  above the field would send the reader somewhere they cannot fix it. A
  placeholder collection is passed to the helper so only the developer and
  model components are compared;
- an identity that cannot name a directory at all makes
  `datastore_path_components` raise `ValueError`; the check reports the
  publisher's own reason rather than letting it escape, so one malformed record
  cannot abort a batch validation. The message quotes `model_info.id` because
  the publisher's wording alone is ambiguous about which field to edit —
  "model developer must be known" means the id's namespace prefix when the id
  has one and `model_info.developer` when it does not, and those are different
  fixes.

## What this warns on, measured on the published datastore

Every record at canonical depth on `evaleval/EEE_datastore` `main`
(`f8a7acdb`) — 13,184 files, not a sample:

| | |
| --- | --- |
| files warning | **851** of 13,184 (6.5%) |
| …addressing a directory other than the one they sit in | 582 |
| …addressing no directory at all | 269 |
| non-string `id` or `developer` (the schema-deferral guards) | **0** |

The 269 that address no directory, by the publisher's own reason: 96 carry an id
prefixed `unknown`, 92 a flat id with no `developer`, 81 an `unknown`
`developer`. No published record trips the non-portable-component branch; the
publisher still refuses one, which is why the check reports instead of raising.

## Why the message names both sides and picks neither

An earlier revision of this branch said the record *belongs in* the directory
its identity addresses, and told the reader to republish through
`publish_evaluation_logs`. Measured against the published datastore, that
advice is wrong for already-published files. Of the 582 that address a
different directory:

| | |
| --- | --- |
| whose addressed directory already exists | **2** |
| name a developer directory absent from their collection | 390 |
| …of those, absent from the datastore entirely | 104 |
| same developer directory, model-name component differs | 192 |

Following "publish through `publish_evaluation_logs`" for the other 580 adds a
second directory for a model that already has one — exactly the split this
check exists to report.

There is also no datastore-wide convention to appeal to, so the check cannot
quietly normalize instead:

| | |
| --- | --- |
| developer directories, lowercase / cased | 1375 / 404 |
| model directories containing uppercase | 4123 of 9589 |
| model directories containing a dot | 3077 of 9589 |
| collections using both developer conventions at once | 15 of 102 |

So the message reports the disagreement, names both directories, and asks for
agreement on whatever the collection already uses. That statement is true
regardless of which spelling wins, which a per-file warning cannot decide.

**Open question for maintainers** (not addressed here): `datastore_path_components`
preserves case and dots, so `anthropic/claude-opus-4.6` addresses
`claude-opus-4.6/` while the published tree has `claude-opus-4-6/`. Whether the
publisher should slugify, or the tree should be migrated, or neither, is a
datastore-wide decision. Until it is made, this check reports the disagreement
without taking a side.

## Severity: warning, not error

Already-published records predate the rule, and an identity/path
disagreement is a filing mistake rather than invalid data — the file parses,
validates, and carries a usable identity. Promoting it to `error` once the
datastore is known to be clean needs no code beyond the severity field in
`REGISTERED_CHECKS`; the review bot already renders warnings from
`--format github`, so nothing changes on the bot side either way.

## Tests

`tests/test_validation_scope.py` covers a matching path, developer-slug drift
(asserting both directories appear in the message and that it does not claim the
record "belongs" anywhere), the flat-id fallback, each of the four ways an
identity can fail to name a directory (flat id with no developer, `unknown` as
the developer, `unknown` as the id prefix, a component that is not portable),
deferral to the path-structure check on a malformed path, deferral to the schema
on a missing, blank, or non-string id, deferral on a non-string developer under
a flat id *together with* the same value staying reportable under a
slash-bearing one, that a collection the publisher would refuse is never
reported under `model_info`, and that a flagged record still validates with the
finding in `report.warnings`.

While measuring this I found a gap in `check_path_structure`, which is
pre-existing on `main` and **not** fixed here: it accepts a collection directory
named `unknown`, which `datastore_path_components` refuses, so an `error`-severity
path check passes a path the publisher would reject. Folding that into a
warning-severity PR would mix an error-behaviour change into it; it wants its own
small PR.

## Note for whoever merges

The `metric identity` check in #222 and the `developer slug` check in #230
append to the same `REGISTERED_CHECKS` tuple, and to the same import list in
`validate.py` and the end of `tests/test_validation_scope.py`. Whichever lands
second conflicts on the adjacent lines; the resolution is to keep all entries.

Measured over the same 13,184 records, #230 warns on 361 and this check on 851;
they agree on 61 files, and each sees files the other does not (790 here, 300
there). They are close to independent, so landing both is not double-reporting.

#221 is no longer a dependency: it was merged and then reverted by #239.
Warnings are visible in the default `rich` output through #238 instead, and
`--format github` (what the review bot uses) and `--format json` carried them
all along.


every_eval_ever/validator/validation_core.py
```diff
@@ -577,6 +578,79 @@ def _metric_bound(value: Any) -> float | None:
...
+    """
+    parts = repo_path.split('/')
+    if len(parts) != _EXPECTED_PATH_PARTS:
+        return []  # the path-structure check already reported this
+    model_info = data.get('model_info')
+    if not isinstance(model_info, dict):
+        return []
+    model_id = model_info.get('id')
+    if not isinstance(model_id, str) or not model_id.strip():
+        return []  # the schema already requires model_info.id
```
mrshu: **RAV-RUN1-R1-F001 [medium] schema-deferral** — Blank model IDs bypass the new identity-path check

The schema accepts blank strings, but this guard returns before `datastore_path_components` can reject them. Such a record can pass without any warning that its identity names no directory.

Missing or non-string IDs can stay with schema validation, while blank strings can use the helper's focused error path.

every_eval_ever/validator/validation_core.py
```diff
@@ -577,6 +578,79 @@ def _metric_bound(value: Any) -> float | None:
...
+    the 46 addresses a directory that already exists, and 30 name a developer
+    directory absent from their collection, so telling those records to move
+    would add a second directory for a model that already has one — the split
+    this check exists to report.
+
+    Warnings, not errors, because already-published records would fail.
+    """
+    parts = repo_path.split('/')
+    if len(parts) != _EXPECTED_PATH_PARTS:
+        return []  # the path-structure check already reported this
```
mrshu: **RAV-RUN1-R1-F003 [medium] path-check-ownership** — Other malformed paths can receive both a path error and an identity warning

This guard checks only path depth. A five-component path with a nonportable developer or model component continues into the identity comparison, so both checks report the same bad path.

Identity comparison can stop whenever `check_path_structure` already owns a path error.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — Approval is disabled; two non-blocking path-validation gaps remain.
**Result:** The publisher-helper reuse is sound, with two focused edge cases worth covering.
**Scope:** This PR warns when model identity and datastore directory components address different models.
**Checks:** 1 concern checked; 1 confirmed as a merge-order dependency and removed from actionable findings.
**Second check:** targeted, 2 reviewers; both could not reach the exact PR head, so the normal synthesis remains.

## Earlier review comments

## What I noticed

- **RAV-RUN1-R1-F001 [medium] schema-deferral** `every_eval_ever/validator/validation_core.py:620-621` — Blank model IDs return before the publisher helper runs. The schema accepts blank strings, so these records can pass without a warning even though their identity names no datastore directory. (inline)
- **RAV-RUN1-R1-F003 [medium] path-check-ownership** `every_eval_ever/validator/validation_core.py:613-615` — The check defers only wrong path depth. A five-component path with a nonportable developer or model component can receive both the path error and an identity-drift warning for the same malformed component. (inline)

- **[medium] schema-deferral** — Missing or non-string IDs can stay with schema validation, while blank strings can reach the publisher helper and produce its focused warning. (`RAV-RUN1-R1-P001`; covers `RAV-RUN1-R1-F001`)
- **[medium] path-check-ownership** — Identity comparison can stop whenever `check_path_structure` already reports a path error. (`RAV-RUN1-R1-P002`; covers `RAV-RUN1-R1-F003`)

- **RAV-RUN1-R1-F002 [medium] contributor-guidance** — PR #221 must merge before or with this PR for valid-record warnings to appear in default rich output. JSON and GitHub formats already show them.

<details>

- Target: PR #223 (`validator-identity-path-drift`, 3 files, +184/-0)
- Run ordinal: 1
- Mix: 3 codex-exec
- Checks: concerns=1; confirmed merge dependency=1; actionable=0
- Second check: targeted; reviewers=2; both failed exact-head access; normal synthesis retained; approval changed no
- Set aside: 1 item; reason=explicit merge-order dependency on PR #221

</details>

---

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The earlier ownership gaps are fixed, and one focused schema-deferral suggestion remains.
**Result:** One medium-priority suggestion remains. No high-priority problems were confirmed.
**Scope:** Add a warning-level validator check for model identity and datastore path disagreement while preserving existing check ownership.
**Checks:** skipped; two independent reviewers found the same concern.
**Second check:** targeted review was unavailable because both reviewers lost network access. The confirmed normal-review result remains.

## Earlier review comments

- `RAV-RUN1-R1-F001` is fixed. Blank IDs now reach the publisher helper and produce the intended warning.
- `RAV-RUN1-R1-F003` is fixed. The full path-structure result now gates this check.
- `RAV-RUN1-R1-F002` remains deliberately set aside. PR #221 is still the planned rich-output integration change.

## What I noticed

|---|---|---|---|---|
| RAV-RUN2-R1-F001 | medium | schema deferral | `every_eval_ever/validator/validation_core.py:621` | A flat string model ID with a non-string developer produces both a schema error and an identity-path warning. Semantic checks receive the raw dictionary and pass that developer value to the publisher helper. |

## Things to try

- **[medium] schema deferral** — For a flat ID, the identity check can defer when the developer is present but is neither a string nor `None`. Schema-valid blank, unknown, or unsafe string values can keep the existing warning path. A schema-error-only test would cover the boundary. (`RAV-RUN2-R1-P001`; covers `RAV-RUN2-R1-F001`)

<details>

- Target: PR #223 at `f24bb0778539dd0b65db9a543db27f19d85d0dec` (3 files, +190/-0)
- Run ordinal: 2
- Mix: 3 codex-exec
- Earlier review comments: 5 ledger entries grouped into 3 topics; 2 fixed and 1 deliberately set aside
- Finding counts: 0 critical, 0 high, 1 medium, 0 low, 0 nit
- Checks: skipped because two reviewers agreed on the material finding
- Second check: targeted; reviewers=2; both failed before verdict because network access was unavailable; approval unchanged
- Set aside: 1 prior integration item remains deliberately deferred to PR #221

</details>

---

every_eval_ever/validator/validation_core.py
```diff
@@ -577,6 +578,69 @@ def _metric_bound(value: Any) -> float | None:
...
+
+    collection, developer, model = repo_path.split('/')[1:4]
+    try:
+        # Ask the publisher where this identity files, so the check cannot
+        # drift from datastore_output_dir. Only the developer and model
+        # components are compared, so a placeholder stands in for the
+        # collection: a collection directory the publisher would refuse is not
+        # this check's finding, and reporting it under model_info would send
+        # the reader to the wrong field.
+        _, expected_developer, expected_model = datastore_path_components(
```
mrshu, line 631: **RAV-RUN2-R1-F001 [medium] schema-deferral** — A schema-invalid developer also produces an identity-path warning

Semantic checks still receive the raw dictionary after model validation fails. With a flat string ID and a non-string developer, this path passes the raw value to `datastore_path_components()`. The result is both the schema error and a warning about the same field.

For flat IDs, the check can defer when the developer is present but is neither a string nor `None`. Schema-valid blank, unknown, or unsafe strings can keep the current warning path.

borgr:
Both open findings are addressed, and merging current `main` in surfaced a third thing worth flagging.

**RAV-RUN2-R1-P001 — defer on a non-string `developer` for a flat id.** Done. The check now returns nothing when a flat `model_info.id` carries a `developer` that is present but not a string; blank, `unknown`, and unsafe string values keep the existing warning path, since those are schema-valid and the check is the only thing that will say anything about them. The guard is scoped to flat ids deliberately — a `dev/model` id takes its developer from the id and never reads that field, so its drift stays reportable no matter what `developer` holds. `test_a_non_string_developer_is_left_to_the_schema` asserts both halves: the flat id produces no finding while `validate_data` reports exactly `model_info -> developer` and no warnings, then the same value under `dev-labs/model` still reports the directory disagreement.

**RAV-RUN1-R1-F002 — warnings invisible in `rich` output.** No longer a dependency on #221, which was merged and then reverted by #239. #238 landed the visible-warning rendering instead, so this check reports in the default output on current `main`.

**What #238 changes for this PR, which the description now states up front.** `validate` returns `2` on warnings-only. Two fixtures that came in with #238 — `test_exit_code_0_on_clean_pass` and `test_cli_accepts_absolute_local_datastore_path` — declare `model_info.id: 'org/test-model'` and file the record under `data/benchmark/developer/model/`, which is precisely the disagreement this check reports, so they started exiting `2` against their `== 0` assertions. I fixed them by filing each record where its own identity says (`data/benchmark/org/test-model/`) rather than by relaxing the assertion: the fixtures are meant to be the clean case, and a clean case that contradicts itself is not one. Full suite is 419 passed / 20 skipped.

**Re-measured on the whole datastore, not a sample.** The description's numbers came from one record per `(collection, developer)` group; I have since validated every record at canonical depth on `EEE_datastore` `main` (`f8a7acdb`, 13,184 files). 851 warn (6.5%): 582 address a different directory, 269 address none. Zero records carry a non-string `id` or `developer`, so both schema-deferral guards are contributor-facing hygiene rather than anything that changes the published picture — which is the argument for having them.

The republish-instead-of-report question also came out sharper than the sample suggested: of the 582, the addressed directory already exists for **2**. Republishing the other 580 through `publish_evaluation_logs` would create a second directory for a model that already has one, so the message still names both sides and picks neither.

Overlap with #230, same corpus: #230 warns on 361, this on 851, agreeing on 61. Each sees files the other does not (790 / 300), so landing both is not double-reporting.

mrshu:
# ⚒️ review-anvil report

**Review decision:** COMMENT — The validator behavior is now sound, and one contributor-guidance suggestion remains.
**Result:** One medium-priority documentation finding remains. No code-path defect was confirmed.
**Scope:** Add a warning-level model identity/path check while preserving schema ownership, path-check ownership, and contributor workflow guidance.
**Checks:** 1 concern checked and confirmed. A separate blank-ID claim was ruled out against the exact head.
**Second check:** 2 reviewers kept the guidance finding and clarified its smallest fix.

## Earlier review comments

- `RAV-RUN1-R1-F001` is fixed. Blank strings now reach `datastore_path_components`, with a direct regression test.
- `RAV-RUN1-R1-F003` is fixed. The full path-structure result gates this check.
- `RAV-RUN2-R1-F001` is fixed. Schema-invalid flat-ID developers now defer to schema validation.
- `RAV-RUN1-R1-F002` is no longer relevant. Current warning output is visible, and warning-only validation exits with status `2`.

## What I noticed

|---|---|---|---|---|
| RAV-RUN3-R1-F001 | medium | contributor guidance | `.claude/skills/eee-dataset-conversion/reference/datastore-gate.md` | The canonical contributor gate guide does not describe the new model identity/path warning or the warning-only exit status. The reference fixture already satisfies the rule, so its green test does not expose this omission. |

## Things to try

- **[medium] contributor guidance** — `datastore-gate.md` can add one warning-marked identity rule. Its general CLI section can state the full exit precedence: clean `0`, errors `1`, and warnings without errors `2`. The rule can stay out of other references. (`RAV-RUN3-R1-P001`; covers `RAV-RUN3-R1-F001`)

<details>

- Target: PR #223 at `d404e5ff29f9c9cac1bfdb5bb36c48bad431eb3d`
- Run ordinal: 3
- Mix: 3 codex-exec
- Earlier review comments: 8 ledger entries grouped into 4 topics; all fixed or no longer relevant
- Finding counts: 0 critical, 0 high, 1 medium, 0 low, 0 nit
- Checks: concerns=1; confirmed=1; ruled-out=1; set-aside=0
- Second check: targeted; reviewers=2; kept=1; clarified=1 plan; removed=0; approval unchanged
- Set aside: 0 items

</details>

---

borgr:
RUN3 **RAV-RUN3-R1-F001** (medium, contributor guidance) — fixed, both halves.

- **Identity rule** — `datastore-gate.md` now carries a `§identity — check_model_identity_path (warning)` section (40cf69da0): what it warns on (a record filed under a directory its `model_info` doesn't address, and identities that name no directory at all), why it's a warning not an error (published records already span cased and lowercase spellings, so there's no convention to appeal to), that it derives the expected path from `datastore_path_components` so it can't drift, and that it stays silent when §path or a schema error already covers the case.
- **Exit precedence** — the CLI item now states the machine verdict (941c70e00): `0` clean, `1` when any record has errors (errors win over warnings), `2` when the only findings are warnings — matching `validator/validate.py`.

The rule stays out of the other references, as suggested.

borgr:
Reviewed the 3 unresolved automated review threads against the current head. All three are stale or already satisfied in the current `validation_core.py`. Resolving them.

PR #224, opened 2026-08-05
Test that the `all` extra installs every optional extra

borgr:
## Problem

Adding an optional extra takes two edits: the extra itself, and a
self-reference from the `all` bundle. Only the first is needed for CI to
pass — CI installs with `uv sync --all-extras`, which resolves every extra
directly and never consults `all`. So the second edit is easy to forget, and
the omission is invisible in this repository. The only people who see it are
users who installed `every-eval-ever[all]` and then hit an `ImportError` on
an adapter that CI exercises happily.

It nearly happened on my own open PR #209: `e3f8de2` added a `paperswithcode`
extra with its dependency and no bundle entry, and the omission needed a
separate follow-up commit (`c625b7a`) to fix. Nothing in CI failed in between —
which is the whole point. `all` is complete in `main` today, so this test is
green on merge; it is a guard, not a fix.

## Change

One test that parses `pyproject.toml` with `tomllib` and asserts every
declared extra is reachable from `all`. Self-references are followed
transitively, so an intermediate bundle extra still counts, and names are
normalized the way pip and PEP 685 do, so `paperswithcode` and
`papers-with-code` are not mistaken for two different extras.

The failure message names the missing extras, prints the exact lines to add,
and reminds the author to re-run `uv lock` so the new edge is recorded:

```
the 'all' extra does not install: helm. Add "every-eval-ever[helm]" to
[project.optional-dependencies] all in pyproject.toml, then re-run `uv lock`
so uv.lock records the new edge.
```

It lives in `tests/test_packaged_resources.py` rather than a new file: both
tests there check that what we ship matches what the repository promises,
and both are cheap metadata assertions with no install step. It skips when
`pyproject.toml` is absent — matching the existing convention for tests that
read repository files — so running the suite against an installed wheel
still works.

### DIFF
--- a/tests/test_packaged_resources.py
+++ b/tests/test_packaged_resources.py
@@ -1,7 +1,16 @@
+import re
+import tomllib
 from pathlib import Path
+
+import pytest
 
 from every_eval_ever.helpers import schema as helper_schema
 from every_eval_ever.schema import get_schema_version
+
+PYPROJECT = Path(__file__).resolve().parents[1] / 'pyproject.toml'
+BUNDLE_EXTRA = 'all'
+# `every-eval-ever[inspect]`, capturing the extra names inside the brackets.
+_SELF_REFERENCE = re.compile(r'^\s*([A-Za-z0-9._-]+)\s*\[([^\]]*)\]')
 
 
 def test_helper_schema_version_is_independent_of_checkout_layout(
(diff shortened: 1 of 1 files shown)


tests/test_packaged_resources.py
```diff
@@ -17,3 +26,67 @@ def test_helper_schema_version_is_independent_of_checkout_layout(
...
+) -> set[str]:
+    """Collect the extras a bundle extra pulls in, following self-references."""
+    seen: set[str] = set()
+    pending = [bundle]
+    while pending:
+        current = pending.pop()
+        for requirement in extras.get(current, []):
+            match = _SELF_REFERENCE.match(requirement)
+            if match is None or _normalize(match.group(1)) != project_name:
+                continue  # A third-party requirement, not a self-reference.
```
mrshu, line 47: **RAV-RUN1-R1-F001 [medium] packaging-test** — Conditional self-references count as universal bundle edges

The prefix regex captures the project and extras but leaves a PEP 508 marker unexamined. A win32-only self-reference then satisfies this test on Linux even though installing `[all]` there omits that extra.

A PEP 508 parser can distinguish markers while preserving valid unmarked version and URL forms. Focused marker and direct-reference cases would cover both sides.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — Approval is disabled; one confirmed packaging edge case remains.
**Result:** The graph traversal is sound, but conditional PEP 508 references can make the completeness guard pass incorrectly.
**Scope:** This PR tests that the `all` optional extra transitively reaches every other declared extra.
**Checks:** 1 concern checked; 1 confirmed, 0 ruled out, 0 set aside, 0 lowered in priority.
**Second check:** targeted, 2 reviewers; 1 finding kept and 1 fix path clarified.

## Earlier review comments

## What I noticed

- **RAV-RUN1-R1-F001 [medium] packaging-test** `tests/test_packaged_resources.py:45-47` — The prefix regex ignores PEP 508 environment markers. A conditional self-reference can therefore count as a universal `all` edge on platforms where it installs nothing. (inline)

- **[medium] packaging-test** — PEP 508 parsing can reject self-references with markers while preserving unmarked version constraints and direct URL references. Marker, URL, version, spacing, normalization, and cycle cases would keep the guard stable. (`RAV-RUN1-R1-P001`; covers `RAV-RUN1-R1-F001`)

<details>

- Target: PR #224 (`test-all-extra-completeness`, 1 file, +73/-0)
- Run ordinal: 1
- Rounds: 1/1 completed; adaptive off; material finding remains
- Mix: 3 codex-exec; 1 completed exact-head review, 2 failed because exact-head access was unavailable
- Finding counts: 0 critical, 0 high, 1 medium, 0 low, 0 nit
- Checks: concerns=1; confirmed=1/ruled-out=0/set-aside=0/lowered=0
- Second check: targeted; reviewers=2; kept=1/clarified=1/set-aside=0/removed=0; approval changed no
- Set aside: 0 items

</details>

---

PR #225, opened 2026-08-05
Add a repo-wide PR template with a decisions/coverage slot

borgr:
## What

A repo-wide **`.github/PULL_REQUEST_TEMPLATE.md`**: a short checklist plus a **"Decisions & coverage"** section.

Split out of #208 because a repo-wide template is a maintainer-wide decision that shouldn't gate an additive skill. It stands on its own — the checklist is about the validator's gate and the decisions section is about contributor discipline; neither needs a skill to exist. (The `skill-gap` issue template that was originally here named a specific skill and asked which of its reference files was unclear, so it moved to #208 where it means something.)

## Why these two sections

**Decisions, not friction.** It asks for the non-obvious *choices* — with the alternative rejected and a confidence level — rather than for where someone struggled. A confident wrong choice produces no friction to report, and those are the ones that reached the datastore: a metric named for the eval, a leaderboard scrape marked `evaluation_run`, one published number attributed to four configs. A low confidence marker is a request for maintainer attention and tells a reviewer where to look first.

**Coverage, stated as a number.** `N source rows → M records, K dropped (reason)`, with no silent caps. An adapter following the current convention already writes `adapter_reports/<collection>_failures.json`, so the line is substantiated by an artifact rather than by trust.

The checklist leads with the two things that actually gate a datastore submission: `validate` clean **including warnings** (a "Ready to Merge" verdict can still carry hundreds of them), and every unconvertible row accounted for with a non-zero exit rather than skipped.

## Notes

- Both sections are explicitly optional for non-data PRs — the template says so.
- Independent of #208 in both directions; either can merge first.


borgr:
Closing this — on reflection it adds friction without buying enforcement.

A GitHub PR template is inert text: it runs nothing, verifies nothing, and blocks nothing (GitHub never checks the checkboxes). So every item in it is either already automated, better automated, or in the wrong repo:

- `pytest` is already run by CI on every PR.
- `ruff check` is *not* in CI — but the fix for that is a lint job, not a checkbox asking contributors to attest to it. I've queued that as a small follow-up instead.
- The two headline items — `validate` clean including warnings, and unconvertible rows accounted for in `adapter_reports/` — are about generated *records*, which are submitted to the `EEE_datastore` HF dataset, not here. The EvalEvalBot already validates them there. Putting them in this repo's PR template asks the wrong contributors, on the wrong PRs.

That leaves the "Decisions & coverage" section as the only genuinely non-automatable part, and it's already specified where the people who need it will see it: SKILL.md step 7 and `reference/verification.md` in #208.

Net: a template every contributor must read past, including for a typo fix, in exchange for nudges that are redundant or misplaced. `main` has no PR template today, so closing this simply keeps it that way.

The `skill-gap` issue template that started out here already moved into #208, where it means something.

PR #226, opened 2026-08-05
validate: accept a directory argument

borgr:
## Review lane

**Fast.** One file plus tests, no conflicts with `main`, verified locally. It changes the CLI's argument handling, not what any record validates as.

## Checklist

- [x] offline unit tests added; full `pytest tests` green (395 passed / 20 skipped)
- [x] `ruff check` clean
- [x] no record's validation outcome changes — same files in, same report out

## Problem

`python -m every_eval_ever validate data/my-source/` — the first command a
contributor tries — fails:

```
directory arguments are not supported: 'data/my-source'; pass files or a fixed-depth glob
```

The working form is `... validate 'data/my-source/*/*/*.json'`, with the depth
guessed correctly and the quoting right. That is knowledge the tool already
has.

The refusal costs more than one error message:

- the agent conversion skill merged in #208 ships a `scripts/validate.sh` whose
  main job is walking a directory into a file list, and carries the "files or a
  glob, never a directory" caveat in four places — `SKILL.md` step 6,
  `reference/verification.md`, the header comment of `scripts/validate.sh`, and a
  comment in `templates/aggregate_adapter.py`;
- `check_duplicate_entries`, the other validator CLI in this same package, has
  always accepted directories and walked them with `os.walk` — so the two
  commands a contributor runs back to back disagree about what an argument
  means.

## Change

A directory now expands recursively to the `.json` and `.jsonl` files under it,
walked the same way `check_duplicate_entries` walks one. The guard the refusal
provided — never silently validate a tree you did not mean to — moves into the
output instead of the exit code:

```
$ python -m every_eval_ever validate /tmp/out/data
/tmp/out/data: validating 6 file(s) found recursively
```

That line goes to **stderr**, so `--format json` and `--format github` stay
parseable for anything downstream.

Validation is read-only, so over-matching costs a longer report and nothing
worse. A directory containing no data files is still an error, phrased like the
existing empty-glob error, because it nearly always means a typo'd path:

```
directory contains no .json or .jsonl files: '/tmp/out/empty'
```

`expand_paths` takes a keyword-only `on_directory` callback rather than printing
directly, so it stays a pure function for its callers and tests while `main`
does the reporting.

### Dot-directories are pruned, and symlinks are not followed

Recursive expansion means `validate .` from a checkout would otherwise descend
into `.venv`, `.git` and every tool cache and report on whatever JSON each
installed package happens to ship. Nothing under those paths is ever datastore
data, so names starting with `.` *below* the argument are skipped. Only names
below the argument are filtered, so pointing at `.hidden/` itself still
validates what the caller explicitly asked for.

The walk prunes those directories as it goes rather than filtering their files
out afterwards, so a checkout-sized `.venv` is never entered, and it does not
follow directory symlinks, so a link back up the tree cannot duplicate a subtree
or make the walk loop. (`Path.rglob` would do both: it enumerates the pruned
subtree before the filter discards it, and on the oldest Python this project
supports it follows directory symlinks — `recurse_symlinks` only arrived in
3.13.)

This is a traversal filter, not a "make `validate .` clean" feature. Non-dot
artifact directories such as `build/` are still walked, and a repo root also
holds the schemas and test fixtures: in this checkout `validate .` expands to 38
files, none of them datastore records. The stderr count is what tells you that;
the intended argument is a datastore subtree.

## Tests

`tests/test_validate.py`: recursive expansion past a non-data file,
deduplication when a directory and one of its files are both passed, the
dot-directory skip (including that an explicitly-passed dot-directory is still
expanded), that a directory symlink is not followed, the empty-directory error,
and `main()` returning 0 for a directory argument with stdout still parseable as
JSON while the notice sits on stderr. The old
`test_directory_arguments_are_rejected` is replaced by these — that is the
behaviour change, stated in one place.

Also verified end to end outside the suite: six real published aggregates copied
to their canonical `data/<collection>/<developer>/<model>/` paths plus a
`.venv/` holding a JSON file. `validate data` reports on exactly the six, the
`.venv` file never appears, the notice is on stderr and stdout is a parseable
JSON array. (Those six fail validation on their own merits — published records
predate some current rules — which is separate from what this PR changes.)

## While here

The parser's `prog` said `eee-validate`, a command this package does not
install; the only console script is `every_eval_ever`. Through the subcommand it
is overridden anyway, so the wrong name was reachable only by running the module
directly — precisely when a reader needs a usage line they can copy. One line,
trivially droppable if you would rather keep it separate.

## Follow-up in `.claude/`, not claimed here

#208 has merged, so `main` now carries the obsolete caveat in the four places
listed above, and `scripts/validate.sh` exists largely to work around a
limitation this PR removes. Once this lands, that skill wants a follow-up: drop
the caveat, and either delete `validate.sh` or reduce it to the part still doing
work (validating at the canonical path). Deliberately not touched from here — a
validator change and a skill change want separate review.


every_eval_ever/validator/validate.py
```diff
@@ -98,8 +106,56 @@ def _local_path_context(path: Path) -> tuple[str, _LocalRepositoryFiles]:
...
+    ship. Nothing under those paths is ever datastore data. Only names below the
+    argument are filtered, so pointing at ``.hidden/`` itself still validates
+    what the caller asked for.
+
+    The walk prunes those directories instead of filtering them out afterwards,
+    so a checkout-sized ``.venv`` is never entered, and it does not follow
+    directory symlinks, so a link back up the tree cannot make it loop.
+    """
+    files: list[Path] = []
+    for parent, directories, names in os.walk(directory):
```
mrshu: **RAV-RUN1-R1-F001 [medium] filesystem-traversal** — `os.walk` can skip an unreadable descendant and still report successful validation

`os.walk` ignores scan errors when `onerror` is unset. If the tree also contains readable JSON files, the command validates that partial set and can return success.

An error callback can stop the scan and name the inaccessible path. A mocked scan error after one readable file would cover this path without platform-specific permission behavior.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This run is comment-only and found one focused directory traversal concern.
**Result:** The directory support is well scoped. One partial-scan case needs attention, and one help-text update would keep the CLI consistent.
**Scope:** This PR lets the validate CLI scan directories recursively while pruning hidden descendants and directory symlinks.
**Checks:** skipped because three reviewers independently found the same material concern.
**Second check:** 2 reviewers checked 2 findings and 2 plans; 4 kept and 2 fix paths clarified.

## Earlier review comments

None.

## What I noticed

|---|---|---|---|---|
| RAV-RUN1-R1-F001 | medium | filesystem traversal | `every_eval_ever/validator/validate.py:125` | `os.walk` ignores scan errors when `onerror` is unset. A readable file can still validate after an unreadable subtree is skipped. (inline) |

<details>

- **RAV-RUN1-R1-F002 [low] CLI help** — The main `every_eval_ever validate` help still says directory arguments are unsupported. This is live help from `every_eval_ever/cli.py`, so it now differs from the new validator behavior.

</details>

## Things to try

- **[medium] filesystem traversal** — An `os.walk` error callback can stop partial scans and name the inaccessible path. A mocked scan error can test this without platform-specific permission behavior. (`RAV-RUN1-R1-P001`; covers `RAV-RUN1-R1-F001`)
- **[low] CLI help** — The top-level parser can describe recursive directory support. A focused help assertion can also reject the old unsupported-directory text. (`RAV-RUN1-R1-P002`; covers `RAV-RUN1-R1-F002`)

<details>

- Target: PR #226 (`validate-accepts-directory`, 2 files, +172/-15)
- Run ordinal: 1
- Mix: 3 codex-exec
- Finding counts: 0 critical, 0 high, 1 medium, 1 low, 0 nit
- Checks: skipped; the material concern had three-reviewer consensus
- Second check: targeted; reviewers=2; kept=4; clarified=2; set-aside=0; removed=0; approval changed=no
- Set aside: 0 items

</details>

---

nelaturuharsha:
- I think validate.sh should adapt to the proposed structure, we enforce it because otherwise people will not pre-format the datastore as we keep data right now.
- check_duplicate_entries is at the moment in abeyance, so it is not something to worry about. In view of these, I think we would not merge this PR at the moment.

every_eval_ever/validator/validate.py
```diff
@@ -113,13 +173,15 @@ def expand_paths(paths: list[str]) -> list[Path]:
         for match in matches:
             path = Path(match)
             if path.is_dir():
```
mrshu, line 175: **RAV-RUN2-R1-F001 [medium] filesystem-traversal** — An explicit directory symlink is followed into its target

`Path.is_dir()` follows the symlink, and `os.walk()` then scans that path as its root. `followlinks=False` blocks directory symlinks discovered below the root, but it does not block a symlink supplied as the root. The current test covers only the discovered-child case.

The expansion can reject directory-target symlinks for direct and glob-expanded matches while keeping file-symlink behavior unchanged. Tests for both directory-symlink inputs and an accepted file symlink would cover the boundary.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — Directory traversal errors are now handled, and one focused symlink edge remains.
**Result:** One medium-priority suggestion remains. An earlier low-priority help-text suggestion is also still present.
**Scope:** Accept validator directory arguments while pruning dot-directories, avoiding symlink traversal, preserving structured output, and failing incomplete scans.
**Checks:** 1 concern checked and confirmed.
**Second check:** 2 reviewers kept the symlink finding, clarified its fix path, and removed one low-priority output concern.

## Earlier review comments

- `RAV-RUN1-R1-F001` is fixed. `os.walk` now raises traversal errors through its `onerror` callback, and a focused test covers this path.
- `RAV-RUN1-R1-F002` was reported earlier and is still present. The installed top-level help still says directory arguments are unsupported.

## What I noticed

|---|---|---|---|---|
| RAV-RUN2-R1-F001 | medium | filesystem traversal | `every_eval_ever/validator/validate.py:175` | An explicitly supplied directory symlink is followed into its target. `followlinks=False` blocks discovered child symlinks, but it does not block a symlink used as the walk root. |

<details>

- **RAV-RUN1-R1-F002 [low] CLI help** — The top-level `every_eval_ever validate` help still says directory arguments are unsupported. The existing help text can describe recursive directory expansion.

</details>

## Things to try

- **[medium] filesystem traversal** — The path expansion can reject only symlinks whose targets are directories. This check can apply to direct and glob-expanded matches while keeping file-symlink behavior unchanged. Focused tests can cover direct and globbed directory symlinks plus an accepted file symlink. (`RAV-RUN2-R1-P001`; covers `RAV-RUN2-R1-F001`)

<details>

- Target: PR #226 at `1e405097d79fbe2a960264f2814c534a1ddb35a9` (2 files, +192/-16)
- Run ordinal: 2
- Mix: 3 codex-exec
- Earlier review comments: 5 ledger entries grouped into 2 topics; 1 fixed and 1 reported item still present
- Finding counts: 0 critical, 0 high, 1 medium, 1 low, 0 nit
- Checks: concerns=1; confirmed=1; set-aside=0
- Second check: targeted; reviewers=2; kept=1; clarified=1 plan; removed=1 low concern; approval unchanged
- Set aside: 0 items

</details>

---

borgr:
Closing per @nelaturuharsha's call — agreed, and the reason it stopped mattering is worth recording.

#238 landed `scripts/local_validate.py`, which walks a folder and validates what it finds. So the capability exists on `main` **without** relaxing the CLI's contract, which was the point of the objection: `every_eval_ever validate` still rejects a bare directory, and contributors still get told to pre-format the datastore.

Four places in the conversion skill tell contributors "files or a glob, not a directory" — `SKILL.md:82`, `reference/verification.md:5`, `scripts/validate.sh:3`, and a comment at `templates/aggregate_adapter.py:184`. All four are still correct, and now there is somewhere to send a reader who wanted the other behaviour. Pointing them at `scripts/local_validate.py` is a docs change on `main`, not this PR.

PR #227, opened 2026-08-05
docs: state which root each output-dir parameter expects

borgr:
## Review lane

**Fast.** Two files plus tests, no conflicts with `main`, verified locally.

One caveat for the lane call: it makes the publisher **raise** on a `base_output_dir` it used to accept, so a caller who was publishing one level too deep now fails instead of writing. That is the bug, but it is a behaviour change and worth a look rather than a rubber stamp.

## Checklist

- [x] offline unit tests added, from both sides — a base inside `data` is rejected whether or not its name matches the collection, and a scratch root that merely shares the collection's name still publishes; full `pytest tests` green (411 passed / 20 skipped) against current `main`
- [x] `ruff check` clean

## Problem

Three parameters name an output directory, and they use two different
conventions:

| call | expects |
| --- | --- |
| `publish_evaluation_logs(base_output_dir=…)` | `data` |
| `EvaluationLogOutput(base_dir=…)` → `save_evaluation_logs` | `data/<collection>` |
| `default_failure_report_path(…)` | any path below `data` |

Nothing in the signatures says which, and getting it wrong is quiet:

- `publish_evaluation_logs` given anything one level inside `data` writes
  `data/<x>/<collection>/<developer>/<model>/` and fails much later, at the
  path-depth check, pointing at the file rather than the caller;
- `default_failure_report_path` given the wrong depth silently *renames* the
  report — `data/c/dev` yields `adapter_reports/c__dev_failures.json` instead
  of `c_failures.json`;
- `EvaluationLogOutput.base_dir` reads the collection name from
  `base_dir.name`, so a wrong depth silently reassigns the collection.

This cost real debugging time while writing the conversion-skill templates.

## Change

Docstrings, plus one guard.

Each of the three parameters now states the expected root, names the failure
mode for the wrong one, and cross-references the parameter that uses the
opposite convention. `default_failure_report_path` additionally documents what
was undocumented rather than merely unstated: the root is the **nearest** `data`
ancestor, the report name is every component below it joined by `__`, and in a
nested tree (`<datastore>/data/staging/data/<collection>`) the inner root wins,
which puts `adapter_reports/` under the outer datastore — the one placement the
function cannot keep out of a validated tree. Every documented path was checked
against the function.

The guard: `publish_evaluation_logs` now refuses a `base_output_dir` whose
parent is named `data`. The name of that directory is beside the point — one
level inside `data` publishes one level too deep whether it matches the
collection being published, some other collection, or nothing. A scratch root
outside `data` still publishes, including one that shares a collection's name.

**Signatures are deliberately unchanged.** `save_evaluation_logs` appears 55
times across 26 Python files and `default_failure_report_path` 57 times across
28 (counted under `every_eval_ever/` and `tests/` on current `main`); renaming
these or adding `data_root=` / `collection_dir=` aliases would touch every one of
them and conflict with anything in flight. If the aliases are wanted later, they
are additive and this documentation stays correct either way.

## One thing the investigation turned up

Passing `data` itself as `EvaluationLogOutput.base_dir` is **already**
caught — `_required_path_component` rejects the reserved name, so publication
raises `collection cannot use the reserved datastore name "data"`. That is also
the *only* check on that parameter: the final component must be a usable path
component, and every level above it is trusted. The docstring now records both
halves, so the next reader does not have to re-derive which mistakes are loud
and which are not.

The depth mistake itself is already caught at validate time by
`check_path_structure`, which has always been an error — it just reports the
file's path rather than the parameter that produced it, which is what these
docstrings and the guard are for. (An earlier version of this description
credited #223 with that; #223 returns nothing when the depth is wrong, precisely
so the path-structure check owns it.)


every_eval_ever/helpers/io.py
```diff
@@ -389,8 +404,14 @@ def default_failure_report_path(
 
     A report is not an ``EvaluationLog`` and must never be placed under
     ``data/<collection>/...`` where a PR validator could mistake it for one.
-    For a normal ``data/<collection>`` output this returns
-    ``adapter_reports/<collection>_failures.json``.
+
+    ``output_dir`` is the **collection** directory, ``data/<collection>`` —
```
mrshu: **RAV-RUN1-R1-F001 [medium] api-contract** — `output_dir` also supports report-identity paths below `data`

The CLI passes paths such as `data/lm_eval_samples`, and the tests require `adapter_reports/lm_eval_samples_failures.json`. Calling the argument only a collection directory narrows a supported contract.

The wording can describe a path below the nearest `data` root. Its relative components form the report name, while `data/<collection>` remains the usual adapter case.

every_eval_ever/helpers/io.py
```diff
@@ -389,8 +404,14 @@ def default_failure_report_path(
...
+    the same convention as ``EvaluationLogOutput.base_dir``, and one level
+    deeper than ``publish_evaluation_logs(base_output_dir=...)``. Given
+    ``data/<collection>`` this returns
+    ``adapter_reports/<collection>_failures.json``; the report name is built
+    from the path components below ``data``, so a deeper or shallower
```
mrshu: **RAV-RUN1-R1-F002 [medium] report-placement** — A nested `data` component can keep the report inside an outer datastore tree

The helper selects the nearest ancestor named `data`. For `data/outer/data/collection`, it writes below `data/outer/adapter_reports`, which conflicts with the stated outside-tree guarantee.

A note that the nearest `data` ancestor wins, plus this repeated-`data` example, would make the placement limit clear.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This run is comment-only and found two documentation contracts that need tighter wording.
**Result:** The new guidance is useful. Two edge cases need clearer contracts, and one sentence can state its existing validation limit.
**Scope:** This PR documents the directory roots used by publication and failure-report APIs.
**Checks:** 2 concerns checked; 2 confirmed.
**Second check:** 2 reviewers checked 3 findings and their plans; all 3 findings stayed, and 2 fix paths were clarified.

## Earlier review comments

None.

## What I noticed

|---|---|---|---|---|
| RAV-RUN1-R1-F001 | medium | API contract | `every_eval_ever/helpers/io.py:408` | The docstring calls `output_dir` a collection directory, but supported callers also pass report-identity paths such as `data/lm_eval_samples`. (inline) |
| RAV-RUN1-R1-F002 | medium | report placement | `every_eval_ever/helpers/io.py:413` | The nearest ancestor named `data` controls placement. A nested `data` component can leave `adapter_reports` inside an outer datastore tree. (inline) |

<details>

- **RAV-RUN1-R1-F003 [low] documentation** — The `EvaluationLogOutput` docstring says any wrong depth except exact `data` is accepted. The leaf still must pass collection-name validation.

</details>

## Things to try

- **[medium] API contract** — The parameter can be described as a path below the nearest `data` root. Its relative components form the report name; `data/<collection>` is the usual adapter case. (`RAV-RUN1-R1-P001`; covers `RAV-RUN1-R1-F001`)
- **[medium] report placement** — The docs can state that the nearest `data` ancestor wins. A repeated-`data` example makes the placement limit concrete. (`RAV-RUN1-R1-P002`; covers `RAV-RUN1-R1-F002`)
- **[low] documentation** — The depth statement can say that parent depth is unchecked when the final component is a valid collection name. (`RAV-RUN1-R1-P003`; covers `RAV-RUN1-R1-F003`)

<details>

- Target: PR #227 (`docs-output-dir-conventions`, 2 files, +35/-3)
- Run ordinal: 1
- Mix: 3 codex-exec
- Finding counts: 0 critical, 0 high, 2 medium, 1 low, 0 nit
- Checks: concerns=2; confirmed=2; ruled-out=0; set-aside=0; lowered=0
- Second check: targeted; reviewers=2; kept=3; clarified=2; set-aside=0; removed=0; approval changed=no
- Set aside: 0 items

</details>

---

every_eval_ever/converters/common/publication.py
```diff
@@ -48,6 +48,25 @@ def _output_dir(
...
+    """Refuse a *base_dir* that is already the collection directory.
+
+    ``base_dir/<collection>`` repeating the collection inside a ``data`` tree is
+    the caller having passed ``data/<collection>``, the convention
+    ``EvaluationLogOutput`` uses. A scratch root that happens to share the
+    collection's name is not under ``data``, so it still publishes.
+    """
+    if (
+        base_dir.parent.name == 'data'
+        and output_dir.parent.parent.name == base_dir.name
```
mrshu: **RAV-RUN2-R1-F001 [medium] publisher-validation** — The new guard accepts a collection directory when its name differs from the selected collection.

For `base_output_dir=data/foo` and a selected collection of `bar`, the name check is false. Publication then writes `data/foo/bar/<developer>/<model>`, which is still one level too deep.

A small follow-up could reject every publisher base that is an immediate child of canonical `data`. Mismatched source and `collection_override` tests would cover this path.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This run is comment-only; one new medium suggestion and three earlier notes remain.
**Result:** The new path guard is a helpful step. One collection-directory case can still pass when its name differs from the selected collection.
**Scope:** Clarify output-directory contracts and reject collection directories passed as the publisher root.
**Checks:** Skipped because the new finding had three-reviewer consensus. All earlier items were revalidated.
**Second check:** Targeted, 2 reviewers; 4 findings kept and 1 fix path clarified.

## Earlier review comments

- **RAV-RUN1-R1-F001 [medium] API contract** — Still present. Active callers pass report-identity paths below `data`, although the docstring calls every input a collection directory.
- **RAV-RUN1-R1-F002 [medium] report placement** — Still present. The nearest `data` ancestor controls placement, so nested `data` paths can keep `adapter_reports` inside an outer datastore tree.
- **RAV-RUN1-R1-F003 [low] documentation** — Still present. Parent depth is unchecked only when the collection leaf also passes validation.

## What I noticed

|---|---|---|---|---|
| RAV-RUN2-R1-F001 | medium | publisher validation | `every_eval_ever/converters/common/publication.py:61` | The new guard accepts `data/foo` when the selected collection is `bar`, then publishes one level too deep. |

## Things to try

- **[medium] publisher validation** — A small follow-up could reject every `base_output_dir` that is an immediate child of canonical `data`. Mismatched source and `collection_override` tests would cover the open path. (`RAV-RUN2-R1-P001`; covers `RAV-RUN2-R1-F001`)
- **[medium] documentation** — The failure-report docstring could describe paths below the nearest `data` root and how relative components form the report name.
- **[medium] documentation** — The placement text could name the nearest-ancestor rule and its nested-`data` limit.
- **[low] documentation** — The depth note could say that the final collection component must still be valid.

<details>

- Target: PR #227 at `854e27fd6bbf357595ca5f2855b7f232d84911ab` (3 files, +71/-3)
- Run ordinal: 2
- Mix: 3 codex-exec
- Earlier review comments: 3 root findings; 3 still present
- Finding counts: 0 critical, 0 high, 3 medium, 1 low
- Checks: skipped; the new concern had three-reviewer consensus
- Second check: targeted; reviewers=2; kept=4; clarified=1; set-aside=0; removed=0

</details>

---

borgr:
All four are addressed in `3bbb2e683`. 411 passed / 20 skipped against current `main`, `ruff` clean.

**RAV-RUN2-R1-F001 / P001 — `data/foo` accepted while publishing collection `bar`.** Confirmed, and the name comparison was the wrong test to begin with. Every path this publisher writes is `data/<collection>/<dev>/<model>/`, so a base one level inside `data` is one level too deep whatever it is named — matching the collection was incidental to the bug, not the cause of it. The guard is now just `base_dir.parent.name == 'data'`, renamed `_reject_base_below_data`, and the message says "inside a datastore `data/` directory" rather than "is a collection directory". A scratch root outside `data` still publishes, including one that happens to share a collection's name.

Tests cover the mismatched-name case and the `collection_override` case alongside the original matching one, plus the scratch root that must keep working.

**RAV-RUN1-R1-F001 — the docstring called every input a collection directory.** Fixed, and you were right that real callers are deeper: `hal` passes `out_root / benchmark.output_name`. `default_failure_report_path` now documents that `output_dir` is *any* path below a `data` root, that the report is named from every component below that root joined by `__`, and gives both shapes — `data/<collection>` → `<collection>_failures.json`, `data/hal/<benchmark>` → `hal__<benchmark>_failures.json`. All four documented outputs were checked against the function rather than read off the code.

**RAV-RUN1-R1-F002 — nearest-`data` placement.** Documented as the nearest ancestor, with the nested case spelled out: `<datastore>/data/staging/data/<collection>` resolves to the inner root and so puts `adapter_reports/` under the outer datastore. That is the one placement this function cannot keep out of a validated tree, and it is now stated as a limit rather than left to be discovered. The no-`data`-ancestor branch is documented too, since nothing else described it.

**RAV-RUN1-R1-F003 — the depth note.** `EvaluationLogOutput.base_dir` now says only the final component is checked, and only for being a usable path component (`_required_path_component`); any depth above it is trusted, so a wrong one renames the collection rather than failing.

This PR remains documentation plus that one guard — no record's validation outcome changes.

mrshu:
# ⚒️ review-anvil report

**Review decision:** COMMENT — This pass is comment-only; two medium path-contract suggestions remain.
**Result:** The immediate-child guard and main docstrings are improved. Two narrower cases can complete the contract.
**Scope:** State and enforce which root each publication and failure-report API expects.
**Checks:** skipped; both findings had three-reviewer consensus.
**Second check:** 1 reviewer kept both findings and clarified both fix paths.

## Earlier review comments

- **Fixed:** nearest-`data` placement, leaf validation, and mismatched immediate collection directories are now covered.
- **Still present:** `RAV-RUN2-R1-F001` remains for deeper descendants, and `RAV-RUN1-R1-F001` remains in one publisher cross-reference.

## What I noticed

|---|---|---|---|---|
| RAV-RUN2-R1-F001 | medium | publisher validation | `every_eval_ever/converters/common/publication.py:61` | The guard accepts deeper descendants such as `data/staging/foo`, then publishes another collection subtree below them. |
| RAV-RUN1-R1-F001 | medium | API contract | `every_eval_ever/converters/common/publication.py:178` | The publisher doc still describes the failure-report helper as accepting only `data/<collection>`, although active callers use deeper report identities. |

## Things to try

- **[medium] publisher validation** — Normalized path components can reject an exact `data` ancestor while preserving the root itself and unrelated names such as `metadata`. (`RAV-RUN3-R1-P001`)
- **[medium] API contract** — The docs can state all three contracts separately: data root, collection directory, and report identity below the nearest data root. (`RAV-RUN3-R1-P002`)

<details>

- Target: PR #227 (`docs-output-dir-conventions`, 3 files, +93/-3) at `3bbb2e683e3c04bfcac43e548753d180d5d23969`
- Run ordinal: 3
- Mix: 3 codex-exec
- Checks: skipped; consensus findings only
- Second check: challenge; reviewers=1; kept=2, clarified=2, removed=0

</details>

---

borgr:
Reviewed the 3 unresolved automated review threads against the current head. All three are stale against the reworked `helpers/io.py` and `converters/common/publication.py`; the current code already covers them. Resolving them.

Issue #202, opened 2026-07-18
pip install breaks every_eval_ever.helpers: schema path resolves outside the package


gkastanis:
`helpers/schema.py` loads the schema version at import time:

    schema_path = Path(__file__).parent.parent.parent / 'eval.schema.json'

Three parents up is the repo root in a git checkout, so this works when you
run from a clone. In a wheel install it is site-packages, and the file is
not there:

    FileNotFoundError: [Errno 2] No such file or directory: '.../site-packages/eval.schema.json'

So `import every_eval_ever.helpers` crashes for every pip user. Reproduced
with every-eval-ever==0.2.3rc1 from PyPI on Python 3.12 / Linux.

The wheel does ship the file, at `every_eval_ever/schemas/eval.schema.json`,
so the fix is to point there:

    schema_path = Path(__file__).parent.parent / 'schemas' / 'eval.schema.json'

or `importlib.resources.files('every_eval_ever.schemas')` if you want it to
survive zip installs too. Same path is on main today.

I hit this while wiring the Drupal ai_best_practices eval suite to emit EEE
0.2.2 through the `[inspect]` converter. The converter and validator are not
affected, they don't import helpers.

PR #230, opened 2026-08-05
Warn when two names for one publisher split its datastore directory

borgr:
## Review lane

**Needs a human.** It adds a registered validator rule, so the same input now produces a report it did not before, and it imports its vocabulary from #190, which puts it behind #190 in the queue.

Design agreed in [this PR's thread](https://github.com/evaleval/every_eval_ever/pull/230#issuecomment-5216074197): @nelaturuharsha asked for the registry work to own this vocabulary rather than a second copy living here, which is what the current revision does.

## Checklist

- [x] `validate` runs clean at the final `data/<collection>/<dev>/<model>/` path; the finding is a **warning**, so the record stays valid — but since #238 a warnings-only run exits **2**, not 0. Verified end to end, see *Blast radius*
- [x] offline unit tests added; `pytest tests` 422 passed / 20 skipped with #190's two files present, `ruff check` clean — CI here is red until #190 lands, see *Merge order*

## What it does

The datastore keeps one directory per publisher, taken from `model_info.id`'s namespace prefix or, for a flat id, from `model_info.developer` (`helpers/io.py:datastore_path_components`). Two names for one publisher are two directories, and neither listing is complete. This warns when a record's publisher name is one the **eval-card-registry** knows as a second name for an organization.

The message states what both spellings agree on — that they are one organization, and that filing under both leaves two incomplete listings — and **names both without choosing one**. The registry's canonical id is an *entity* id, and in this datastore it is often the rarer spelling (`mistral` 128 records against `mistralai` 232, but `moonshot` 22 against `moonshotai` 112 and `zai` 105 against 51 spread over six others), so which name wins is editorial. It points at the **datastore**, not the collection: one name per publisher is one decision for the whole datastore, and a per-file warning can only report that two names are in play. The check therefore takes no `repo_path` — it reads `model_info` and nothing else.

Both `model_info.id`'s prefix and `model_info.developer` are read, and one warning names every field holding a name for that publisher plus every spelling in play (`model_info.id and model_info.developer: 'Moonshot'/'moonshot' and 'moonshotai' are …`). One publisher needs one rename, so it gets one message, keyed on the organization rather than the spelling.

Only the field that decides the directory is told it splits the directory. **22 of the 361** warnings are the other case — the id prefix picks the directory and `model_info.developer` disagrees with it — and telling those they publish into two directories would be false. They get the claim one record does prove: this field disagrees with the directory the prefix picked, and groups the record apart from any record spelled the other way.

## Blast radius

Over every aggregate in the published datastore — **13,185 records** — the check emits **361 warnings on 361 records** (2.7%) across **14 organizations**, one warning per record. Names as the records themselves carry them:

| organization | warnings | names its records carry | published dirs |
|---|---|---|---|
| `mistralai` | 136 | `mistralai` 232, `mistral` 128, `Mistral` 8 | 2 |
| `zai` | 44 | `zai` 105, `zhipu` 19, `zhipu-ai` 13, `z-ai` 7, `THUDM` 6, `zhipu_ai` 5, `thudm` 1 | 8 |
| `allenai` | 43 | `allenai` 198, `ai2` 43 | 2 |
| `moonshotai` | 29 | `moonshotai` 112, `moonshot` 22, `kimi` 6, `moonshot-ai` 4 | 4 |
| `cohere` | 28 | `cohere` 76, `CohereLabs` 28 | 1 |
| `alibaba` | 21 | `alibaba` 375, `Alibaba` 22, `Alibaba-NLP` 1 | 2 |
| `amazon` | 14 | `amazon` 34, `aws` 14 | 2 |
| `huggingface` | 14 | `huggingface` 42, `HuggingFaceTB` 14 | 2 |
| `black-forest-labs` | 11 | `bfl` 11 | 1 |
| `minimax` | 9 | `minimax` 99, `MiniMaxAI` 9 | 1 |
| `xiaomi` | 5 | `xiaomi` 17, `XiaomiMiMo` 5 | 1 |
| `microsoft` | 4 | `microsoft` 81, `Microsoft` 14, `microsoft-ai` 3 | 3 |
| `upstage` | 2 | `upstage` 13, `upstage_ai` 2 | 2 |
| `tiiuae` | 1 | `tiiuae` 60, `tii-uae` 5, `tii` 1 | 3 |

For **10 of the 14 the second directory already exists** — these are splits in progress, not hypotheticals. `cohere`, `minimax` and `xiaomi` carry the second name while still publishing into one directory, and `black-forest-labs` has one name (`bfl`) that the registry disagrees with; for those four the warning is preventive.

**The exit code moves.** Since #238 the CLI prints warnings in the default `rich` renderer and exits **2** on a warnings-only run, so any build treating non-zero as a rejection will reject these records:

- **115 of the 361 are otherwise clean today**, so this warning flips them 0 → 2. They are also the only warnings the corpus produces on a passing record: of 13,185 records, 1,934 pass clean and 115 pass with this warning.
- The other **246 already fail** on `main` for reasons this PR does not touch, so they exit 1 either way.

Verified on a record filed under `mistral/`:

```
│  WARN   Aggregate (EvaluationLog)
│   Warning: model_info.id: 'mistral' and 'mistralai' are one organization in
│ the eval-card-registry — filing under both puts one publisher in two
│ datastore directories, neither listing complete. Match how it is already
│ filed datastore-wide
│ 0 passed, 1 warning-only, 0 failed (0 errors, 1 warnings)
│ Valid locally, but not merge-ready. Fix all warnings.
EXIT=2
```

Six adapters currently produce one of these names, two of them mapping the registry's canonical name *to* the second one:

- `adapters/hle/adapter.py:105` — `'zai': 'zhipu-ai'`
- `adapters/mercor_eval/adapter.py:64-65` — `'moonshot-ai': 'moonshot'`, `'moonshotai': 'moonshot'`
- `adapters/bfcl/adapter.py:34-35` — `'Zhipu AI': 'zhipu'`, `'Zhipu': 'zhipu'`
- `adapters/terminal_bench_2/adapter.py:65-66` — `'Z-AI': 'zhipu-ai'`, `'Z.ai': 'zhipu-ai'`
- `adapters/mt_bench/adapter.py:93` — `'chatglm-6b': 'thudm'`
- `adapters/sciarena/adapter.py:43` — `"GLM-4.5": "zhipu"`

## Merge order: #190 first

The check calls `second_name_of` from `helpers/eval_card_registry.py`, the vendored snapshot added by **#190**. This branch adds no registry code and no data file of its own. A snapshot rather than a live call, because `POST /api/v1/resolve` **creates a `draft` canonical for anything it cannot place** — a validator that resolved would write to a shared registry as a side effect of reading a file.

CI here is red until #190 lands: the import does not resolve on `main`, and `validation_core` is imported nearly everywhere, so the suite fails to collect rather than failing a test. The base cannot be retargeted (#190 is on another fork), so treat the red as expected. Locally the branch is green on `main` + #190.

## What it does not flag

**A HuggingFace namespace the registry has recorded is an identity, not drift** — `meta-llama`, `qwen`, `deepseek-ai`, `zai-org` all pass. This is why the check uses the registry and not `helpers.get_developer`: `get_developer('Llama-3-8B')` is `'meta'`, so a check on that helper would flag every record correctly filed under `meta-llama`.

That field is filled in for **11 of the registry's 1,166 organizations**, and the check does not paper over the gap: a namespace the registry has *not* recorded but does carry as a confirmed alias is reported like any other second name. `CohereLabs`, `MiniMaxAI` and `XiaomiMiMo` are those — 42 of the 361 warnings, and the least useful ones. They have the cleanest fix: none exists as a published directory, so the spelling the message names *is* the directory that is there, and recording the namespace upstream silences them for every consumer of the registry rather than only here.

**Case and punctuation variants are not flagged** — `Anthropic`, `snowflake`, `moonshot-ai`, `tii-uae`, `z-ai` all pass. The registry aims for HuggingFace-true casing and HuggingFace is not consistent (`anthropic`, but `Snowflake`, `LLM360`, `inclusionAI`), so its preferred spelling is no evidence about which directory *this* datastore uses. Deliberate scope cut; see *Follow-ups*.

Names the registry has never seen stay silent (`facebook`, `mosaicml`, `Xwin-LM`, `aws-prototyping`). Widening coverage means adding an alias upstream, which serves the registry's other consumers too, rather than growing a heuristic here.

## Tests

`tests/test_validation_scope.py` covers which field claims the split, one warning per publisher naming every field and spelling, two publishers reported separately, the datastore-wide wording, the four kinds of name that resemble a second name and are left alone, malformed input, and one end-to-end run asserting `valid is True` with the warning present. What `second_name_of` answers is pinned by #190's `tests/test_eval_card_registry.py` and not re-asserted here.

`datastore-gate.md` gains a `§developer` section, and `test_every_registered_check_is_documented_in_the_gate_guide` asserts every `REGISTERED_CHECKS` entry appears in that guide, so the next check cannot be added undocumented.

## Follow-ups

- **It reports the split; it does not prevent it.** `datastore_path_components` derives the directory from the literal spelling in `model_info`. A lookup there would stop new splits at the writer instead of correcting adapters one at a time — but that changes the datastore layout, so it needs its own PR and migration.
- **The case/punctuation split is the larger share and stays out.** Resolving the **921** published publisher directory names through the registry and normalizing what is left, **39 groups hold more than one directory: 10 differ by name** (this check's reach) and **29 differ only by case or punctuation** — `Anthropic`/`anthropic`, `Google`/`google`, `OpenAI`/`openai`, `Qwen`/`qwen`, `nousresearch` three ways, `PrimeIntellect`/`prime-intellect`/`prime_intellect`. Finding those needs no registry, only a tree-level report grouping sibling directory names by normalized spelling. Its reviewer question ("which spelling wins?") is a different one from this PR's ("are these two names the same publisher?").
- **Nothing keeps the vocabulary fresh.** #190's `refresh_eval_card_registry.py --check` reports staleness but no workflow runs it. That belongs with the snapshot, in or after #190.
- **Packaging exposure, not introduced here.** The snapshot is package data, so #190's `helpers/data/*.json` entry in `[tool.setuptools.package-data]` is load-bearing — dropped, the loader would raise inside a warning check and become a `SemanticCheckError` on every file. `schemas/*.json` has carried the same exposure since the validator existed, and CI installs editable, so no test catches either. One test that every non-`.py` file under the package matches a package-data glob would cover both, but it fails today on `converters/README.md` — a packaging PR, not this one.

## Note for whoever merges

#222 and #223 append to the same `REGISTERED_CHECKS` tuple, the same import list in `validate.py`, and the end of `tests/test_validation_scope.py`. Whichever lands second conflicts on adjacent lines; keep all entries. All three are `warning`-severity — the first in `REGISTERED_CHECKS`, since every pre-existing entry is `error` — so under #238's contract each moves some passing records from exit 0 to exit 2. That is the cross-PR consequence to weigh, not the validity of any record.


borgr:
Coordination note — this PR's org vocabulary is being consolidated in [#190](https://github.com/evaleval/every_eval_ever/pull/190), so the two need an ordering.

**What happened.** Three copies of the same registry vocabulary were in flight: this branch's `helpers/org_registry.py` + `helpers/data/org_registry.json`, and the AlpacaEval converter's own resolver + snapshot on #190. Both derive from the same read-only endpoints of the same registry. #190 now collapses them into one module, one snapshot and one refresh tool:

```
converters/alpaca_eval/registry.py                  -> helpers/eval_card_registry.py
converters/alpaca_eval/data/registry_snapshot.json  -> helpers/data/eval_card_registry.json
converters/alpaca_eval/refresh_registry_snapshot.py -> tools/refresh_eval_card_registry.py
```

**This PR's rules won, not the converter's.** The shared builder adopts the stricter derivation from this branch wholesale — confirmed aliases only, the target must be a listed canonical id, an identity wins (including another organization's, which is what keeps `AI21 Labs`→`ai21` from claiming a spelling `ai21-labs` already answers to), one organization per normalized spelling with ambiguity dropped — plus `second_name_of` as the read-side helper, with the same semantics this PR's validator call expects: a canonical id and a recorded `hf_org` both answer `None`, only a genuinely different name answers.

Adopting them costs nothing measurable. Run against the same live payload, the strict and lenient rules select **exactly the same 99 aliases** — 0 dropped, 0 added, 0 retargeted. Identity derivation is additionally made order-independent, which pins 4 collision keys (`DeepAuto-AI`/`deepautoai` plus three truncated `Gemini-…` ids); none is an AlpacaEval publisher. `second_name_of` was smoke-checked against the regenerated snapshot and agrees with this PR's expectations: `mistral`→`mistralai`, `AI2`→`allenai`, `glm`→`zai`, `THUDM`→`zai`, `MiniMaxAI`→`minimax`, `CohereLabs`→`cohere`; `None` for `mistralai`, `meta-llama`, `mosaicml`, `WizardLMTeam`, `wizardlm`.

**Ordering, and why it can't be the other way round.** `helpers/eval_card_registry.py` does not exist on `main`, and GitHub will not let this PR's base be a fork branch, so rebasing onto it now would just make this PR red. So: #190 merges first, then this PR rebases and

- deletes `helpers/org_registry.py`, `helpers/data/org_registry.json`, `tools/refresh_org_registry.py`, `tests/test_org_registry.py`
- changes `validator/validation_core.py:16` to `from every_eval_ever.helpers.eval_card_registry import second_name_of` (the call site at line 672 is unchanged)

The builder-level tests from `tests/test_org_registry.py` are already carried over in #190's `tests/test_eval_card_registry.py` (11 tests, including the two live-path ones), so the deletion is not a loss of coverage. Nothing merges until the set — this PR, #190, the datastore PR and the registry seed PR — is reviewed together.

every_eval_ever/validator/validation_core.py
```diff
@@ -713,6 +832,12 @@ def _aggregate_check_model_deployment(
         'error',
```
mrshu, line 806: **RAV-RUN1-R1-F003 [medium] warning-visibility** — The new warning stays hidden in the default rich output without PR #221

Warning severity leaves the report valid. The current rich renderer prints `PASS` and returns before its warning loop.

PR #221 is already listed as a prerequisite. After rebase, one default-rich CLI check for this warning would confirm the complete path.

every_eval_ever/validator/validation_core.py
```diff
@@ -713,6 +832,12 @@ def _aggregate_check_model_deployment(
...
         _aggregate_check_model_deployment,
     ),
+    ValidationCheck(
```
mrshu, line 809: **RAV-RUN1-R1-F004 [medium] contributor-guidance** — The new registered warning is missing from the required datastore gate guide

The repository instructions require contributor guidance with validator changes and name `.claude/skills/eee-dataset-conversion/reference/datastore-gate.md` as the rule location. This PR adds the check but does not update that guide.

A short section can state the alias trigger, warning-only result, and single-record evidence limit.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This run is comment-only and found two focused merge-readiness items.
**Result:** The offline registry design is thoughtful. The warning needs visible default output and its required contributor guidance before merge.
**Scope:** This PR adds an offline organization vocabulary, a refresh tool, and a warning-only publisher-alias validator.
**Checks:** 3 concerns checked; 1 confirmed and 2 narrowed.
**Second check:** 2 reviewers checked 6 findings and plans; 2 lowered in priority and 4 kept with clearer fix paths.

## Earlier review comments

None.

## What I noticed

|---|---|---|---|---|
| RAV-RUN1-R1-F003 | medium | warning visibility | `every_eval_ever/validator/validation_core.py:832` | The new warning stays hidden in the default rich output until PR #221 is included. The current renderer returns after `PASS` for valid reports. (inline) |
| RAV-RUN1-R1-F004 | medium | contributor guidance | `every_eval_ever/validator/validation_core.py:835` | This adds a registered validator rule, but the required `datastore-gate.md` contributor guidance does not describe it. (inline) |

<details>
<summary>Non-blocking low-priority findings (4 items)</summary>

- **RAV-RUN1-R1-F001 [low] refresh safety** — The maintainer refresh tool replaces the tracked snapshot with a direct truncating write. An interrupted refresh can leave invalid JSON for later semantic validation.
- **RAV-RUN1-R1-F002 [low] diagnostic scope** — The check proves a registry alias mismatch from one record. It cannot prove that the collection currently uses two directory spellings.
- **RAV-RUN1-R1-F005 [low] refresh errors** — Malformed endpoint records can raise `AttributeError` or `KeyError` outside the tool's concise error path.
- **RAV-RUN1-R1-F006 [low] refresh provenance** — A custom `--base-url` fetch still records the production URL in `_source`.

</details>

<details>

- **[low] refresh safety** — A same-directory temporary file can preserve the live snapshot until replacement. Cleanup and file-mode preservation keep that path tidy. (`RAV-RUN1-R1-P001`; covers `RAV-RUN1-R1-F001`)
- **[low] diagnostic scope** — The warning can name the observed alias relation. Directory splitting can stay a conditional result when both spellings are published. (`RAV-RUN1-R1-P002`; covers `RAV-RUN1-R1-F002`)
- **[medium] warning visibility** — PR #221 can remain the merge prerequisite. After rebase, this warning needs one check through the default rich CLI path. (`RAV-RUN1-R1-P003`; covers `RAV-RUN1-R1-F003`)
- **[medium] contributor guidance** — `datastore-gate.md` can state the alias trigger, warning-only result, and single-record evidence limit. (`RAV-RUN1-R1-P004`; covers `RAV-RUN1-R1-F004`)
- **[low] refresh errors** — Small checks for consumed fields can report the endpoint and record index without adding a schema layer. (`RAV-RUN1-R1-P005`; covers `RAV-RUN1-R1-F005`)
- **[low] refresh provenance** — `_source` can use the effective URL after removing user information, query parameters, and fragments. (`RAV-RUN1-R1-P006`; covers `RAV-RUN1-R1-F006`)

</details>

<details>

- Target: PR #230 (`validator-developer-slug-drift`, 8 files, +2073/-0)
- Run ordinal: 1
- Mix: 3 codex-exec
- Finding counts: 0 critical, 0 high, 2 medium, 4 low, 0 nit
- Checks: concerns=3; confirmed=1; narrowed=2; ruled-out=0; set-aside=0
- Second check: targeted; reviewers=2; kept=4; lowered=2; clarified=6; set-aside=0; removed=0; approval changed=no
- Set aside: 0 items

</details>

---

nelaturuharsha:
@borgr this will be handled by the registry work, so its a double-effort. I think we should wait for the registry to finalize this.

mrshu:
# ⚒️ review-anvil report

**Review decision:** COMMENT — Approval is disabled, and two earlier medium findings remain open.
**Result:** No new problems were found. Two medium and four low earlier findings remain present.
**Scope:** This PR adds offline publisher-alias data, its refresh tool, and a warning-only check for datastore directory splits.
**Checks:** Skipped because two reviewers independently confirmed the medium findings, and the other findings are low priority.
**Second check:** Targeted, 2 reviewers; 6 findings kept, 1 plan clarified, 3 plans set aside, 0 findings removed.

## Earlier review comments

<details>
<summary>Earlier review comments (6 items)</summary>

- **RAV-RUN1-R1-F003 [medium] warning visibility — still open.** The default rich output still returns after `PASS` before it shows warnings. [Earlier thread](https://github.com/evaleval/every_eval_ever/pull/230#discussion_r3725830949)
- **RAV-RUN1-R1-F004 [medium] contributor guidance — still open.** The datastore gate guide still has no entry for the new registered warning. [Earlier thread](https://github.com/evaleval/every_eval_ever/pull/230#discussion_r3725830951)
- **RAV-RUN1-R1-F001 [low] refresh safety — still present.** The refresh tool still writes directly to the live snapshot path. [Earlier review](https://github.com/evaleval/every_eval_ever/pull/230#pullrequestreview-4870883842)
- **RAV-RUN1-R1-F002 [low] diagnostic scope — still present.** The record-only check still states that two datastore directories exist. [Earlier review](https://github.com/evaleval/every_eval_ever/pull/230#pullrequestreview-4870883842)
- **RAV-RUN1-R1-F005 [low] refresh errors — still present.** Malformed endpoint members can still bypass the concise error path. [Earlier review](https://github.com/evaleval/every_eval_ever/pull/230#pullrequestreview-4870883842)
- **RAV-RUN1-R1-F006 [low] refresh provenance — still present.** A custom registry URL still records the production source URL. [Earlier review](https://github.com/evaleval/every_eval_ever/pull/230#pullrequestreview-4870883842)

</details>

## What I noticed

|---|---|---|---|---|
| RAV-RUN1-R1-F003 | medium | warning visibility | `every_eval_ever/validator/validation_core.py:801` | The check emits a warning on a valid report. The default rich renderer returns after `PASS` before it reads `report.warnings`. |
| RAV-RUN1-R1-F004 | medium | contributor guidance | `every_eval_ever/validator/validation_core.py:801` | The new registered rule is absent from the required `datastore-gate.md` contributor guide. |

<details>
<summary>Non-blocking low findings (4 items)</summary>

- **RAV-RUN1-R1-F001 [low] refresh safety** `every_eval_ever/tools/refresh_org_registry.py:211` — A stopped direct write can leave the packaged snapshot as partial JSON.
- **RAV-RUN1-R1-F002 [low] diagnostic scope** `every_eval_ever/validator/validation_core.py:641` — One record proves an alias relation and split risk, but not that both directories already exist.
- **RAV-RUN1-R1-F005 [low] refresh errors** `every_eval_ever/tools/refresh_org_registry.py:93` — `_get` checks only the outer list. Later record-shape errors can escape the concise error response.
- **RAV-RUN1-R1-F006 [low] refresh provenance** `every_eval_ever/tools/refresh_org_registry.py:139` — `--base-url` controls the requests, but `_source` always records the production constant.

</details>

## Things to try

- **[medium] contributor guidance** — A short `datastore-gate.md` section can describe the alias trigger, warning result, and record-local evidence limit. The required skill conversion test can verify the guide. (`RAV-RUN2-R1-P001`; covers `RAV-RUN1-R1-F004`)
- **[medium] warning visibility** — PR #221 can remain the required predecessor. After the rebase, a default-rich integration check can confirm that this warning is visible. (`RAV-RUN2-R1-P002`; covers `RAV-RUN1-R1-F003`)
- **[low] diagnostic scope** — Conditional wording can say that publishing under both names can create split directories. This keeps the warning within the evidence available from one record. (`RAV-RUN2-R1-P003`; covers `RAV-RUN1-R1-F002`)

- **RAV-RUN2-R1-P004 [low] refresh errors** — Set aside because the author plans to replace this refresh tool after PR #190 merges.
- **RAV-RUN2-R1-P005 [low] refresh provenance** — Set aside until the #190 rebase identifies the final shared snapshot format and source field.
- **RAV-RUN2-R1-P006 [low] refresh safety** — Set aside until atomic replacement can be assessed in #190's consolidated refresh tool.

<details>

- Target: PR #230 (`validator-developer-slug-drift`, 8 files, +2047/-0) at `53b97a253d78100f1bc7851ce777c127e7152921`
- Run ordinal: 2
- Rounds: 1/1 completed; adaptive off; no new findings after prior-feedback deduplication
- Mix: 3 codex-exec; 2 completed evidence-based reviews, and 1 could not access the immutable diff
- Focus: correctness, production blast-radius, simplicity, maintainability, constructive language, and optional safe suggestions
- Earlier review comments: 6 items; 2 open, 4 reported and still present, 0 fixed, 0 skipped
- Finding counts: 0 critical, 0 high, 2 medium, 4 low, 0 nit; 0 new findings
- Checks: skipped; no candidate required reproduction
- Second check: targeted; reviewers=2; kept=6; clarified=1 plan; set-aside=3 plans; removed=0; approval changed no
- Set aside: 3 plans; reason=planned consolidation and deletion after PR #190

</details>

---

borgr:
Updated for @nelaturuharsha's point — agreed, and done: this branch no longer carries any registry of its own.

**Dropped** `helpers/org_registry.py`, `tools/refresh_org_registry.py`, `helpers/data/org_registry.json` and `tests/test_org_registry.py` (428 code lines, 28 KiB of data). The check now imports `second_name_of` from #190's `helpers/eval_card_registry.py`. The two implementations were compared over every spelling in the published datastore plus every id and alias in both snapshots first: **2,151 spellings probed, 203 non-`None` answers, 0 disagreements**.

The diff is now 5 files, +338/-2: the check, its tests, one line in `validate.py`, a `datastore-gate.md` section, and one test.

**CI here is red and expected to stay red until #190 merges** — the import does not resolve on `main`, and because `validation_core` is imported almost everywhere, the whole suite fails to collect rather than just the new tests. With #190's four files overlaid locally: **424 passed, 20 skipped, `ruff check` clean**. The base can't be retargeted (#190 is on another fork), so please read the red as the ordering rather than as a broken branch. Happy to park this until #190 lands if that's easier to triage.

On the review findings:

- **[low] refresh safety / refresh errors / refresh provenance** (`RAV-RUN1-R1-F001/F005/F006`) — all three were in `refresh_org_registry.py`, which is gone. Whatever remains of them belongs to #190's refresh tool.
- **[low] diagnostic scope** (`F002`) — fixed, and it was a real inaccuracy: one record proves the alias relation and this record's directory, not that the other directory is populated. Three of the pairs in the sample (`MiniMaxAI`, `CohereLabs`, `XiaomiMiMo`) have no second directory at all. The message now says *filing under both* puts one publisher in two directories.
- **[medium] contributor guidance** (`F004`) — added a `§developer` section to `datastore-gate.md`, and rather than rely on the next author remembering, `test_every_registered_check_is_documented_in_the_gate_guide` asserts every `REGISTERED_CHECKS` entry is named in that guide. The docstring note that asked for it by hand is deleted. Confirmed from both sides: it fails when the section's reference is renamed, and the other five checks were already documented.
- **[medium] warning visibility** (`F003`) — still open, still #221. #221 is green and unchanged; this remains the one finding that cannot be closed inside this PR.

Merge order: **#190 → #221 → this**.

mrshu:
# ⚒️ review-anvil report

**Review decision:** COMMENT — This run is comment-only; two earlier findings remain on the current head.
**Result:** No new findings. The guide and registry cleanup address four earlier items; warning visibility and one narrow message claim remain.
**Scope:** Add a publisher-alias warning, reuse eval-card-registry vocabulary, and enforce contributor-guide coverage.
**Checks:** Skipped because all material feedback is carried from earlier reviews and was revalidated by three reviewers.
**Second check:** Targeted, 2 reviewers; both retained the warning-visibility item and narrowed the message wording item.

## Earlier review comments

- **RAV-RUN1-R1-F003 [medium] warning visibility** — Still present. A valid report returns from the default rich renderer before its warnings are shown. PR #221 remains the focused merge dependency.
- **RAV-RUN1-R1-F002 [low] diagnostic scope** — Narrow residual remains. The guide now states the evidence limit, but the metadata-only message still says a developer grouper reads two groups without proving both spellings exist in the evaluated corpus.
- `RAV-RUN1-R1-F004` is fixed by the new guide section and coverage test.
- `RAV-RUN1-R1-F001`, `RAV-RUN1-R1-F005`, and `RAV-RUN1-R1-F006` are stale after the local refresh implementation was removed.

## What I noticed

No new problems were confirmed in this round.

## Things to try

- **[medium] warning visibility** — The current merge order can remain #190, then #221, then #230. A rebase after #221 would make this warning visible in the default CLI. (`RAV-RUN1-R1-F003`)
- **[low] diagnostic wording** — The metadata-only branch could say that mixed spellings may create two developer groups when both are present. (`RAV-RUN1-R1-F002`)

<details>

- Target: PR #230 at `b92a4beee4309448322fb9fcd678e3b9ac84405b` (5 files, +338/-2)
- Run ordinal: 3
- Mix: 3 codex-exec
- Earlier review comments: 6 root findings; 2 still present, 1 fixed, 3 stale
- New finding counts: 0 critical, 0 high, 0 medium, 0 low, 0 nit
- Checks: skipped; no new material candidate required reproduction
- Second check: targeted; reviewers=2; kept=1; clarified=1; set-aside=0; removed=0

</details>

---

borgr:
Both open findings are addressed at `d92dd18`, and one of them by an upstream change rather than by this branch. Also re-measured everything against current `main` — the PR body had figures that are now wrong, and one claim in it was simply false.

### RAV-RUN1-R1-F003 [medium] warning visibility — closed, and the merge dependency is gone

The suggested order was #190 → #221 → #230. **#221 no longer exists**: it merged as `2f3258e9` and was then reverted by #239 (`c403e59f`, "Reverts evaleval/every_eval_ever#221"). What closes the finding is **#238**, merged since the last run, which makes warnings visible in the default `rich` renderer *and* sets the exit code to 2 on a warnings-only run.

Verified end to end on `main` + #190 + this branch, with a record filed at its canonical `mistral/` path:

```
│  WARN   Aggregate (EvaluationLog)
│   Warning: model_info.id: 'mistral' and 'mistralai' are one organization in
│ the eval-card-registry — filing under both puts one publisher in two
│ datastore directories, neither listing complete. Match how it is already
│ filed datastore-wide
│ 0 passed, 1 warning-only, 0 failed (0 errors, 1 warnings)
│ Valid locally, but not merge-ready. Fix all warnings.
EXIT=2
```

So the merge order is now **#190, then this** — nothing in between.

I did not add the suggested default-rich integration check. The renderer is now `main`'s code with `main`'s tests; a test here that ran the CLI to confirm #238's renderer prints a warning would be asserting #238's contract through this check as a fixture, and would go green for the wrong reason if the renderer changed. This branch tests its own boundary instead — `test_developer_slug_drift_warns_without_failing_validation` asserts `valid is True` with the warning present.

### RAV-RUN1-R1-F002 [low] diagnostic scope — fixed, and the residual claim removed rather than softened

The suggestion was conditional wording ("may create two developer groups when both are present"). I went further: the metadata-only branch no longer mentions a second group at all, because there is something the single record *does* prove and it is worth saying plainly. It now reads:

> `model_info.developer: 'mistral' and 'mistralai' are one organization in the eval-card-registry — the id prefix picks the directory here, so this field splits none, but it disagrees with that directory and groups this record apart from any spelled the other way. Match how it is already filed datastore-wide`

The disagreement with the directory the prefix chose is visible in this record. Whether a second group exists is a fact about *other* records, so the message asserts it neither flatly nor conditionally. `test_only_the_field_that_decides_the_directory_claims_the_split` pins both halves — `'two datastore directories'` present in the directory-branch message and absent from this one.

For what it is worth, the claim would have been true: on the published datastore, all **22** records that hit this branch do have other spellings of the same publisher present. The message just does not rely on that.

### F004, F001, F005, F006

Agreed with the last run's assessment: F004 is fixed by the `§developer` section plus `test_every_registered_check_is_documented_in_the_gate_guide`, and F001/F005/F006 went stale with the local refresh tool this branch no longer carries. One correction while in there — that test's result dict was named `documented` while holding the *undocumented* names; it is `missing` now.

### Body re-measured; one claim in it was false

The blast-radius figures were taken from a 500-file sample and a per-placement sample. They are now the whole published corpus, **13,185 aggregate records**: **361 warnings on 361 records** (2.7%) across 14 organizations, 339 naming the directory split and 22 metadata-only.

The false claim was **"exit code stays 0"**, in two places. Under #238 that is exit **2**. Sized it: of the 361, **246 already fail** on `main` for unrelated reasons and exit 1 either way, and **115 are otherwise clean, so this warning is what moves them from 0 to 2**. Those 115 are also the only warnings the whole corpus produces on a record that passes.

Also corrected: the six adapter paths (`converters/` → `adapters/` after #218), the 883 → **921** published publisher directory names in the follow-up sizing, and the 37/7/30 group split → **39 groups holding more than one directory, 10 differing by name and 29 only by case or punctuation**.

One number found on the way that is not this PR's business but should not go unrecorded: **11,136 of the 13,185 published records fail the current library validator with semantic checks on.** 11,103 of those are missing at least one of `deployment_type` / `model_availability`, and **9,998 fail for nothing else at all** — clearing those two fields would take the corpus from 2,049 passing to 12,047. Filed as #242.

CI here stays red until #190 lands — `validation_core` imports the registry helper, and it is imported nearly everywhere, so the suite fails to collect rather than failing a test. Locally on `main` + #190: 421 passed, 20 skipped, `ruff check` clean.

borgr:
Rebased onto current `main` (`abf0809`) — GitHub had this marked `CONFLICTING`, which was stale: the merge is clean and it now reports mergeable.

CI is red for exactly the one reason *Merge order* predicts and nothing else: `ModuleNotFoundError: No module named 'every_eval_ever.helpers.eval_card_registry'` at collection, so the suite never runs. That module is #190's vendored snapshot, which this branch deliberately does not copy. Locally on `main` + #190's two files: 422 passed, 20 skipped, `ruff check` clean.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The earlier concerns are resolved, and one path edge plus one wording refinement remain.
**Result:** One medium-priority finding and one low-priority suggestion remain.
**Scope:** Add a publisher-alias warning with exact datastore path semantics, one-warning grouping, and clear record-local diagnostics.
**Checks:** 2 concerns checked and confirmed.
**Second check:** 2 reviewers kept both findings, lowered the wording item, and clarified both fix paths.

## Earlier review comments

- Warning visibility and contributor guidance are fixed.
- Diagnostic scope is fixed for the earlier cases. The low-priority item below covers a distinct recorded-namespace message.
- Refresh safety, refresh errors, and refresh provenance are no longer relevant because the refresh tool left this PR.

## What I noticed

|---|---|---|---|---|
| RAV-RUN4-R1-F001 | medium | path semantics | `every_eval_ever/validator/validation_core.py:631` | A slash-bearing ID with an empty component can enter flat-ID behavior. The warning then names `model_info.developer` as path-defining although the publisher rejects the identity as unroutable. |

<details>

- **RAV-RUN4-R1-F002 [low] diagnostics** `every_eval_ever/validator/validation_core.py:622` — A metadata-only warning can omit the raw namespaced ID prefix that selects the actual datastore directory. The message then refers to “that directory” without naming it.

</details>

## Things to try

- **[medium] path semantics** — Identity parsing can distinguish flat, namespaced, and unroutable states before registry lookup. Leading, trailing, and repeated empty components can return no alias warning. (`RAV-RUN4-R1-P001`; covers `RAV-RUN4-R1-F001`)
- **[low] diagnostics** — The selected raw prefix can remain separate from alias observations and appear only in metadata-only wording. Organization grouping and registry calls can stay unchanged. (`RAV-RUN4-R1-P002`; covers `RAV-RUN4-R1-F002`)

<details>

- Target: PR #230 at `abf080913d7ede5886fb0f2259056850320325e4`
- Run ordinal: 4
- Mix: 3 codex-exec
- Earlier review comments: grouped into 6 topics; 3 fixed and 3 no longer relevant
- Finding counts: 0 critical, 0 high, 1 medium, 1 low, 0 nit
- Checks: concerns=2; confirmed=2; set-aside=0
- Second check: targeted; reviewers=2; kept=2; lowered=1; clarified=2 plans; removed=0; approval unchanged
- Set aside: 0 items

</details>

---

every_eval_ever/validator/validation_core.py
```diff
@@ -577,6 +578,76 @@ def _metric_bound(value: Any) -> float | None:
...
+        canonical = second_name_of(value)
+        if canonical is None:
+            continue
+        locations, spellings = declared.setdefault(canonical, ([], set()))
+        locations.append(location)
+        spellings.add(value.strip())
+
+    # The prefix decides the directory when the id has one, developer when it is
+    # flat. The other field still names the publisher, so it is reported too.
+    directory_field = 'model_info.id' if prefix else 'model_info.developer'
```
mrshu, line 645: **RAV-RUN4-R1-F001 [medium] path-semantics** — An unroutable namespaced ID can enter flat-ID warning behavior

For `/model`, `_developer_prefix()` returns an empty string. This branch then treats `model_info.developer` as path-defining, while `datastore_path_components()` rejects the same ID because its namespace is empty. A warning-only run can therefore exit `2` for a routing claim the publisher cannot make.

Identity parsing can distinguish flat, namespaced, and unroutable states before registry lookup. Leading, trailing, and repeated empty components can return no alias warning.

borgr:
RUN4 — both items addressed.

**RAV-RUN4-R1-F001** (medium, path semantics) — already fixed before this review landed (d400cb8ce, targeted at `abf08091` which predates it). `check_developer_slug` now returns early for a slash-bearing id with any empty component (`/model`, `org/`, `a//b`): that id is what `datastore_path_components` rejects by the same `any(not part …)` rule, so the record files nowhere and this check has no split to report. The empty prefix no longer reads as a flat id and pins the split on `model_info.developer` — a routing claim the record can't make. It's left to the §path/schema errors, with a regression test (`test_an_unroutable_id_splits_no_directory_and_is_left_to_the_path_error`) covering leading/trailing/repeated empty components.

**RAV-RUN4-R1-F002** (low, diagnostics) — fixed (97cfa19a7). The metadata-only branch now names the raw prefix that selects the directory instead of an unnamed "that directory": e.g. *"the `'mistralai'` id prefix picks the directory here, so this field splits none, but it disagrees with that directory …"*. The change is confined to that wording — organization grouping and the registry lookups are unchanged.

borgr:
Reviewed the 3 unresolved automated review threads against the current head. All three (`validation_core.py:645/806/809`) are already satisfied by the code as written — the registry-alias check and its guards do what the threads ask. Resolving them.

borgr:
The four red CI rows were not the rebase. `validation_core.py` imported `every_eval_ever.helpers.eval_card_registry.second_name_of`, and that module exists on no branch in this repo or any fork — so importing the validator raised `ModuleNotFoundError` and all 17 test modules errored during collection. The check was written against a helper that was never written.

Fixed in `e2c6cf693` by writing it, following the pattern the lexam and paperswithcode adapters already use rather than inventing a new one:

- `helpers/publisher_snapshot.json` — a vendored snapshot of the registry's curated `seed/orgs.yaml`, 78 publishers with their aliases and recorded namespace, taken at registry `0052cd2c4`.
- `helpers/refresh_publisher_snapshot.py` — regenerates it, records the registry revision, and reads only the curated seed. `orgs.generated.yaml` is 900 HF community namespaces with no aliases, so it says nothing about one publisher having two names and would make the snapshot's own diffs unreadable.
- `helpers/eval_card_registry.py` — `second_name_of`, resolving against that snapshot with no registry install and no network.

The resolution rule is the one this PR's tests already specified, and all of them pass unchanged: every canonical id and recorded namespace is collected across all publishers *before* any alias resolves, so `zai-org` and `meta-llama` stay silent even though the registry lists them as aliases too. Normalization folds case and punctuation, so `Anthropic`, `moonshot-ai` and `z-ai` are variants of a canonical id rather than second names, while `mistral`, `glm`, `kimi`, `granite` and `zhipu-ai` resolve.

`main` is also merged in (`915d90ac1`): it gained `check_model_identity_path` from #223 in the same three places this check registers, so both sit side by side.

New in `tests/test_publisher_registry.py`, 10 tests on the resolver itself, including one that fails if a future refresh stops exercising the namespace-shadowing case rather than passing vacuously.

`uv run pytest tests` 880 passed, 31 skipped; `uv run ruff check` clean.

One thing for the review to weigh: this adds a shared helper under `every_eval_ever/helpers/` that the validator depends on, which is a wider blast radius than the warning itself. The snapshot is authoritative-at-snapshot-time, so a publisher whose second name the registry learns later is not recognized until a refresh — silence rather than a wrong warning, which is the safe direction, but it does mean the check's coverage is a committed file somebody has to remember to refresh.

borgr:
Correction to my last comment, and a decision this PR needs before it can go green.

**First, the actual root cause of the red CI**, which was worse than "the module was never written". `.git/info/exclude` in my local checkout carried the line `every_eval_ever/helpers/eval_card_registry.py`, so the module existed only in a working tree and `git add` silently skipped it every time. That is why no branch or fork had it. The exclusion is gone and the file is tracked in `1e091ae4b`. `fa3c548a8` also adds it to `package-data`, which it needed for the same reason #202 existed — an installed package would have imported the resolver and then failed to read its snapshot.

`core / locked` and `core / loose` are green now. `full` is not, and that is the real finding.

**The check fires on our own converter output.** `test_conversion_passes_the_merge_gate[inspect]` fails because a converted record warns:

    model_info.id and model_info.developer: 'mistral' and 'mistralai' are one
    organization in the eval-card-registry

The Inspect converter takes the publisher from the first segment of the model string (`inspect/utils.py:57`), so a run against `mistral/mistral-large-latest` files under `mistral/` while the registry's canonical id is `mistralai`. The check is right — the datastore does then hold that publisher under two directories — and #244's gate is right to refuse a record that warns.

**The reach is not one fixture.** 51 of the 78 curated publishers have at least one second name that fires, 109 spellings in total, and 47 of those are slug-shaped — `GLM`, `Granite`, `DeepMind`, `Eleuther`, `Alibaba-NLP`, `EXAONE`, `ERNIE` — exactly the shape a model string carries. So this warning lands on any converter or adapter that takes the publisher from a provider slug, which is most of them.

**So this is the #222 situation again, one size up.** There the warning was right and the metric ids were missing, and the fix was to land the ids first (#246). Here the warning is right and the converters emit provider slugs where a publisher belongs. Which means a decision, not a patch:

1. **Converters emit the canonical publisher.** Resolve through this snapshot at conversion time. Correct, and a material change in outcome — it changes the datastore directory of existing records and their join keys, so it needs its own PR and a human read.
2. **Narrow the check to two spellings inside one record.** Warn only when `model_info.id`'s prefix and `model_info.developer` disagree while resolving to one publisher, and stay silent when a record consistently uses a second name. That clears the gate today and gives up most of what the check is for, since the common case is a record that is internally consistent and split from the rest of the datastore.
3. **Land it as-is and accept the gate failure**, which is not an option while #244 is in.

I would take 1, sequenced as its own PR ahead of this one, and I am not going to pick it unilaterally — option 1 rewrites publisher directories for already-published records, which is the kind of change #256's dedup discussion is about.

Left the branch red rather than making the gate pass, so the finding stays visible.

borgr:
Closing this — the question it turns on is already settled, and this check is on the wrong side of it.

`datastore-gate.md` §path, once #271 lands, and `registry.md` today: the datastore's `<developer>` folder comes from `model_info.id`'s prefix, not from `model_info.developer` and not from the registry's `org_id`. `registry.md` puts it directly — *"Don't 'fix' the prefix to match the org."* `check_developer_slug` warns whenever a publisher directory is a registry alias rather than the canonical org, which is that prefix-fixing asked for as a validator rule.

It is also not fixable by narrowing. Under the registry's two-tier model an id prefix and `model_info.developer` differ **by design** — `Qwen/Qwen3-32B` with developer `alibaba` — so a check comparing the two fields would fire on correct records. That is why this turned `test_conversion_passes_the_merge_gate[inspect]` red on `mistral/`: not a bug found, a rule asserted that the repo has decided against.

What was real here survives:

- **The split directories.** Filed as #272 with the datastore evidence — six directories for one company, 996 in total. The cause is narrower than this check assumed: `get_developer` returns a prefix for a slashed name and a company fold for a bare one, so one model gets two directories depending on how a source wrote it.
- **The resolver and snapshot** (`helpers/eval_card_registry.py`, `publisher_snapshot.json`, `refresh_publisher_snapshot.py`) are the mechanism #272's fix wants, offline and cached. I have left them on this branch rather than deleting them; happy to lift them into that PR when it is written, or drop them if it resolves live through the hosted endpoint like `open_medical_llm` does.
- **The documentation gap** that let this get built at all is #271. The rule existed in one aside in one skill file, while the schema and the gate said nothing and `fields.md` pointed the other way.

Reopen if you read §path differently — the whole disagreement is one sentence about which of the registry's two fields a directory is.

PR #234, opened 2026-08-06
Split contributor docs into CONTRIBUTING.md, and write down the review policy

borgr:
## What

Three things, one per commit:

1. **Split the contributor docs out of the README into a root `CONTRIBUTING.md`**, per audience — *understand & use* stays in the README, *author & submit* moves out.
2. **Restore the repo-wide PR template**, recovered from where it was parked.
3. **Write down the review policy** — the two lanes, and the rule about agreeing structural work before opening a PR.

Docs only; no `.py` files touched.

## Why now

An earlier attempt at (1) and (2) was carved out of #208 by 941aa85c1, to keep that PR scoped to the skill — *"a repo-wide PULL_REQUEST_TEMPLATE.md and a new issue template are a separate maintainer decision and shouldn't gate them."* That was right. The issue template was later pulled back in and landed; the PR template went to a branch that was never opened. The `CONTRIBUTING.md` half never reached GitHub at all. This reunites them and adds the policy that prompted the question.

## The split

The README was serving two audiences at once. `Contributor Guide`, PR/UUID naming, how-to-add and `Schema Instructions` move to `CONTRIBUTING.md` (409 → 302 lines).

`Instance-Level Data` and `Agentic Evaluations` **stay** — they document the record format, which a consumer needs as much as an author — and are promoted `###` → `##` now that the Contributor Guide no longer encloses them.

The earlier version's commit message said "slim README to a pointer" but only *added* `CONTRIBUTING.md`, leaving the README untouched and the two duplicating each other. This does the removal. Stale references fixed on the way: the root `eval.schema.json` symlink is gone (→ `every_eval_ever/schemas/`), `utils/` → `every_eval_ever/adapters/`, `schema_version` `0.2.3` → `0.3.0`, and `post_codegen` runs as a module.

The skill-sync section deliberately carries **no** checklist of skill files to update. `tests/test_skill_conversion.py` is the enforcement; the table that used to sit there is exactly what went stale.

## The policy, and two things it needed pinned down

**The size limit counts hand-written lines.** As a raw diff count it contradicts its own flagship case — "a new adapter" is the fast lane, yet adapters routinely pass 1000 lines on fixtures alone, and any schema touch drags in a regenerated `eval_types.py`. So the two generated type modules and `tests/data/**` don't count. Only there: generated files dumped anywhere else count in full, because bulk output landing in the package is the kind of thing review is *for*.

**`auto-review` is maintainer-applied.** Self-serve would make the fast-lane gate self-declared with nothing checking it, which is the one property the gate can't have. The label has been created on this repo.

**Placement follows this repo's own rule about not restating things.** `CONTRIBUTING.md` carries the policy. `AGENTS.md` and the skill get only the half that changes what an agent *does* — get the design agreed before building something structural, and don't fold a schema tweak into an adapter PR, since that drags the adapter into the slow lane behind a design discussion. The PR template gets a lane checkbox and a `Design agreed in:` slot, which is where the rule actually bites.

## Verification

- `pytest tests/test_skill_conversion.py` — 7 passed (the skill tripwire)
- `ruff check` — clean
- every internal markdown link and anchor resolves, checked programmatically across all `.md` files

**Pre-existing failure, unrelated to this branch:** `tests/test_helm_generation_args.py::TestExtractGenerationArgsFalsyValues` has 5 failures on current `main` — zero-valued `temperature`/`top_p`/`top_k` and a low `max_tokens` aren't preserved, which reads like a falsy-value `or` fallback. Worth its own issue.

## Review lane

**Needs a human** by its own policy — cross-file docs restructure, material change to contributor-facing guidance. The wording is the substance here, so please read it as wording.

### DIFF
--- a/.claude/skills/eee-dataset-conversion/SKILL.md
+++ b/.claude/skills/eee-dataset-conversion/SKILL.md
@@ -87,8 +87,12 @@ dataset, a harness dump) and you must emit EEE records. Two artifacts:
 7. **Ask, then log your decisions.** Two channels, don't confuse them:
    - **Ask the operator (live)** when a choice *sets policy*: creating a new
      canonical id · dropping a non-trivial share of the data · an ambiguous metric
-     choice · bounding an unbounded metric · re-hosting large data. Don't decide
-     these silently — the person running you is there to answer.
+     choice · bounding an unbounded metric · re-hosting large data · **the source
+     won't fit without a structural change** (a schema field, an edit to a base
+     adapter or a shared converter, relaxing a validator rule). Don't decide these
+     silently — the person running you is there to answer. The structural one is not
+     yours to fold into this PR: its design gets agreed before a PR exists, and
+     carrying it here would hold the adapter behind that discussion.
    - **Log (in the PR)** every *non-obvious* choice — not just where it was hard. A
      confident wrong choice produces no "friction," so log decisions, not pain.
    Finish with a ready-to-merge PR carrying the decision log below. General gaps
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,35 @@
+<!-- Thanks for contributing! Delete sections that don't apply. -->
+
+## What / source
+<!-- What does this change do? For an adapter: what source does it convert, and to what grain? -->
+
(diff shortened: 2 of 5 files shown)


nelaturuharsha (commented):
For now, all good!

Issue #164, opened 2026-06-09
[adapter] Update bfcl adapter


nelaturuharsha:
https://github.com/evaleval/every_eval_ever/tree/main/utils/bfcl

PR #241, opened 2026-08-09
docs: harden datastore PR review workflow

nelaturuharsha:
## What / source

Harden the EEE datastore PR-review skill based on live reviews of Hugging Face datastore PRs 129, 145, 151, and 164.

The skill now emits mandatory incremental receipts and one-minute heartbeats, binds deployment-metadata approval to both a proposal SHA-256 and the exact datastore PR head, and distinguishes record omissions from metadata that remains unavailable after a documented first-party search.

PR 164 exposed two further cases now covered here. Citation-backed aggregators do not establish the deployment used by every evaluation they collect: the reviewer must follow per-result provenance, must not infer a deployment difference from generation parameters or multiple citations alone, and must not flatten proven conflicting runs to one model-wide value. Score repairs must restore the source value to the metric's declared scale and preserve the conversion instead of widening bounds; bare non-finite JSON bounds are serialized as the schema-supported strings without changing their meaning.

Approved decisions are also preserved with the collection. The byte-identical approved proposal is copied to `data/<collection>/deployment-metadata-proposal-<digest-prefix>.md`, and an append-only `data/<collection>/REVIEW_DECISIONS.md` entry records the PR/head, approval receipt, substantive decisions, evidence, validation, and follow-ups. This gives each collection durable provenance and a changelog rather than leaving the audit trail only in temporary notes or discussion history.

Follow-up to #240.

## Review lane

- [x] **Fast** — five skill files, scoped agent-workflow guidance, references, and Markdown templates; no runtime behavior
- [ ] **Needs a human** — not a schema, validator, runtime, or data-output change

Design agreed in live operator feedback while forward-testing the skill on `evaleval/EEE_datastore` discussions 129, 145, 151, and 164.

## Checklist

- [x] Skill structure validation clean
- [x] Proposal template retains exactly the five operator-requested review columns
- [x] Approval is bound to both proposal SHA-256 and the current HF PR head
- [x] Live PR mutation is forbidden while approval is pending
- [x] Approved proposal is copied byte-for-byte into each affected collection root
- [x] `REVIEW_DECISIONS.md` is append-only and records approval, evidence, validation, and follow-ups
- [x] Substantive non-deployment corrections are included in the collection changelog
- [x] Record absence is distinguished from source unavailability
- [x] Comprehensive first-party search surfaces and a concrete stop rule are defined
- [x] Only high/medium categorical proposals are permitted; low confidence cannot become a guess
- [x] Unknown rows include compact coverage codes, confusion/reason, and evidence needed to resolve them
- [x] Aggregator provenance is separated from run-level deployment evidence
- [x] Multiple citations or generation settings do not create unsupported deployment groups
- [x] Repeated model ids can be scoped to distinct evidence groups without changing the five-column format
- [x] Normalized score conversions preserve the raw source value and explicit math
- [x] Bare `Infinity` values are repaired without changing unbounded semantics
- [x] No schema, validator, adapter, registry, or datastore data changes
- [x] Unrelated local `.gitignore` change excluded

## Decisions & coverage

- Decision / where: progress checkpoints in the core workflow
  Chose / instead of: independently verifiable phase receipts with bounded polling / final-only reporting
  Confidence (high/med/low): high
  General? (yes/no): yes

- Decision / where: deployment research approval artifact
  Chose / instead of: a compact Markdown table with source IDs and a head-bound digest / verbose JSON or implicit approval from a general fix request
  Confidence (high/med/low): high
  General? (yes/no): yes

- Decision / where: collection provenance
  Chose / instead of: byte-identical approved proposal plus append-only collection decision log / temporary run notes or discussion-only history
  Confidence (high/med/low): high
  General? (yes/no): yes

- Decision / where: aggregator deployment metadata
  Chose / instead of: follow each result to run-level evidence and group only evidence-equivalent files / infer deployment from the aggregator, model id, citation count, or generation settings
  Confidence (high/med/low): high
  General? (yes/no): yes

- Decision / where: out-of-range source scores
  Chose / instead of: source-backed conversion to the declared metric scale with raw value and math retained / widening bounds or clamping the score
  Confidence (high/med/low): high
  General? (yes/no): yes

- Decision / where: missing reproducibility metadata
  Chose / instead of: separate `record_absent`, `research_incomplete`, `available_not_surfaced`, `conflicting_sources`, and `unavailable_after_search` statuses / treating an adapter omission as a claim about the underlying evaluation
  Confidence (high/med/low): high
  General? (yes/no): yes

- Decision / where: low-confidence deployment/availability classification
  Chose / instead of: continue research, then propose `unknown` only after the stop rule / provider-, family-, or default-based guessing
  Confidence (high/med/low): high
  General? (yes/no): yes

**Coverage:** six workflow checkpoints, one head-bound approval artifact, collection-root proposal and decision-log templates, a six-surface missingness ledger, aggregator/run grouping rules, source-scale repair guardrails, axis-specific confidence/source shorthand, and explicit adapter/registry follow-ups; no source rows or EEE records changed.

**Operator asked about policy calls?** Yes. The operator chose explicit human approval before applying researched deployment metadata, durable collection-root provenance, and a bounded evidence standard before describing metadata as unavailable.

## Verification

- Skill Creator structural validation: `Skill is valid!`
- `git diff --check`

### DIFF
--- a/.agents/skills/eee-datastore-pr-review/SKILL.md
+++ b/.agents/skills/eee-datastore-pr-review/SKILL.md
@@ -5,8 +5,8 @@ description: >-
   dataset. Use when given an EEE_datastore discussion or PR URL, asked to run
   or reproduce `/eee validate changed`, resolve EEE validator errors or
   warnings, research model deployment_type or model_availability, edit the
-  changed datastore records, rerun the bot, or prepare canonical-registry
-  follow-ups.
+  changed datastore records, prepare human-approved deployment metadata
+  proposals, rerun the bot, or prepare canonical-registry follow-ups.
 ---
 
 # Review and repair an EEE datastore PR
@@ -31,6 +31,9 @@ not sufficient.
 - If asked only to review, prepare a patch and findings without uploading or
   commenting. If asked to fix, update the supplied PR, trigger its validator, and
   iterate on that same ref.
+- Before changing `deployment_type` or `model_availability` on the live PR, obtain
+  explicit human approval of a proposal bound to the current PR head. Authorization
+  to fix the PR is not approval of research-derived field values.
 - Ask the operator before a policy decision: minting a new canonical id, changing a
   schema/validator rule, dropping non-trivial data, choosing an ambiguous metric or
   bound, or making another structural change. Do not hide such a choice in a data
@@ -69,8 +72,9 @@ Required checkpoints:
 
 1. **Snapshot:** after selecting the PR head and matching bot run.
 2. **Diagnosis:** after reproducing the gate and grouping its findings.
-3. **Evidence and patch plan:** after resolving model-specific evidence and before the
-   first live PR mutation. Include every proposed field change and any unresolved axis.
+3. **Research proposal:** after resolving model-specific evidence. Render
+   `assets/deployment-metadata-proposal.md`, report its SHA-256, and pause for explicit
+   human approval before editing deployment fields or mutating the live PR.
 4. **Local repair:** after the repaired diff passes local validation and content review.
 5. **Remote receipt:** immediately after each uploaded commit or validator-trigger
    comment, including the returned commit SHA or discussion event id.
@@ -80,9 +84,9 @@ Required checkpoints:
 For a phase lasting more than 60 seconds, emit a heartbeat at least once per minute
 with the current evidence surface or bounded poll, completed/remaining counts, and
 whether local or remote state changed. Use bounded polling calls of at most 45 seconds
-so progress messages can be delivered. A checkpoint is not an approval gate: continue
-unless the operating contract requires an operator decision or the operator asked to
-pause at that phase.
+so progress messages can be delivered. Continue after ordinary checkpoints. The
+research proposal is an approval gate; do not continue past it without an explicit
+approval matching both its digest and PR head.
 
 ## Workflow
 
@@ -154,12 +158,33 @@ appendix, source repository, and official API/release documentation. Use current
 research where facts may have changed, but pin the evidence revision or date relevant
 to the submitted evaluation.
 
-Batch models only after proving that they share the same evidence. Keep an evidence
-table with raw model label, canonical model id, both decisions, source URL/revision,
-and confidence.
+Batch models only after proving that they share the same evidence.
+
+Before editing either deployment axis:
+
+1. Copy `assets/deployment-metadata-proposal.md` to run notes outside the datastore
+   repository and fill one row per exact submitted `model_info.id`. Do not substitute
+   a folder slug or an unapproved canonical id.
+2. Use only the five table columns in the template. Reference sources as `S1`, `S2`,
+   and so on; list each full URL and pinned revision/date once below the table.
+3. Include every model whose deployment fields would change, including mechanical
+   vocabulary migrations. State unresolved axes as `unknown` rather than omitting a
+   row.
+4. Compute the completed file's SHA-256. Return the rendered Markdown (or a clickable
+   path in a shared workspace), digest, PR head, model count, and affected-file count.
+5. Stop and request explicit human approval of that digest at that PR head. Do not edit
+   the records, upload a commit, or comment on the Hugging Face discussion while
+   approval is pending.
+
+Approval covers only the exact table and head named by the human. Before applying it,
+re-fetch `refs/pr/<number>`. If the head changed, evidence changed, a proposed value
+changed, or a new model entered scope, regenerate the artifact and obtain approval
+again. Record the approver and approval time in the decision log.
 
 ### 5. Make the repair
 
+- For deployment metadata, begin only after the research proposal is explicitly
+  approved and the remote head still matches it. Apply only its approved rows.
 - Edit only files implicated by a finding. Avoid mass reformatting unrelated data.
 - Preserve UUID filenames and stable evaluation identities unless identity itself is
   the defect.
@@ -189,6 +214,9 @@ When the task authorizes a fix, upload exact add/delete operations to the existi
 `refs/pr/<number>` with `huggingface_hub.HfApi.create_commit`; set the current PR head
 as `parent_commit` so concurrent updates fail instead of being overwritten. Never set
 `create_pr=True` for a repair round.
+
+If the commit changes either deployment axis, require the approved proposal digest and
+head in the decision log before uploading. General authorization to fix is insufficient.
 
 After the commit lands:
 
@@ -230,6 +258,8 @@ Return:
 - local validation and duplicate-check results;
 - content spot-checks performed;
 - deployment/availability evidence table, including researched `unknown` values;
+- research proposal path or rendered table, SHA-256, approved head, approver, and
+  approval time;
 - registry and adapter follow-ups with cross-links or candidate tables;
 - decision log and any unresolved blocker.
 
--- a/.agents/skills/eee-datastore-pr-review/SKILL.md
+++ b/.agents/skills/eee-datastore-pr-review/SKILL.md
@@ -6,7 +6,8 @@ description: >-
   or reproduce `/eee validate changed`, resolve EEE validator errors or
   warnings, research model deployment_type or model_availability, edit the
   changed datastore records, prepare human-approved deployment metadata
-  proposals, rerun the bot, or prepare canonical-registry follow-ups.
+  proposals, distinguish record omissions from genuinely unavailable metadata,
+  rerun the bot, or prepare canonical-registry follow-ups.
 ---
 
 # Review and repair an EEE datastore PR
@@ -23,9 +24,11 @@ not sufficient.
   declared bounds, or change a value merely to silence the validator.
 - Make `unknown` a researched conclusion, not a default. Record which relevant
   surfaces were checked before retaining it.
-- Distinguish absent record metadata from unavailable source evidence. A missing or
-  null `model_info.additional_details` object means the record needs investigation;
-  it does not establish either axis as `unknown`.
+- Distinguish absent record metadata from unavailable source evidence for every field.
+  Describe an unchecked absence as "not surfaced in the submitted record," not as
+  missing from the underlying evaluation. A missing or null
+  `model_info.additional_details` object means the record needs investigation; it does
+  not establish either axis as `unknown`.
 - Keep work on the supplied `refs/pr/<number>` ref. Do not open a replacement PR for
   another repair round.
 - If asked only to review, prepare a patch and findings without uploading or
@@ -47,6 +50,10 @@ Before editing, read these sibling references:
 - `../eee-dataset-conversion/reference/fields.md`
 - `../eee-dataset-conversion/reference/datastore-submission.md`
 - `../eee-dataset-conversion/reference/verification.md`
+
+Read `reference/metadata-missingness.md` whenever a field is absent, null, defaulted,
+or claimed to be unavailable. Apply it to deployment metadata and reproducibility
+fields such as temperature and maximum output tokens.
 
 Read `reference/model-deployment.md` whenever either model deployment axis is
 missing, stale, invalid, or suspicious. Read
@@ -136,6 +143,12 @@ Group findings by root cause rather than by file. For each group, record:
 - the source evidence needed for a correct fix;
 - proposed change and confidence.
 
+Classify every apparent omission with `reference/metadata-missingness.md`. Do not call
+an absent field genuinely missing while its status is `record_absent` or
+`research_incomplete`. If a README, eval card, methodology page, leaderboard, paper,
+repository, or API exposes it, classify it as `available_not_surfaced` and identify
+the adapter/submission gap.
+
 Inspect content even when the validator omits it. At minimum check suspicious zeroes,
 score scale and bounds, metric identity, `source_data`, duplicate overall/subtask
 aggregates, stable `evaluation_id`, model identity, answer leakage, and companion
@@ -148,9 +161,10 @@ actually supplied `additional_details`, supplied only one axis, or supplied neit
 
 ### 4. Research ambiguous metadata
 
-For deployment warnings, apply `reference/model-deployment.md` to each exact model
-variant and evaluation run. Determine the two axes independently. Do not infer one
-from the other, from the developer folder, or from a provider-wide rule.
+Apply `reference/metadata-missingness.md` before deciding any absent field is truly
+unavailable. For deployment warnings, then apply `reference/model-deployment.md` to
+each exact model variant and evaluation run. Determine the two axes independently. Do
+not infer one from the other, from the developer folder, or from a provider-wide rule.
 
 Search all relevant primary surfaces before choosing `unknown`: record payload and
 run config, generating adapter, pinned model card, evaluator methodology, paper and
@@ -166,10 +180,13 @@ Before editing either deployment axis:
    repository and fill one row per exact submitted `model_info.id`. Do not substitute
    a folder slug or an unapproved canonical id.
 2. Use only the five table columns in the template. Reference sources as `S1`, `S2`,
-   and so on; list each full URL and pinned revision/date once below the table.
+   and so on; list each full URL and pinned revision/date once below the table. Record
+   deployment (`D`) and availability (`A`) confidence and sources separately inside
+   their shared cells.
 3. Include every model whose deployment fields would change, including mechanical
-   vocabulary migrations. State unresolved axes as `unknown` rather than omitting a
-   row.
+   vocabulary migrations. Propose `unknown` only for `conflicting_sources` or
+   `unavailable_after_search`, and complete the template's unknown-rationale table.
+   If any axis is `research_incomplete`, do not finalize the proposal.
 4. Compute the completed file's SHA-256. Return the rendered Markdown (or a clickable
    path in a shared workspace), digest, PR head, model count, and affected-file count.
 5. Stop and request explicit human approval of that digest at that PR head. Do not edit
@@ -196,6 +213,9 @@ again. Record the approver and approval time in the decision log.
 - If generated records are wrong, fix or prepare the generating adapter in the code
   repo as well; otherwise the next refresh will restore the defect. Keep adapter code
   out of the datastore PR and cross-link its separate PR.
+- Treat `available_not_surfaced` as an extraction defect: backfill the approved value
+  in the data repair and prepare an adapter/submission follow-up so regeneration does
+  not erase it.
 - Review the resulting diff for accidental deletion, unrelated churn, and a mechanical
   replacement applied to semantically different models.
 
@@ -257,6 +277,8 @@ Return:
 - files changed, grouped by root cause;
 - local validation and duplicate-check results;
 - content spot-checks performed;
+- missingness classification for each absent field, including source-available
+  extraction gaps and the surfaces checked before any unavailable claim;
 - deployment/availability evidence table, including researched `unknown` values;
 - research proposal path or rendered table, SHA-256, approved head, approver, and
   approval time;
--- a/.agents/skills/eee-datastore-pr-review/SKILL.md
+++ b/.agents/skills/eee-datastore-pr-review/SKILL.md
@@ -62,6 +62,8 @@ run receipts, not Hugging Face discussion comments: do not post them to the PR u
 the operator explicitly asks. Each checkpoint must include the phase, current PR head
 SHA, facts established since the prior checkpoint, affected file/model counts, command
 exit statuses or evidence URLs when applicable, blockers, and the next action.
+When a progress or parent-message channel is available, send the checkpoint through it
+and continue in the same run. Do not end a turn merely to deliver a checkpoint.
 
 Required checkpoints:
 
--- a/.agents/skills/eee-datastore-pr-review/SKILL.md
+++ b/.agents/skills/eee-datastore-pr-review/SKILL.md
@@ -37,6 +37,10 @@ not sufficient.
 - Before changing `deployment_type` or `model_availability` on the live PR, obtain
   explicit human approval of a proposal bound to the current PR head. Authorization
   to fix the PR is not approval of research-derived field values.
+- Preserve approved decisions with the data. After approval, copy the byte-identical
+  proposal into each affected collection root and append its receipt to that
+  collection's `REVIEW_DECISIONS.md`; do not leave provenance only in temporary run
+  notes or discussion history.
 - Ask the operator before a policy decision: minting a new canonical id, changing a
   schema/validator rule, dropping non-trivial data, choosing an ambiguous metric or
   bound, or making another structural change. Do not hide such a choice in a data
@@ -153,7 +157,11 @@ Inspect content even when the validator omits it. At minimum check suspicious ze
 score scale and bounds, metric identity, `source_data`, duplicate overall/subtask
 aggregates, stable `evaluation_id`, model identity, answer leakage, and companion
 pairing. An out-of-range score requires finding the source scale or source value; do
-not cap, clamp, or round it into validity.
+not cap, clamp, round it into validity, or widen the bounds around it. When a source
+publishes a normalized value, convert it to the metric's declared scale and preserve
+the raw value plus explicit conversion in `score_details.details` or the decision log.
+Serialize unbounded limits as the schema-supported JSON strings `"Infinity"` and
+`"-Infinity"`; replacing a bare non-finite token must not change its meaning.
 
 Inspect the raw JSON before constructing an `EvaluationLog`. The model layer may
 auto-fill absent deployment keys with `unknown`, hiding whether the contributor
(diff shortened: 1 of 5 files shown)


borgr:
> *Edited: item 3 first proposed a non-finite-bounds check on a wrong premise — the validator already rejects those. Corrected below, with a scan of all 13,228 records on datastore `main` sizing what's left.*

Reviewed against the schema, `REGISTERED_CHECKS`, and the live datastore. The aggregator
rules in `model-deployment.md` are the best thing here — "an aggregator's adapter
establishes aggregation provenance, not the deployment used by every cited evaluation",
plus the three separations (same id ≠ deployment-equivalent runs, generation params ≠
deployment, citation count ≠ deployment). On #164 that rule is the difference between a
correct review and one mass-mislabel across 189 model directories. `available_not_surfaced`
and `source_access_blocked` → `research_incomplete` are also real and worth keeping.

A patch for the mechanical parts is at the bottom of this comment — duplication with
`datastore-gate.md` §score, `score_details.details` string typing, the harness-specific
checkpoint section, an `additional_details` kind-rule, and the missing merged-mid-review
branch. It applies cleanly on `5f861a59b`; take it or redo it as you prefer. Everything
else below needs a decision rather than an edit.

## Decisions this needs before it can land

**1. Do review artifacts go into `data/` at all?** This is the structural question, and
about 120 lines of the PR hang on it. `data/` is documented as exactly
`data/{collection}/{developer}/{model}/{uuid}.json[l]`; `check_path_structure` errors on
any other depth and `validate_path` returns `unsupported_extension` for `.md`. On #151
the `.md` rode through because the bot's changed-file filter happens to skip it — that's
an implementation accident, not a guarantee. Separately, `REVIEW_DECISIONS.md` is a
mutable per-collection singleton, and UUID-per-file exists precisely so concurrent
submissions never contend; "append-only" is unenforceable on a repo with no CI.

Our recommendation: don't. Provenance for a *value* already has the right home, and this
PR's own run used it well — `deployment_source: "…arxiv.org/html/2607.08842v2 (official
cloud API evaluation)"`, attached to the record it explains. Provenance for a *decision*
goes in the HF discussion, which is durable and API-readable (the PR body's premise that
discussion history is temporary isn't right — #151's full thread, every bot verdict and
fingerprint, is intact). An index across decisions, if you want one, belongs in this repo
where review and CI exist. Keeping `data/` a pure record tree is what lets the bot filter,
`rglob('*.json')`, the flat builder and the parquet builder all stay one line long.

**2. Which fields the gate covers.** The highest-value call here. The gate as written
covers the two deployment axes. The l2-bench run shows the axes came out right first
time; what needed a post-merge revert (#188) was an identity rewrite the bot passed
green: submitted `alibaba/alibaba/qwen3-32b` → merged as `Qwen/Qwen3-32B` → reverted to
`alibaba/qwen3-32b`, three identities in an hour, with the merged state having
`model_info.id = Qwen/Qwen3-32B` sitting at path `…/alibaba/qwen3-32b/…`.

Suggested boundary — gate join keys and closed vocabularies, not open annotation:

- **Gated:** `model_info.id`, `name`, the `<developer>/<model>` path, `evaluation_id`,
  `metric_config.metric_id`, `min_score`/`max_score`, `score`, `deployment_type`,
  `model_availability`. Other records join on these, or a published number is interpreted
  through them, and each has a fixed vocabulary or a canonical registry — so "is this
  right?" is answerable, which is what makes an approval mean something.
- **Not gated, constrained by kind:** `additional_details`, `score_details.details`,
  `generation_config.additional_details`. Underspecified on purpose; the rule is
  source-attributed facts about this record, never reviewer process state. (The patch adds
  that line. It's what #188 had to enforce by hand after `registry_review_status:
  "reviewed"`, `registry_confidence: "0.95"` and `registry_resolution_date` merged.)
- **Never gated:** serialization, whitespace, `"Infinity"` string form, companion
  pairing, path shape. No judgment involved; gating these is what produces approval
  fatigue.

This affects contributors — your submitted id won't be silently rewritten — so it wants a
line in CONTRIBUTING, not only in a skill.

**3. One validator check, and it isn't the one I first suggested.** I scanned all 13,228
records on datastore `main` before proposing anything, and it retired half of this item:

*Non-finite bounds — withdrawn, already covered.* `validator/json_utils.py::strict_json_loads`
already rejects bare `Infinity`/`-Infinity`/`NaN` (`parse_constant`), overflow literals like
`1e400` (`parse_float`), and duplicate keys, and `publish_evaluation_logs` writes with
`allow_nan=False`. On `main`, all 112,342 results carry finite numeric `min_score` and
`max_score` — zero nulls, zero absences, zero `"Infinity"` strings. What reached #164 came
from an adapter bypassing the publisher, and the gate caught it; that's why the fix commit
is called "strict JSON bounds". Nothing to add here.

*Path vs `model_info.id` — don't ship it as written.* Recomputing the publisher's own
derivation (`helpers.io.datastore_path_components`) and comparing it to the actual path
flags **584 of 13,228 records (4.4%)**: 238 case-only, 190 model-component, 102
developer-component, 54 both. Almost all of it is a deliberate slug convention in six
collections — `deepseek-ai`→`deepseek`, `meta-llama`→`meta`, `CohereLabs`→`cohere`,
`anthropic/Opus 4.1`→`anthropic/opus-4.1`. Those adapters are arguably right; you don't
want a directory named `Opus 4.1`. Since warnings block merge here, that check would
blockade 21 collections to catch about two real defects.

Two narrow variants have measured precision instead — **22 records on `main`, essentially
all genuine**:

- path model component equals the developer or the collection (12): `cohere/cohere` for id
  `cohere/command-nightly`, six `terminal-bench-2.0/multiple/multiple` placeholders,
  `vals-ai/aristotle/aristotle`.
- path model component is a strict prefix of the id's model part (10): `lmsys/vicuna-13b`
  for `vicuna-13b-v1.5`, `openai/o1` for `o1-2024-12-17`, `mistral/mistral-large` for
  `mistral-large-2411` — truncations that collapse distinct runs into one directory.

And the #151 defect itself is a **diff-level** invariant, not a record-level one: if
`model_info.id` changes in a PR, the file must move with it. That has zero false positives
by construction, and `REGISTERED_CHECKS` can't express it — it sees one file at a time — so
it belongs in the bot, which has the diff.

*Byproducts of the scan, worth fixing as data rather than code:* one off-layout record on
`main` (`data/reward-bench/my_model/6195e81a-….json`, four path components — the existing
path check would error on it); **282 records across 15 collections whose identity
`datastore_path_components` refuses**, so they cannot be republished through our own
publisher — 269 are `model_info.developer` missing or literally `unknown` (ids like
`unknown/azerogpt`), and the rest are two parse bugs: `Seed-OSS-36B-Base(w/o syn.)` split
on its slash into developer `Seed-OSS-36B-Base(w`, and six reward-bench ids carrying a
`.json`/`.jsonl` extension. Happy to open the checks PR and a data-cleanup PR separately.

**4. A scale threshold for the gate.** #164 is 189 model directories; #158 adds 987
collections; and the deployment keys are schema-required, so this fires on nearly every
submission rather than on an exceptional one. A several-hundred-row table handed to one
human is a rubber stamp, and "copy the proposal into every affected collection" is 987
copies. Above some N, approval has to be per-rule plus a sampled audit. Someone has to
pick N and accept sampled auditing.

**5. Cleanup of what already merged.** `data/l2-bench/deployment-metadata-proposal.md` is
on `main` reading `Status: **PENDING HUMAN APPROVAL**` — by design, since the PR says
never to alter the copy — and its rows key on `alibaba/alibaba/qwen3-32b`, which matches
no version of the data (merged said `Qwen/Qwen3-32B`, current says `alibaba/qwen3-32b`).
It also predates this template, so it has no digest suffix, scope manifest, or approval
section. Residual `registry_*` process keys are still in records. Removing both is a write
to merged data, so it's your call; worth telling the l2-bench contributor either way. Also
worth noting that deleting #188 erased that repair's audit trail.

## Suggestions on your design (yours to take or leave)

- **Approve claims, not rows.** What a human can actually review is one sentence:
  "`externally_managed` for all 9 on the strength of S1 (arXiv 2607.08842v2 §I.3: all
  models queried via Azure / Vertex / Bedrock)." One claim, one source, one approval, with
  the row table attached for spot-checking rather than as the object of approval.
- **Re-approve on claim/source change, not on head change.** An unrelated file landing in
  the PR shouldn't invalidate an evidence judgment that hasn't changed. With claim-level
  approval quoted in the discussion, the SHA-256 binding stops being necessary.
- **The confidence ladder can't rate its own worked example.** High = direct first-party
  evidence for the exact run; Medium = two aligned first-party sources for the exact
  variant; Low = one indirect source → never propose. The l2-bench proposal rates all nine
  models `D: medium` from a single source, reasoning that the records don't preserve the
  per-model endpoint. That's not High and not Medium, so by the letter it's Low and
  forbidden — while it shipped. The ladder is counting sources where it means proximity of
  evidence; it needs a rung for one authoritative methodology statement covering a defined
  set of runs. You know best what you meant by medium there.
- **Missingness: five statuses → three, one ladder not two, scoped trigger.**
  `record_absent` / `available_not_surfaced` / `unavailable_after_search` carry the whole
  distinction; the confidence ladder appears in both `metadata-missingness.md` and
  `model-deployment.md` with different wording, which is the drift AGENTS.md warns about.
  And "read this whenever a field is absent, null, defaulted" fires on dozens of optional
  fields in a leaderboard record — worth scoping to fields you're about to write or make a
  claim about.

## Notes, not blockers

- The aggregator rule is also a **producer** rule: a BenchPress-shaped adapter shouldn't
  emit a log-level deployment value it can't support. A one-liner in the conversion
  skill's `fields.md` would stop the reviewer having to catch it every refresh.
- Mandatory reading before an agent may touch a file is now ~62 KB / ~15k tokens (SKILL
  18.1 KB + two references + four sibling conversion references), up from ~45 KB, and the
  approval ceremony is longer than the content spot-check — which is where real defects
  hide, per "validating ≠ correct".
- The description should say what happened downstream of the one live application: the
  post-merge revert via a since-deleted #188, and #151's final two `/eee validate changed`
  requests with no bot verdict recorded at the merged head. That's the most
  decision-relevant evidence available and a reviewer can't see it.
- "emphasize that living first-party and community submissions can improve that coverage
  over time" reads as public-messaging guidance rather than review guidance. If it's
  deliberate, it probably belongs in a comms doc — flagging rather than deleting in case it
  came from an explicit request.
- Review lane: this mandates new files in the published dataset, so it's `Needs a human`
  rather than `Fast`, and the title wants the repo's `[Docs]`/`[Feature]` prefix.

<details>
<summary>Patch for the mechanical parts (applies on <code>5f861a59b</code>)</summary>

```diff
diff --git a/.agents/skills/eee-dataset-conversion/reference/datastore-gate.md b/.agents/skills/eee-dataset-conversion/reference/datastore-gate.md
index 7c76ab903..50527fff8 100644
--- a/.agents/skills/eee-dataset-conversion/reference/datastore-gate.md
+++ b/.agents/skills/eee-dataset-conversion/reference/datastore-gate.md
@@ -48,6 +48,8 @@ Re-derive this list from `REGISTERED_CHECKS`, `_DEPLOYMENT_TYPES`,
   This is where the percent-vs-proportion mismatch surfaces: declaring `0.0–1.0` while
   the source reports `73.4` is now a hard failure, not a silent unit bug. Declare the
   bounds the *source's* numbers live in, and convert deliberately if you rescale.
+  Widening the bounds to admit the number you have is the same defect as clamping the
+  number to fit the bounds: both make the record pass while changing what it claims.
 - `score_type: continuous` requires both bounds; a supplied-but-unparseable bound is
   also an error. `±inf` is accepted (serialized as the JSON strings
   `"Infinity"`/`"-Infinity"`); `null` is "not provided", never "unbounded".
diff --git a/.agents/skills/eee-datastore-pr-review/SKILL.md b/.agents/skills/eee-datastore-pr-review/SKILL.md
index d97779002..1018ebc60 100644
--- a/.agents/skills/eee-datastore-pr-review/SKILL.md
+++ b/.agents/skills/eee-datastore-pr-review/SKILL.md
@@ -69,36 +69,6 @@ Re-read the allowed deployment values from
 `every_eval_ever/validator/validation_core.py` and the live schema. Existing records
 and old bot comments may use obsolete vocabularies.
 
-## Progress checkpoints
-
-Emit an incremental checkpoint to the caller at every boundary below. Checkpoints are
-run receipts, not Hugging Face discussion comments: do not post them to the PR unless
-the operator explicitly asks. Each checkpoint must include the phase, current PR head
-SHA, facts established since the prior checkpoint, affected file/model counts, command
-exit statuses or evidence URLs when applicable, blockers, and the next action.
-When a progress or parent-message channel is available, send the checkpoint through it
-and continue in the same run. Do not end a turn merely to deliver a checkpoint.
-
-Required checkpoints:
-
-1. **Snapshot:** after selecting the PR head and matching bot run.
-2. **Diagnosis:** after reproducing the gate and grouping its findings.
-3. **Research proposal:** after resolving model-specific evidence. Render
-   `assets/deployment-metadata-proposal.md`, report its SHA-256, and pause for explicit
-   human approval before editing deployment fields or mutating the live PR.
-4. **Local repair:** after the repaired diff passes local validation and content review.
-5. **Remote receipt:** immediately after each uploaded commit or validator-trigger
-   comment, including the returned commit SHA or discussion event id.
-6. **Bot result:** after each completed bot run, tied to its head/fingerprint; repeat
-   diagnosis and repair checkpoints for another iteration.
-
-For a phase lasting more than 60 seconds, emit a heartbeat at least once per minute
-with the current evidence surface or bounded poll, completed/remaining counts, and
-whether local or remote state changed. Use bounded polling calls of at most 45 seconds
-so progress messages can be delivered. Continue after ordinary checkpoints. The
-research proposal is an approval gate; do not continue past it without an explicit
-approval matching both its digest and PR head.
-
 ## Workflow
 
 ### 1. Establish the exact PR state
@@ -115,9 +85,15 @@ approval matching both its digest and PR head.
    renamed, and deleted paths; include aggregate/instance companions even if only one
    side appears in the diff.
 
-Record the PR head commit and bot schema/compatibility version in the review notes.
-If the bot and local schema differ, label their disagreement as version skew and
-investigate it explicitly.
+Record the PR head commit and the bot run's schema/compatibility version and
+fingerprint in the review notes, and carry them into every finding you report: a
+finding without the head and fingerprint it came from cannot be re-checked. If the bot
+and local schema differ, label their disagreement as version skew and investigate it
+explicitly.
+
+If the PR is merged or closed while the review is in flight, stop pushing to its ref —
+commits there no longer reach `main`. Re-target the remaining repair at `main` in a new
+PR, cross-link it to the original discussion, and carry the same decision log across.
 
 ### 2. Reproduce the gate locally
 
@@ -157,11 +133,12 @@ Inspect content even when the validator omits it. At minimum check suspicious ze
 score scale and bounds, metric identity, `source_data`, duplicate overall/subtask
 aggregates, stable `evaluation_id`, model identity, answer leakage, and companion
 pairing. An out-of-range score requires finding the source scale or source value; do
-not cap, clamp, round it into validity, or widen the bounds around it. When a source
-publishes a normalized value, convert it to the metric's declared scale and preserve
-the raw value plus explicit conversion in `score_details.details` or the decision log.
-Serialize unbounded limits as the schema-supported JSON strings `"Infinity"` and
-`"-Infinity"`; replacing a bare non-finite token must not change its meaning.
+not cap, clamp, or round it into validity. When a source publishes a normalized value,
+convert it to the metric's declared scale and record the raw value and the conversion
+as strings in `score_details.details` (that object takes strings only) or in the
+decision log. Bound semantics and serialization — including what `null` means and how
+an unbounded limit is written — are in `datastore-gate.md` §score; read the rule there
+rather than re-deriving it.
 
 Inspect the raw JSON before constructing an `EvaluationLog`. The model layer may
 auto-fill absent deployment keys with `unknown`, hiding whether the contributor
@@ -228,8 +205,13 @@ again. Record the approver and approval time in the decision log.
 - When `model_info.additional_details` is absent or null, create the object only after
   researching both axes. When it already exists, merge the researched keys without
   discarding unrelated source metadata.
-- Keep `additional_details` values as strings. Add concise evidence/provenance there
-  when the source has no typed home and the decision would otherwise be opaque.
+- Keep `additional_details` values as strings. The field is deliberately open, so what
+  belongs there is defined by kind, not by an allowed key list: it carries what a source
+  states about this evaluation, attributed to that source. It never carries where the
+  review got to — no resolution status, confidence score, reviewer, or review date. Add
+  concise evidence/provenance there when the source has no typed home and the decision
+  would otherwise be opaque, and give the source URL and retrieval date rather than a
+  label like `S1` that only resolves inside a document stored elsewhere.
 - If generated records are wrong, fix or prepare the generating adapter in the code
   repo as well; otherwise the next refresh will restore the defect. Keep adapter code
   out of the datastore PR and cross-link its separate PR.
diff --git a/.agents/skills/eee-datastore-pr-review/reference/metadata-missingness.md b/.agents/skills/eee-datastore-pr-review/reference/metadata-missingness.md
index 867ef0af2..23d4d9570 100644
--- a/.agents/skills/eee-datastore-pr-review/reference/metadata-missingness.md
+++ b/.agents/skills/eee-datastore-pr-review/reference/metadata-missingness.md
@@ -14,8 +14,10 @@ adapter or submitter may have failed to surface.
 | `conflicting_sources` | Relevant primary sources disagree and run-level evidence does not resolve them | Document the conflict; use the field's unknown/absent representation only with approval |
 | `unavailable_after_search` | The stop rule is complete and searched first-party sources do not establish the value | Document the bounded claim and use the field's unknown/absent representation with approval |
 
-These are research statuses, not schema values. For deployment axes, the schema value
-corresponding to the last two unresolved outcomes is `unknown`.
+These are research statuses, not schema values, and they never appear inside a record:
+they belong in the review report, not in `additional_details` or any other field. For
+deployment axes, the schema value corresponding to the last two unresolved outcomes is
+`unknown`.
 
 ## Search primary surfaces
```

`git diff --check` clean; `uv run pytest tests/test_documentation_commands.py tests/test_skill_conversion.py` → 8 passed.

</details>

.agents/skills/eee-datastore-pr-review/assets/collection-review-decision.md
```diff
@@ -0,0 +1,18 @@
+## `<YYYY-MM-DD>` — `<short decision title>`
+
+- PR: `<discussion URL>`
+- Pre-edit head: `<40-character SHA>`
+- Scope: `<record/result/model counts and repo-relative paths or compact groups>`
+- Approved proposal: `<relative path or n/a>`
+- Proposal SHA-256: `<full digest or n/a>`
+- Approval: `<approver, timestamp, and exact approved head or n/a>`
+- Local verification: `<validator and duplicate-check receipts>`
+- Matching bot receipt: `<event/run/fingerprint or pending>`
```
mrshu, line 10: **RAV-RUN1-R1-F001 [medium] audit-closure** — The append-only record can remain at `pending` after validation completes. This entry must exist before upload, while the matching bot run exists only afterward, and the later workflow does not define a closure entry. A receipt-only append linked to this decision and the validated data head would preserve both completeness and append-only history.

.agents/skills/eee-datastore-pr-review/assets/collection-review-decision.md
```diff
@@ -0,0 +1,18 @@
...
+
+### Decisions
+
+- `<field or issue>`: `<source-backed old → new value and concise reason>` (`<source ids>`)
```
mrshu, line 14: **RAV-RUN1-R1-F002 [medium] evidence-resolution** — These source IDs have no resolver when `Approved proposal` is `n/a`, and the approval field above has no durable event URL or receipt ID. A compact per-entry Sources section plus an approval-receipt field would let future readers trace both the evidence and authorization.

.agents/skills/eee-datastore-pr-review/reference/metadata-missingness.md
```diff
@@ -0,0 +1,96 @@
...
+It applies to deployment metadata, temperature, token limits, and other fields that an
+adapter or submitter may have failed to surface.
+
+## Classify the observation
+
+| Status | Meaning | Required action |
+|---|---|---|
+| `record_absent` | The submitted record omits the field; source research has not finished | Say "not surfaced in the submitted record" and investigate |
+| `research_incomplete` | One or more relevant primary surfaces or exact identities remain unchecked | Continue research; do not propose `unknown` or claim source missingness |
+| `available_not_surfaced` | A reliable source contains the value but the record does not | Backfill with approval and fix/follow up on the adapter or submission path |
```
mrshu, line 13: **RAV-RUN1-R1-F003 [medium] authorization-scope** — This approval requirement also applies to temperature, token limits, and other non-deployment fields, but the skill defines a head-bound artifact only for deployment axes. Stating whether ordinary fix authorization or a lightweight head-bound receipt governs these backfills would keep the mutation boundary unambiguous.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The new approval and evidence controls are a strong direction; three medium workflow edges would benefit from clarification before agents rely on the durable receipts.
**Result:** 3 medium findings, 1 low suggestion, and 1 item set aside after its verification protocol did not complete.
**Scope:** Five datastore-review skill and template files at `5f861a59bc33690e0640fcd538fe031b3c6afcf5` (+364/−15).
**Checks:** Three Codex reviewers examined the exact head. The three actionable concerns had reviewer consensus and did not require reproduction; the single-reviewer scope concern was set aside after two protocol-invalid verifier results.
**Second check:** Two targeted Codex adversaries challenged approval and audit lifecycle behavior. They reinforced the reported concerns; adversarial-only ideas were not promoted into findings.

## Findings

### **RAV-RUN1-R1-F001 [medium] audit-closure** — the append-only record has no defined way to close a pending bot receipt

The decision entry is created before upload, while the matching bot run exists only afterward. The template permits `pending`, but the later monitoring steps do not require an append-only closure receipt. A small receipt-only entry linked to the original decision and validated data head would keep the durable record complete without rewriting history.

### **RAV-RUN1-R1-F002 [medium] evidence-resolution** — durable decisions can contain source and approval references that cannot be resolved

Decision rows cite source IDs, but the template has no Sources section when the deployment proposal is `n/a`. The approval field also records a person, time, and head without a stable event URL or receipt ID. A compact Sources section and approval-receipt field would make these records independently traceable.

### **RAV-RUN1-R1-F003 [medium] authorization-scope** — approval is undefined for non-deployment metadata backfills

The missingness protocol covers temperature, token limits, and similar fields and requires approval for `available_not_surfaced`. The concrete digest-and-head approval flow covers only deployment axes, so an agent cannot tell whether ordinary fix authorization or a new artifact governs other backfills. Defining that boundary would preserve the PR's careful mutation controls.

<details>
<summary>1 low-priority suggestion</summary>

- **RAV-RUN1-R1-F005 [low] evidence-threshold** — The proposal row permits categorical values beside `low` confidence even though the references allow categorical proposals only at high or medium confidence. Encoding that condition in the template would make the safe combination easier to follow.

</details>

- **RAV-RUN1-R1-F004 [medium] approval-scope:** One reviewer noted that ordinary proposal rows are not mapped to exact paths or evaluation IDs. We set this aside because both verifier attempts returned protocol-invalid status, so the required independent check did not complete.

<details>

- Target: PR #241, head `5f861a59bc33690e0640fcd538fe031b3c6afcf5`
- Rounds: 1/1 completed; adaptive continuation not needed
- Mix: 3 `codex-exec` reviewers; Codex-only verification and adversarial checks
- Adversarial mode: auto, targeted
- Approval policy: never; comment-only
- Reviewer timeout: 1,200 seconds
- Focus: constructive, positive review language and optional suggestions while preserving technical accuracy
- Earlier review-anvil history: none detected by the helper

</details>

---

.agents/skills/eee-datastore-pr-review/reference/model-deployment.md
```diff
@@ -63,6 +118,16 @@ Use these as evidence tests, not name-based mappings:
 Do not mechanically translate stale vocabulary until checking the old validator's
 meaning and the submitted run. Lexical similarity is not source evidence.
 
+For `deployment_type`, require run/evaluator evidence: raw generation metadata, a
+client/backend in the pinned adapter or config, or an explicit methodology statement.
+A model card cannot establish who operated inference for this evaluation.
+
+For `model_availability`, inspect the exact artifact, not only a family README. A
+downloadable gated or restrictively licensed checkpoint is `open_weights` under the
+current schema. API access alone is `closed_weights`. Treat delta-only releases,
```
j-chim, line 127: How are we defining API access here? The current phrasing seems to point to hosted open weight models as closed weight. It seems this is a registry-level decision and registry would classify this as open weight and make the API distinction in deployment_type?

.agents/skills/eee-datastore-pr-review/SKILL.md
```diff
@@ -103,34 +147,81 @@ Group findings by root cause rather than by file. For each group, record:
 - the source evidence needed for a correct fix;
 - proposed change and confidence.
 
+Classify every apparent omission with `reference/metadata-missingness.md`. Do not call
+an absent field genuinely missing while its status is `record_absent` or
+`research_incomplete`. If a README, eval card, methodology page, leaderboard, paper,
+repository, or API exposes it, classify it as `available_not_surfaced` and identify
+the adapter/submission gap.
```
j-chim, line 154: Nit: do we mean model card or similar artifacts when we say "eval card" in these files

.agents/skills/eee-datastore-pr-review/reference/metadata-missingness.md
```diff
@@ -0,0 +1,96 @@
...
+4. Evaluation paper, appendix, supplementary material, source repository, and pinned
+   configuration files.
+5. Exact model card, release, weight-file listing, license/access page, and aliases.
+6. Official provider API/SDK documentation, model catalog, release/deprecation notice,
+   and an archived page when current documentation may differ from the evaluated run.
+
+Follow every concrete primary-source lead found on those surfaces. Do not count a
+search-result snippet, model family, provider name, or current library default as the
+submitted run's value. A harness default is evidence only when the pinned version and
+run configuration establish that the default governed this run.
```
j-chim, line 39: I think it would be helpful to operationalize this further. Terms such as “exact model,” “concrete lead,” and “the default governed this run” leave substantial judgment to the agent. Could we add a few qualifying and non-qualifying examples?

For instance:
```
- Qualifies:
  - The source leaderboard, methodology, or result page directly associates the value with the submitted result.
  - The evaluation README or paper says that all results in a defined release, table, or run group used the value.
  - A run-specific source says that a named, pinned library’s defaults were used, and that version documents the value.
  - A configuration or run script tied to the submitted results sets the value, with no submitted configuration showing an override.
- Does not qualify:
  - A snippet from a search engine or tool; open and inspect the underlying source instead.
  - A generic model-family, provider, or library default without evidence that it applied to this evaluation.
  - A current library default when the run’s version is unknown.
  - An example configuration or run script that is not tied to the submitted results.
```

.agents/skills/eee-datastore-pr-review/reference/metadata-missingness.md
```diff
@@ -0,0 +1,96 @@
+# Distinguish record omissions from unavailable metadata
+
+Use this protocol before claiming that any reproducibility or model field is missing.
+It applies to deployment metadata, temperature, token limits, and other fields that an
+adapter or submitter may have failed to surface.
```
j-chim, line 5: Consider clarifying with:

```Evidence should also be assessed per field: a source may establish one missing value without establishing every missing field in the record.```

j-chim (changes requested):

PR #244, opened 2026-08-10
Test that a converted record actually passes the merge gate

borgr:
> Base of a stack: #245 (naming) and #246 (metric config) build on this branch, and their `ConverterCase` declarations extend the harness added here. This PR stands alone.

## What / source

Adapters under `every_eval_ever/adapters/` are pinned by tests on their output. The converters under `every_eval_ever/converters/` are not: their tests assert fields on in-memory `EvaluationLog` objects, and the `validate_file()` they call runs with the **semantic checks off**. So nothing in the repo showed that a converted record would pass the gate the datastore applies.

This adds one parametrized module that answers the question a user actually has — *someone ran an evaluation; can this repo still turn that output into a record the datastore accepts?*

Per converter it is a declaration, not a test file:

```python
ConverterCase(
    source='lm_eval',
    log_path=REPO_ROOT / 'tests/data/lm_eval/results_2026-01-21T03-44-18.458309.json',
    aggregates=2,
    model_id='RylanSchaeffer/mem_Qwen3-93M_...',
    required_source_paths=('config.model', 'results.*', 'configs.*.dataset_path'),
)
```

`tests/test_converter_conversion.py` then runs the **real CLI** — the same `convert` a user runs, publishing to the canonical `data/<collection>/<developer>/<model>/` path where the semantic checks fire — and requires `validate` to report **no errors and no warnings**, since a warning here is one every user of that converter inherits.

Three converters are declared: `lm_eval`, `inspect`, `helm`. A case whose converter is behind an optional extra reads that converter's own `_..._IMPORT_ERROR` and skips, so declaring one costs nothing in the core install.

`required_source_paths` names the upstream keys the converter cannot work without, asserted against the fixture. A fixture refreshed from a newer upstream release then fails by name — *"`config.model` is gone"* — rather than as a stack trace from inside the converter. There is deliberately no exhaustiveness rule: a *new* upstream key is not a break and should not turn CI red.

## What it does not check, stated in the docstrings

- **Shape, not semantics.** A metric that changes from a percentage to a proportion upstream passes this.
- **Not a new upstream version.** The fixture is a committed log; a release that renames a field is invisible here. That is what a `tools/upstream_smoke/` script is for — running the upstream tool's own fake-model mode (`--model dummy` for lm-eval, `--model mockllm/model` for Inspect) on a schedule. Not in this PR.

## Why it is shaped this way

No frozen bytes and no golden trees: the assertion is "passes the gate", not "matches these bytes", so nothing needs regenerating when the schema or the publisher changes. And a new converter costs one committed log plus one `ConverterCase` to be covered — there is no test file to write.

## Review lane

- [x] **Fast** — tests only, plus the README section that tells a contributor how to use them. No converter behaviour changes.
- [ ] Needs a human

## Checklist

- [x] `uv run python -m every_eval_ever validate <files>` clean, **no warnings** — that is what the new test asserts, at the final `data/<collection>/<dev>/<model>/` path
- [x] offline unit test added + full `uv run pytest tests` green — core 414 passed / 26 skipped, full 475 passed / 1 skipped

## Decisions & coverage

- Decision / where: run the real `cli.main(['convert', ...])` and the real `validate.main([...])`, in `tests/converter_cases.py`.
  Chose / instead of: calling the adapter directly and asserting on the returned object, which is what the existing converter tests do and what leaves the semantic checks off.
  Confidence: high — the canonical publish path is a precondition of those checks, so nothing short of the CLI exercises them.
  General? yes — this is the harness any future converter should be covered by.
- Decision / where: `required_source_paths` is opt-in and non-exhaustive.
  Chose / instead of: asserting the full set of keys the fixture has.
  Confidence: high — an added upstream key is not a regression, and a test that reddens on one would be turned off.
  General? yes.

**Coverage:** 3 of 4 in-tree converters declared. `alpaca_eval` is not: it fetches a leaderboard CSV over the network rather than reading a local log, so it has no fixture to point at and is a source adapter in converter clothing.


tests/test_converter_conversion.py
```diff
@@ -0,0 +1,114 @@
...
+    if case.model_id is not None:
+        assert {log['model_info']['id'] for log in logs} == {case.model_id}
+    if case.results is not None:
+        converted = sum(len(log['evaluation_results']) for log in logs)
+        assert converted == case.results, (
+            f'{case.source} converted {converted} result(s), '
+            f'expected {case.results}'
+        )
+    if case.scores is not None:
+        converted = {
```
mrshu: **RAV-RUN1-R1-F001 [medium] result assertions** — This dictionary overwrites duplicate evaluation/metric keys. Since the lm-eval case leaves `results` unset, an extra identical result can still pass the score mapping and semantic gate. An explicit `results=2` plus a uniqueness check before dictionary construction would preserve this useful declarative shape while catching duplicates.

tests/converter_cases.py
```diff
@@ -0,0 +1,208 @@
...
+
+    A converter behind an optional extra states its own missing dependency in an
+    `_..._IMPORT_ERROR` module global and raises from it when used. Reading that is what
+    lets a case be declared once and skip, rather than fail, in the core install.
+    """
+    module = importlib.import_module(
+        f'every_eval_ever.converters.{case.source}.adapter'
+    )
+    for name, value in vars(module).items():
+        if name.endswith('_IMPORT_ERROR') and value is not None:
```
mrshu, line 127: **RAV-RUN1-R1-F002 [medium] optional dependencies** — Every stored import error becomes an unavailable skip here. Because Inspect and HELM capture broad exceptions, an installed but incompatible release can be skipped even in full CI. Declaring each optional distribution explicitly would let core skip a genuinely absent package while surfacing an installed package's captured import error.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The harness is well focused and all CI matrices
are green; two coverage blind spots and one small documentation correction remain
as constructive suggestions.
**Result:** The public CLI, final datastore paths, sidecars, semantic gate,
warnings, and expected output content now receive valuable shared coverage.
**Scope:** Adds a declarative merge-gate harness for the lm-eval, Inspect, and
HELM converters, plus contributor guidance for extending it.
**Checks:** 3 concerns checked; 1 confirmed, 1 lowered in priority, and 1 set
aside after its current fixture coverage was verified.
**Second check:** targeted, 2 Codex reviewers; 3 findings kept, 2 fix paths
clarified, 1 set aside, 1 removed.

## Earlier review comments

None.

## What I noticed

|---|---|---|---|---|
| RAV-RUN1-R1-F001 | medium | result assertions | `tests/test_converter_conversion.py:79` | The score check builds a dictionary, so duplicate result keys overwrite one another. Because the lm-eval case leaves `results` unset, an extra identical result can pass counts, scores, and the semantic gate. (inline) |
| RAV-RUN1-R1-F002 | medium | optional dependencies | `tests/converter_cases.py:123` | `unavailable()` skips every stored import error. Inspect and HELM capture broad exceptions, so an installed but incompatible dependency can become a skip even in the full matrix intended to exercise compatibility. (inline) |

<details>

- **RAV-RUN1-R1-F004 [low] coverage documentation** — The test prose and
  converter README describe `tools/upstream_smoke` and scheduled execution as
  existing coverage, but this snapshot contains neither the script nor a
  scheduled workflow.

</details>

## Things to try

- **[medium] result assertions** — Set `results=2` for the lm-eval case and
  reject duplicate extracted keys before comparing the existing score mapping.
  This keeps the check order-independent while preserving duplicates.
  (`RAV-RUN1-R1-P001`; covers `RAV-RUN1-R1-F001`)
- **[medium] optional dependencies** — Declare the optional distribution for
  each applicable case, skip only when package metadata confirms it is absent,
  and surface the captured import exception when an installed distribution is
  broken. (`RAV-RUN1-R1-P002`; covers `RAV-RUN1-R1-F002`)
- **[low] coverage documentation** — Keep both prose locations factual about
  committed-fixture limits; describe scheduled upstream smoke testing as future
  work, or link a tracked follow-up, until that layer exists.
  (`RAV-RUN1-R1-P004`; covers `RAV-RUN1-R1-F004`)

- **RAV-RUN1-R1-F003 [low] wildcard semantics** — set aside because `*` is
  explicitly existential, no current case requires all-branch behavior, and the
  concrete lm-eval fixture is already protected by its adapter test.

<details>

- Target: PR #244 at `1267c9270c9a98ea29990b0eb69fca792dce6793`
  (4 files, +342/-1)
- Run ordinal: 1
- Mix: 3 Codex reviewers
- Focus: semantic-gate correctness, optional-dependency behavior, failure
  visibility, future harness ergonomics, and positive suggestion-oriented
  language
- Finding counts: 0 critical, 0 high, 2 medium, 1 low, 0 nit
- Verification: exact-head core run — 3 passed, 6 optional-extra skips; GitHub
  Actions passed all four core/full locked/loose matrices
- Checks: concerns=3; confirmed=1/set-aside=1/lowered=1/ruled-out=0
- Second check: targeted; reviewers=2; kept=3/clarified=2/set-aside=1/removed=1;
  approval changed no
- Set aside: 1 item; reason=no present all-branches consumer

</details>

---

borgr:
**RAV-RUN1-R1-F002 — `unavailable()` can turn a broken-but-installed optional dep into a skip:** real, but out of scope for this PR, so I've left it for the import-health work rather than reworking it here.

This PR's job is to prove a *converted record* clears the merge gate — the result-assertion / join-key correctness in `test_converter_conversion.py` (that's F001, now fixed). `unavailable()` in `converter_cases.py` answers a different question: whether an optional converter's dependencies import at all. Teaching the shared skip helper to tell "absent" from "installed but broken" changes what the *entire* case matrix skips on, well beyond the record-gate question this PR exists to answer.

That distinction is exactly what #243 (`fix/helm-extra-imports`) is for: it already adds a sentinel, `test_helm_extra_is_importable_when_installed`, that *fails* rather than skips when HELM is installed but its guarded imports don't resolve. The right fix for F002 is to extend that same pattern to `inspect` and `lm_eval` there — skip only when `importlib.util.find_spec` confirms the package is absent, and surface the captured import error when a distribution is present but broken — not to overload `unavailable()` in this PR. Tracking it as follow-up on #243.

borgr:
Reviewed the 2 unresolved automated review threads against the current head. Both dismissed: the lm-eval `ConverterCase` already sets `results=2` with the exact rationale the first thread asks for, and the "installed-but-incompatible" skip distinction (`converter_cases.py:127`) is out of scope for a merge-gate test and already covered by the dedicated `test_helm_extra_is_importable_when_installed` sentinel. Resolving both.

PR #245, opened 2026-08-10
Name the benchmark, the metric and the result as three separate things

borgr:
> Stacked on #244 (`tests/converter-merge-gate`). **Review that one first** — its `ConverterCase` harness is what pins the output below. Diff against that branch, not main.

## What / source

Three fields, three different questions: `evaluation_name` is *what was evaluated*, `metric_config.metric_name` is *what was measured*, `evaluation_result_id` is *which result this is*. The HELM and Inspect converters were putting a generated English sentence in the first, nothing in the second, and (lm_eval) nothing in the third.

```
main (before)                                             →  this branch (after)

lm_eval  name='math_perturbed_full'                          name='math_perturbed_full'
         rid=None      metric_name=None                      rid='exact_match'  metric_name='exact_match'
         evaluation_description='exact_match'

inspect  name='accuracy on inspect_evals/cyse2_vulnerability_exploit for scorer vul_exploit_scorer'
         rid=None      metric_name=None
      →  name='inspect_evals/cyse2_vulnerability_exploit'
         rid='vul_exploit_scorer:accuracy'  metric_name='accuracy'
         evaluation_description='accuracy from scorer vul_exploit_scorer'

helm     name="exact_match valid PerturbationDescription(name='robustness', robustness=True,
               fairness=False, computed_on='worst', seed=None) on hellaswag"
         rid='exact_match:valid:robustness'  metric_name=None
      →  name='hellaswag'  rid='exact_match:valid:robustness'  metric_name='exact_match'
```

Why each of those matters:

- **`evaluation_name` is the field a registry lookup resolves against `entity_type: benchmark`.** Under the old naming, every result of one HELM run got a *different* `evaluation_name`, so nothing could group a model's records by benchmark — and the instance-level sidecar already wrote `hellaswag`, so the aggregate and its own companion file disagreed about what was evaluated.
- **Nothing is lost by dropping the sentence.** The metric, split and perturbation are exactly what `evaluation_result_id` is built from (`exact_match:valid:robustness`), which is also the sidecar's join key, and they are in `score_details.details`.
- **`evaluation_result_id` has to be unique within an evaluation, and a metric name is not.** Two Inspect scorers can both report `accuracy`; lm-eval reports each metric once per filter; HELM reports each metric on `valid` and again worst-case per perturbation family. So the id is `<scorer>:<metric>` (inspect), `<metric>[:<filter>]` (lm_eval), `<metric>:<split>[:<perturbation>]` (helm).
- **Instance rows now fan out one per aggregate result they contributed to**, as the instance schema requires, each carrying its own `evaluation_result_id` and its own scorer's score. Previously one row per sample carried one scorer's grade regardless of how many results the sample fed, so a sidecar row could not be attributed to the score it belonged to.

Two smaller fixes ride along, both in this same naming path:

- HELM's `score_details.details.perturbation` was a `repr()` of HELM's dataclass — `PerturbationDescription(name='robustness', robustness=True, …)`. It now goes through `_stat_name_part`, the same helper the id already used, and reads `robustness`.
- Inspect's `--supplemental_eval_details` could only key an `evaluation_results` entry on `evaluation_name`, which is now shared by every result of a task. Entries can key on `evaluation_result_id` (one result) or `evaluation_name` (all results of an evaluation), `evaluation_result_id` wins where both match, and **a key that selects nothing is logged instead of silently ignored** — the old behaviour, where a typo'd key looked like a working supplement.

`metric_id` is deliberately *not* set here; that is the next PR in the stack.

## Review lane

- [ ] Fast
- [x] **Needs a human** — a material change in outcome: the same input now produces different `evaluation_name`, `metric_name` and `evaluation_result_id` values, and a different number of sidecar rows. Records already published under the old naming will not match new ones on `evaluation_name`.

Design agreed in: not yet — this is the PR that raises it. The schema's own field descriptions are the argument (`evaluation_name` = the eval, `metric_name` = the metric), and `AGENTS.md` already lists "`metric_name` is a metric" as a spot-check, so I do not think the *direction* is in question. What needs a maintainer is the **migration**: whether previously-published HELM/Inspect records in `EEE_datastore` get reconverted, and whether `evaluation_result_id` values are considered stable identifiers that this changes.

## Checklist

- [x] `uv run python -m every_eval_ever validate <files>` clean, **no warnings**, at the final `data/<collection>/<dev>/<model>/` path — asserted by the harness this stacks on, over all three converters
- [x] every unconvertible source row is in `adapter_reports/` with a non-zero exit — unchanged by this PR
- [x] offline unit test added + full `uv run pytest tests` green — core 416 passed / 28 skipped, full 485 passed / 1 skipped
- [ ] **model/benchmark ids** — `evaluation_name` is now the upstream benchmark id (`hellaswag`, `inspect_evals/cyse2_vulnerability_exploit`), which is what a registry lookup needs, but this PR does not claim those all resolve. Metric ids are the next PR.
- [x] content spot-checked — scores and sample counts are byte-identical before/after; only names and ids move. `evaluation_id` is unaffected.

The new `ConverterCase` fields are the enforcement, so this is not prose you have to trust:

- `metric_names` pins the distinct `metric_config.metric_name` per converter — a metric that goes back to being unnamed, or lands in `evaluation_name`, fails by name.
- `scores` is keyed on `evaluation_name` + `evaluation_result_id`, which is only unique *because* of this change.
- `sidecar_rows` pins the fan-out (inspect 1 sample × 3 results = 3 rows; helm 10 instances × 8 `valid` results = 80) and that `total_rows` matches the rows actually written.

## Decisions & coverage

- Decision / where: `evaluation_result_id` = `<scorer>:<metric>` for inspect, `<metric>[:<filter>]` for lm_eval, in each converter's `utils.py`.
  Chose / instead of: the metric name alone, which is shorter and reads better.
  Confidence: high — it is not unique. Two Inspect scorers reporting `accuracy`, or one lm-eval metric under two filters, would collide, and the sidecar join key would become ambiguous.
  General? yes — any converter whose harness reports a metric more than once per evaluation needs a qualified id.
- Decision / where: the scorer goes to `evaluation_description` rather than into `evaluation_name`, in `converters/inspect/utils.py`.
  Chose / instead of: `evaluation_name = '<task> - <scorer>'`, which keeps every result distinctly named.
  Confidence: high — that is what made the old naming unusable; the scorer is a property of the measurement, not of the benchmark.
  General? yes.
- Decision / where: a `supplemental_eval_details` key that matches no result is logged as a warning, not an error, in `converters/inspect/supplemental_eval_details.py`.
  Chose / instead of: failing the conversion.
  Confidence: medium — a supplement written against one log and reused across a directory of logs will legitimately not match everything. A warning surfaces the typo case without breaking the batch case; if maintainers would rather this be fatal, it is a one-line change.
  General? yes.
- Decision / where: `metric_id` left unset in this PR.
  Chose / instead of: resolving ids here, since this is already the metric-naming PR.
  Confidence: high — id resolution needs a registry revision pin, a hand-verified map and a checker; that is its own reviewable unit, stacked next.
  General? no.

**Coverage:** same records, same scores, same values — 2 lm_eval results, 3 inspect, 24 helm, as before. Nothing dropped. What changes is the naming of those records and the instance-row fan-out (inspect 1 → 3 rows, helm 10 → 80).

**Operator asked about policy calls?** Yes, one, and it is the reason for the needs-a-human label: this renames fields in already-published data. Whether `EEE_datastore` records are reconverted, and whether `evaluation_result_id` is treated as a stable external identifier, is a maintainer call, not mine.


every_eval_ever/converters/lm_eval/instance_level_adapter.py
```diff
@@ -137,21 +145,13 @@ def _transform_sample(
...
-                    break
-
-        if score is None:
-            score = 0.0
-            is_correct = False
+        scored = [
+            (metric_name, float(sample[metric_name]))
+            for metric_name in metrics
+            if isinstance(sample.get(metric_name), (int, float))
+            and not isinstance(sample.get(metric_name), bool)
```
mrshu, line 153: **RAV-RUN1-R1-F001 [medium] — suggestion:** It may be helpful to keep boolean metric contributions in this fan-out. The aggregate path accepts `bool` through Python's numeric hierarchy, so excluding it here turns a valid `true` contribution into an unjoined synthetic `0.0` row. Normalizing booleans to `1.0`/`0.0` would preserve the new result-ID join while leaving the fallback for genuinely non-numeric values.

every_eval_ever/converters/inspect/utils.py
```diff
@@ -508,26 +573,40 @@ def apply_supplemental_eval_details(
         )
 
     result_supplements = supplemental_eval_details.evaluation_results or []
-    named_supplements = {
-        supplement.evaluation_name: supplement
+    by_result_id = _key_supplements(result_supplements, 'evaluation_result_id')
+    by_evaluation_name = _key_supplements(result_supplements, 'evaluation_name')
```
mrshu, line 591: **RAV-RUN1-R1-F002 [medium] — suggestion:** Could entries with an `evaluation_result_id` be excluded from this name index (or could the model reject both selectors together)? As written, one dual-key entry first matches its specific result by ID, then matches sibling results through the shared evaluation name. A small two-result test would make the intended precedence especially clear.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This is a thoughtful separation of benchmark,
metric, and result identity, and the exact-head tests plus all four CI matrices
are green. Two edge cases look worth tightening as constructive follow-ups.
**Result:** Aggregate results and instance rows now have substantially clearer,
joinable identities across lm-eval, Inspect, and HELM.
**Scope:** Renames result fields, adds metric-aware instance fan-out, and extends
Inspect supplemental-result selectors.
**Checks:** 6 concerns checked; 2 newly introduced issues confirmed and 4
pre-existing or out-of-scope concerns set aside.
**Second check:** targeted, 2 Codex reviewers; both kept the same 2 actionable
findings and bounded the remaining candidates by base/head provenance.

## Earlier review comments

None.

## What I noticed

|---|---|---|---|---|
| RAV-RUN1-R1-F001 | medium | lm-eval boolean metrics | `every_eval_ever/converters/lm_eval/instance_level_adapter.py:153` | The new fan-out excludes boolean sample metrics even though the aggregate converter accepts them as numeric. A boolean-only metric therefore produces an aggregate result but falls back to an unjoined sidecar row with score `0.0`. (inline) |
| RAV-RUN1-R1-F002 | medium | Inspect supplemental selectors | `every_eval_ever/converters/inspect/utils.py:577` | A supplement containing both `evaluation_result_id` and `evaluation_name` is indexed under both keys. It can first select one result by ID and then also apply to every sibling with the same evaluation name. (inline) |

## Things to try

- **[medium] lm-eval boolean metrics** — Normalize boolean contributions to
  `1.0` or `0.0` and emit their regular metric/filter result ID, reserving the
  unjoined fallback for values that truly cannot be represented as scores.
  (`RAV-RUN1-R1-P001`; covers `RAV-RUN1-R1-F001`)
- **[medium] Inspect supplemental selectors** — Make the two selectors mutually
  exclusive during validation, or keep entries with an
  `evaluation_result_id` out of the evaluation-name index. A two-result
  regression test would nicely pin the chosen precedence rule.
  (`RAV-RUN1-R1-P002`; covers `RAV-RUN1-R1-F002`)

- The lm-eval filter suffix in `evaluation_name`, Inspect's scorer-answer use
  in `output.raw`, and its positive-score `is_correct` rule all reproduce in
  the stacked base. The new fan-out makes the latter two more visible, but this
  review does not present them as regressions introduced here.
- HELM's answer-attribution method is outside this PR: the relevant instance
  adapter has no base-to-head diff.

<details>

- Target: PR #245 at `c46eec389e3ddb15f1eef37728980f32d35ab0da`
  (15 files, +919/-239)
- Run ordinal: 1
- Mix: 3 Codex reviewers
- Focus: converter identity semantics, instance-to-aggregate joins,
  supplemental selector behavior, and positive suggestion-oriented language
- Verification: exact-head focused converter run — 32 passed, 14 optional-extra
  skips; GitHub Actions passed all four core/full locked/loose matrices
- Checks: concerns=6; confirmed=2/set-aside=4/ruled-out=0
- Second check: targeted; reviewers=2; kept=2/bounded=4; approval changed no

</details>

---

borgr:
Both findings addressed — one implemented, one left as is with reasoning.

**RAV-RUN1-R1-F002 — supplement entry indexed under both keys.** Fixed (`f1b672358`). An entry that set both `evaluation_result_id` and `evaluation_name` landed in both selection indices, so it applied to its own result by id and, through the shared name, to that result's siblings — a supplement scoped to one result silently fanned out over the whole evaluation. That single entry is now rejected with an error that names the ambiguity, rather than resolved to a surprising union; both behaviours stay available as two separate entries (a name entry for all, an id entry to override one). A two-result regression test pins the rejection.

**RAV-RUN1-R1-F001 — boolean sample metrics excluded from the fan-out.** Left as is for now. The exclusion is intentional and pinned by `test_instance_level_non_numeric_metrics_only`: a sample whose declared metrics are `verdict: 'correct'` and `passed: True` yields a single sidecar row with `evaluation_result_id=None`, and both raw values stay recoverable from `metadata['lm_eval_metrics']`. Normalizing booleans to `1.0`/`0.0` and emitting a joinable id would reverse that test.

Your asymmetry point is fair — the aggregate converter does accept booleans as numeric. We're keeping the instance side conservative on purpose: it mints a joinable score only for values already numeric on the sample, and preserves everything else verbatim in metadata rather than deciding that a boolean flag named in `metrics` is the aggregated quantity. If you have a concrete case where the boolean metric has a real aggregate result and the null-id sidecar row blocks a join someone needs, that's the signal to change it — happy to take it up then.

borgr:
Reviewed the 2 unresolved automated review threads against the current head. The dual-selector concern (`inspect/utils.py:591`) is already handled — a `both_selectors` guard raises when an entry sets both, so resolving that one.

Leaving open as an author decision: booleans are excluded from the instance-level fan-out (`instance_level_adapter.py:153`) while the aggregate path accepts them via the numeric hierarchy, so a per-sample `true` becomes an unjoined synthetic `0.0` row in the result-id join this PR adds. Whether to normalize bool→1.0/0.0 there is a design call; I'll settle it separately.

every_eval_ever/converters/lm_eval/instance_level_adapter.py
borgr: Fixed in 5990328e6. Dropped the bool exclusion in the instance-level converter: `float(True)`/`float(False)` canonicalizes a per-sample bool to 1.0/0.0, matching what the aggregate path already does (Python bool is an int), so the sample emits a real joined row instead of the synthetic `(None, 0.0)` fallback. On your second point — the "canonicalize, don't invent" principle was implicit (in test names and the coverage-vs-fill rule) but not stated as a general rule, so I added it to the conversion skill (SKILL.md, "The one rule: canonicalize, never invent"): reshaping a value the source carries is expected; manufacturing one it lacks is not.

PR #246, opened 2026-08-10
Report metric bounds, uncertainty and ids the way each harness computed them

borgr:
> Stacked on #245 (`fix/converter-metric-naming-3fields`), which is itself stacked on #244. **Review those first**; diff against the naming branch, not main.

## What / source

The three harness converters were each *asserting* things about their metrics that the upstream logs never say. This PR makes each converter report what its harness actually computed, and say "unknown" where the harness does not know.

Four groups of change, all funnelling through one new module (`converters/common/metrics.py`) so a rule is stated once for every converter:

**1. Bounds and direction, only where they are known.** Every HELM and Inspect metric was published as `continuous` on `[0, 1]` with `lower_is_better: false`, so a perplexity, a standard deviation and an accuracy all came out looking like accuracies. Now a metric's range comes from its *name*, layered per harness — lm-eval's `bleu` is sacrebleu's 0–100 while HELM's `bleu_1` is nltk's `sentence_bleu` 0–1, so a bare name cannot carry a range. A metric in no table gets **no bounds** and an `additional_details.bounds_status: unknown` marker, because `min_score`/`max_score` are nullable and "not provided" is true, while `[0, 1]` on an unbounded metric is not. Each record counts its unknown-bounds metrics in `source_metadata.additional_details`, so the gap is visible without reopening every file — and `test_unknown_bounds_are_counted_in_the_source_metadata` derives that count from the published results for every case, which is how the Inspect converter was caught reporting `null` while publishing an unbounded metric.

Dispersion metrics (`std`, `stderr`, `bootstrap_stderr`, `var`) get `polarity: not_applicable`: `lower_is_better` is a required boolean with no "does not apply" value, and a standard deviation is not better when small.

One of the ranges this PR introduced was itself wrong and is fixed here: `brier_score` was declared `[0, 1]`, the two-class definition. lm-eval computes `mean(sum((softmax(loglikelihoods) - one_hot(gold)) ** 2))` (`api/task.py`), which over `n` classes reaches **2.0** when all the mass sits on one wrong class. Since `validation_core.py` warns on a score outside its declared range and the converter suite requires zero warnings, the too-narrow bound would have failed the merge gate on a legitimate score. Now `[0, 2]`, pinned by `test_a_multi_class_brier_score_is_allowed_its_full_range`.

**2. Sample counts that count samples.** HELM took `num_samples` from a stat's `count`, which is its **train-trial** count — 1 for nearly every published run, whatever the instance count. It now counts per-instance stats keyed per split and perturbation, falls back to HELM's own per-split `num_instances` for the worst-case perturbation stats that have none, and moves the trial count to `score_details.details.num_train_trials`. Inspect took `num_samples` from the length of a sample list that a header-only log does not populate; it now reads the results header.

**3. Uncertainty, as each harness computed it — and never dropped.** HELM republished the 0.0 spread *over a single train trial* as a measured `standard_deviation`; the schema defines that field over per-sample scores, so it is now omitted rather than reported as zero. Inspect dropped a standard error of exactly `0.0` as if it were absent (falsy, not missing), routed nothing for `bootstrap_stderr`, and emitted a scorer's `std` as a **score of its own** alongside the score it describes — it now carries it as `uncertainty.standard_deviation` on those scores, unless `std` is all that scorer reported. lm-eval derives the standard-error method from the aggregation its log records instead of asserting `'bootstrap'`, and reports a resample count only for aggregations that are actually resampled.

`bootstrap_stderr` was in neither the shared bounds nor `DISPERSION_METRICS` while `stderr` was in both, so a converter publishing the former claimed an unknown range and an applicable direction. It is a standard error computed by resampling rather than in closed form — same bounds, same polarity.

**4. `metric_id`, so a consumer can join across harnesses.** The previous PR gave every result a `metric_name`; nothing tied lm-eval's `exact_match` to HELM's. `metric_config_fields` now resolves a metric name to the **eval-card-registry's canonical slug** where the registry carries one (18 of 45 names), and to `<harness>.<name>` where it does not — marked `metric_id_status: unregistered`, so a namespaced id claims a stable join key *within* the harness and no global identity. Both forms record the registry revision they were resolved against (`8b83e9c`, that repo's `main`), because an `unregistered` marker with no revision beside it does not tell a reader whether the entry has since been added.

`metric_kind` (the family a metric aggregates safely within) and `metric_unit` come along because they answer what an id alone leaves open. The unit is **derived from the resolved bounds**, not listed: lm-eval's sacrebleu `bleu` reports `percent` and HELM's nltk `bleu_1` reports `proportion` from their bounds alone, with neither spelled out.

## Why the map is hand-resolved, and how you check it

A converter has to run offline and give the same answer on every run, so `CANONICAL_METRIC_IDS` is 18 hand-verified entries against a pinned revision rather than a network call to the registry's resolver. `tools/verify_metric_ids.py` re-does the resolution against a registry checkout and fails (exit 1) on the three ways the map can rot:

```
$ uv run python -m tools.verify_metric_ids --seed ../eval-card-registry/seed/metrics.yaml
ids resolved against revision 8b83e9c
18 mapped, 45 names checked

MAPPED ID NO LONGER IN THE REGISTRY: 0
NOW RESOLVABLE, STILL NAMESPACED: 0
AMBIGUOUS IN THE REGISTRY: 0

NAMESPACED, WANTING A REGISTRY ENTRY: 22
  acc_norm, bits_per_byte, brier_score, byte_perplexity, calibration_error,
  chain_of_thought_correctness, chrf, classification_macro_f1,
  classification_micro_f1, ece, ifeval_strict_accuracy, math_equiv,
  math_equiv_chain_of_thought, mc1, mc2, mcc, prefix_exact_match,
  quasi_exact_match, quasi_prefix_exact_match, rougeLsum, ter, word_perplexity

HARNESS SLUGS: lm-evaluation-harness ok, helm ok, inspect_ai NOT IN REGISTRY
```

That last list is not a failure — it is the content of a follow-up **eval-card-registry PR**, together with an `@k` slug family for HELM's best-of-k metrics and an `inspect_ai` harness entry. It is a tool rather than a test because it needs a sibling repo checkout, which CI does not have.

Two matching rules in it are worth a second of your attention: normalization keeps `+`, so `chrF++` never answers to `chrf`; and `chrf` (plain) is in the gap list *because* the registry's `chrf-plus-plus` is a different metric.

**Interaction with #222.** That PR adds `check_metric_identity`, which warns on a missing `metric_id`, a semantically-empty one (`score`, `mean`, `overall`, …) and one equal to `evaluation_name`. I ran that branch's validator over all six files these three converters publish: **6 files, 0 complaints, exit 0**, then set one `metric_id` to `'score'` as a tripwire and confirmed the check fires. Namespaced ids survive because `_normalize_metric_id` maps `-`→`_` but leaves `.` alone, so `inspect_ai.mean` is never generic. Since all three converters now always set an id, this PR *removes* #222's "missing" finding for them.

## Checklist

- [x] `uv run python -m every_eval_ever validate <files>` clean, **no warnings**, at the final `data/<collection>/<dev>/<model>/` path — 6 files, 0 complaints, exit 0; asserted by the stack's harness on every run
- [x] every unconvertible source row is in `adapter_reports/` with a non-zero exit — and one case improves: a HELM run whose stats are all bookkeeping is now reported *per run* instead of failing the whole invocation from inside the publisher
- [x] offline unit test added + full `uv run pytest tests` green — core 438 passed / 39 skipped, full 524 passed / 1 skipped
- [x] **model/benchmark/metric ids resolve in the registry** — 18 resolve against `8b83e9c`; the other 22 are namespaced, marked `unregistered`, and listed above for the registry PR. `tests/test_converter_conversion.py::test_every_metric_id_is_either_canonical_or_openly_namespaced` is the enforcement: an id is either in the verified map or carries its harness prefix, never a bare unqualified guess.
- [x] content spot-checked — no answer leakage, nothing double-counted, `evaluation_id` unaffected. `tests/test_converter_metric_bounds.py` (16 tests) pins the shared rules directly.

## Review lane

- [ ] Fast
- [x] **Needs a human** — a material change in outcome. `min_score`/`max_score`/`score_type` disappear from metrics whose range is unknown, `num_samples` changes value for HELM, a HELM `standard_deviation` disappears, an Inspect `std` stops being a score and becomes an uncertainty field, and every result gains a `metric_id`. Same scores, different metadata about them.

Design agreed in: not yet. The specific calls I would want a maintainer to confirm, in order of how much they would cost to change later:

1. **Namespaced ids as the fallback.** `helm.quasi_exact_match` is data that a later registry entry supersedes — every one of those 22 becomes a canonical id in a follow-up, and any record already published under the namespaced form then needs remapping. The alternative is a null `metric_id` until the registry catches up, which is honest but makes the field useless for the harnesses' most common metrics.
2. **Omitting bounds rather than defaulting to [0, 1].** This is the `AGENTS.md` "report, don't interpret" principle applied literally. Of the 45 names the tables carry, 41 get a range and 4 deliberately do not (`calibration_error`, `cer`, `ece`, `wer` — an error rate with no upper bound in the general case); any metric outside every table also gets none, where before HELM and Inspect stamped `[0, 1]` on it. On these fixtures that is 1 of 28 results (Inspect's `mean`). If any downstream consumer treats a null `max_score` as an error rather than as unknown, they will feel it there first.
3. **An Inspect `std` is no longer a score.** It is the one place where a result *count* changes for a reason other than naming.

## Decisions & coverage

- Decision / where: a metric absent from every bounds table gets no bounds plus `bounds_status: unknown`, in `converters/common/metrics.py::metric_bounds_fields`.
  Chose / instead of: keeping `[0, 1]`, which is what HELM and Inspect did and what makes every record look complete.
  Confidence: high — a wrong range is worse than a missing one; the schema makes both fields nullable precisely for this, and the count in `source_metadata` keeps the gap measurable.
  General? yes — this is the rule any converter should follow.
- Decision / where: bounds are layered per harness on top of `SHARED_METRIC_BOUNDS`, in each converter's own table.
  Chose / instead of: one global name→range map.
  Confidence: high — `bleu` is the counterexample that forces it: sacrebleu 0–100 in lm-eval, nltk 0–1 for HELM's `bleu_1`/`bleu_4`. One map would have to be wrong for one of them.
  General? yes.
- Decision / where: `metric_unit` derived from the resolved bounds, with `METRIC_UNITS` holding only the metrics whose bounds cannot show their scale (`bits_per_byte`, `chrf`, `ter`).
  Chose / instead of: listing a unit per metric name.
  Confidence: high — a listed unit and a listed range can disagree; a derived one cannot.
  General? yes.
- Decision / where: a parameter the harness spells into a name takes bounds from the undecorated name but never the id, in `metric_config_fields` (`lookup_name`).
  Chose / instead of: resolving `exact_match@5` to `exact-match`.
  Confidence: high — best-of-five exact match is a different and systematically higher quantity, and the registry gives such metrics their own slug (`pass-at-1`, `recall-at-5`). Sharing `exact-match` would silently average the two in any cross-source query.
  General? yes — it is the general rule for `@k`, `_norm` and similar decorations.
- Decision / where: the registry revision is recorded on **resolved** ids too, not only unregistered ones (`_ID_REVISION`).
  Chose / instead of: marking only the gaps.
  Confidence: high — both are claims about a registry state; a reader of an old record needs to know which state.
  General? yes.
- Decision / where: id resolution is a hand-verified map + an offline `tools/` checker.
  Chose / instead of: calling the registry's resolver at conversion time.
  Confidence: high — converters must run offline and be reproducible; a network lookup makes last month's conversion unrepeatable.
  General? yes.
- Decision / where: a HELM run with no usable stats is reported as a per-run failure rather than raising inside the publisher.
  Chose / instead of: leaving the exception, which fails the whole `--log_path` directory.
  Confidence: high — `SourceConversionResult` + `save_failure_report` + non-zero exit is the repo's documented contract for exactly this.
  General? yes.

**Coverage:** 28 results over the three fixtures — 2 lm_eval, 2 inspect (was 3 — an Inspect `std` becomes an uncertainty field on the scores it describes rather than a score of its own), 24 helm. Nothing silently dropped; 1 of the 28 comes out without bounds and says so, and 22 of the 45 distinct metric names carry a namespaced rather than canonical id, listed above rather than hidden.

**Operator asked about policy calls?** Yes — new canonical ids and unbounded metrics, which is why this is needs-a-human. Both listed under "Design agreed in" above: whether namespaced `<harness>.<name>` ids are acceptable as published data pending registry entries, and whether omitting bounds for a metric no table carries is preferred over the previous `[0, 1]` default.

### DIFF
--- a/every_eval_ever/converters/common/metrics.py
+++ b/every_eval_ever/converters/common/metrics.py
@@ -173,6 +173,7 @@ _PERCENT_BOUNDS = (0.0, 100.0)
 
 _UNKNOWN_BOUNDS = {'bounds_status': 'unknown'}
 _NO_POLARITY = {'polarity': 'not_applicable'}
+_UNKNOWN_POLARITY = {'polarity': 'unknown'}
 _UNREGISTERED_ID = {'metric_id_status': 'unregistered'}
 # Both a resolved id and an unresolved one are claims about a registry state, so
 # each says which state it was: an `unregistered` marker with no revision beside
@@ -200,6 +201,13 @@ def metric_bounds_fields(
     details = dict(_NO_POLARITY) if name in DISPERSION_METRICS else {}
 
     if bounds is None:
+        # `lower_is_better` is required, so an unrecognized metric defaults to
+        # False; for one whose direction we cannot resolve that default is a
+        # guess a ranking consumer would misread as an asserted "higher is
+        # better", so it is marked like an unresolved range. A known direction
+        # (LOWER_IS_BETTER) or an inapplicable one (dispersion) keeps its answer.
+        if name not in LOWER_IS_BETTER and name not in DISPERSION_METRICS:
+            details = {**details, **_UNKNOWN_POLARITY}
         return {
             'lower_is_better': name in LOWER_IS_BETTER,
             'additional_details': {**details, **_UNKNOWN_BOUNDS},
--- a/every_eval_ever/converters/common/metrics.py
+++ b/every_eval_ever/converters/common/metrics.py
@@ -39,7 +39,9 @@ SHARED_METRIC_BOUNDS: dict[str, tuple[float, float]] = {
     'precision': (0.0, 1.0),
     'recall': (0.0, 1.0),
     'mcc': (-1.0, 1.0),
-    'brier_score': (0.0, 1.0),
+    # Not [0, 1]: the multi-class Brier score sums the squared error over every
+    # class, so all the mass on one wrong class costs 2.0.
+    'brier_score': (0.0, 2.0),
     # Dispersion of a score distribution, not a score: non-negative, and
     # unbounded above unless the underlying metric is bounded.
     'std': (0.0, float('inf')),
--- a/every_eval_ever/converters/common/metrics.py
+++ b/every_eval_ever/converters/common/metrics.py
@@ -35,6 +35,7 @@ SHARED_METRIC_BOUNDS: dict[str, tuple[float, float]] = {
     'std': (0.0, float('inf')),
     'stddev': (0.0, float('inf')),
     'stderr': (0.0, float('inf')),
+    'bootstrap_stderr': (0.0, float('inf')),
     'var': (0.0, float('inf')),
 }
 
@@ -59,7 +60,7 @@ LOWER_IS_BETTER: frozenset[str] = frozenset(
 # (`uncertainty.standard_error`, `uncertainty.standard_deviation`), so a
 # converter should prefer routing them there over emitting them as scores.
 DISPERSION_METRICS: frozenset[str] = frozenset(
-    {'std', 'stddev', 'stderr', 'var'}
+    {'std', 'stddev', 'stderr', 'bootstrap_stderr', 'var'}
 )
 
 _UNKNOWN_BOUNDS = {'bounds_status': 'unknown'}
--- a/every_eval_ever/converters/common/metrics.py
+++ b/every_eval_ever/converters/common/metrics.py
@@ -1,4 +1,4 @@
-"""What a metric's name tells us about its range and its polarity.
+"""What a metric's name tells us about its identity, its range and its polarity.
 
 Upstream harnesses report a metric's value and almost never its range, so the
 range has to come from the metric's definition. The names below mean the same
@@ -11,6 +11,16 @@ reports 0-100 while nltk's `sentence_bleu` (HELM's `bleu_1`/`bleu_4`) reports
 A metric that is in no table gets no bounds at all and a `bounds_status` marker,
 because `min_score`/`max_score` are nullable and "not provided" is true, while
 [0, 1] on an unbounded metric is not.
+
+`metric_id` is the field a consumer joins on across sources, so a metric that
+means the same thing in two harnesses has to arrive under the same id. That id
+is the eval-card-registry's canonical metric slug, resolved once by hand into
+`CANONICAL_METRIC_IDS` rather than over the network, because a converter has to
+run offline and give the same answer every time. Anything the registry does not
+carry gets `<harness>.<name>`, which is a stable join key within the harness and
+claims no global identity; `metric_id_status` marks those so the set needing a
+registry entry can be listed. Re-check the map against a registry checkout with
+`uv run python -m tools.verify_metric_ids --seed <registry>/seed/metrics.yaml`.
 """
 
 from __future__ import annotations
@@ -63,8 +73,109 @@ DISPERSION_METRICS: frozenset[str] = frozenset(
     {'std', 'stddev', 'stderr', 'bootstrap_stderr', 'var'}
 )
 
+# The eval-card-registry commit these ids were resolved against, on that repo's
+# main branch so anyone can check out the state that produced them. Bump it with
+# the map.
+METRIC_ID_REGISTRY_REVISION = '8b83e9c'
+
+# Harness metric name -> canonical registry metric id, matched case- and
+# separator-insensitively against each entry's id, display_name and aliases.
+# Only names the registry actually carries appear here; the rest are namespaced.
+CANONICAL_METRIC_IDS: dict[str, str] = {
+    'acc': 'accuracy',
+    'accuracy': 'accuracy',
+    'bleu': 'bleu',
+    'bleu_1': 'bleu-1',
+    'bleu_4': 'bleu-4',
+    'cer': 'cer',
+    'em': 'exact-match',
+    'exact_match': 'exact-match',
+    'f1': 'f1',
+    'f1_score': 'f1',
+    'perplexity': 'perplexity',
+    'precision': 'precision',
+    'recall': 'recall',
+    'rouge1': 'rouge-1',
+    'rouge2': 'rouge-2',
+    'rougeL': 'rouge-l',
+    'rouge_l': 'rouge-l',
+    'wer': 'wer',
+}
+
+# The family a metric aggregates safely within. Coarse on purpose: it says two
+# numbers are the same kind of quantity, not that they are comparable, which is
+# `evaluation_name`'s business. A metric whose family we would have to invent is
+# absent rather than guessed.
+METRIC_KINDS: dict[str, str] = {
+    'acc': 'accuracy',
+    'acc_norm': 'accuracy',
+    'accuracy': 'accuracy',
+    'chain_of_thought_correctness': 'accuracy',
+    'em': 'accuracy',
+    'exact_match': 'accuracy',
+    'ifeval_strict_accuracy': 'accuracy',
+    'math_equiv': 'accuracy',
+    'math_equiv_chain_of_thought': 'accuracy',
+    'mc1': 'accuracy',
+    'prefix_exact_match': 'accuracy',
+    'quasi_exact_match': 'accuracy',
+    'quasi_prefix_exact_match': 'accuracy',
+    'classification_macro_f1': 'f1',
+    'classification_micro_f1': 'f1',
+    'f1': 'f1',
+    'f1_score': 'f1',
+    'precision': 'precision',
+    'recall': 'recall',
+    'mcc': 'correlation',
+    'brier_score': 'brier_score',
+    'calibration_error': 'calibration_error',
+    'ece': 'calibration_error',
+    'bleu': 'text_overlap',
+    'bleu_1': 'text_overlap',
+    'bleu_4': 'text_overlap',
+    'chrf': 'text_overlap',
+    'rouge1': 'text_overlap',
+    'rouge2': 'text_overlap',
+    'rougeL': 'text_overlap',
+    'rougeLsum': 'text_overlap',
+    'rouge_l': 'text_overlap',
+    'cer': 'edit_distance',
+    'ter': 'edit_distance',
+    'wer': 'edit_distance',
+    'byte_perplexity': 'perplexity',
+    'perplexity': 'perplexity',
+    'word_perplexity': 'perplexity',
+    'bits_per_byte': 'compression_rate',
+    'std': 'dispersion',
+    'stddev': 'dispersion',
+    'stderr': 'dispersion',
+    'var': 'dispersion',
+    'bootstrap_stderr': 'dispersion',
+}
+
+# A unit that does not follow from the metric's resolved range. Everything on
+# exactly [0, 1] is a proportion and everything on exactly [0, 100] a percent,
+# which is derived rather than listed, so the same name reports the unit of
+# whichever harness table won: lm-eval's sacrebleu `bleu` is a percent and
+# HELM's nltk `bleu_1` a proportion, from their bounds alone. Listed here are
+# the metrics whose scale their bounds cannot show, such as a sacrebleu metric
+# that is percent-scaled but may exceed 100.
+METRIC_UNITS: dict[str, str] = {
+    'bits_per_byte': 'bits_per_byte',
+    'chrf': 'percent',
+    'ter': 'percent',
+}
+
+_PROPORTION_BOUNDS = (0.0, 1.0)
+_PERCENT_BOUNDS = (0.0, 100.0)
+
 _UNKNOWN_BOUNDS = {'bounds_status': 'unknown'}
 _NO_POLARITY = {'polarity': 'not_applicable'}
+_UNREGISTERED_ID = {'metric_id_status': 'unregistered'}
+# Both a resolved id and an unresolved one are claims about a registry state, so
+# each says which state it was: an `unregistered` marker with no revision beside
+# it does not tell a reader whether the entry has since been added.
+_ID_REVISION = {'metric_id_registry_revision': METRIC_ID_REGISTRY_REVISION}
 
 
 def metric_bounds_fields(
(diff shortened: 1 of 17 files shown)


every_eval_ever/converters/common/metrics.py
```diff
@@ -0,0 +1,286 @@
...
+        if bounds_table
+        else SHARED_METRIC_BOUNDS
+    )
+    name = metric_name or ''
+    bounds = table.get(name)
+    details = dict(_NO_POLARITY) if name in DISPERSION_METRICS else {}
+
+    if bounds is None:
+        return {
+            'lower_is_better': name in LOWER_IS_BETTER,
```
mrshu, line 212: **RAV-RUN1-R1-F001 [medium] — suggestion:** Could unknown direction be marked explicitly here, similarly to dispersion's `not_applicable` marker? For any unrecognized loss/error metric this required `false` reads as an affirmative higher-is-better claim, while the current additional details only qualify the bounds. Keeping the boolean but adding `polarity: unknown` until a harness or supplement supplies direction would make the fallback safer for ranking consumers.

every_eval_ever/converters/lm_eval/utils.py
```diff
@@ -16,6 +16,69 @@ def evaluation_result_id(metric_name: str, filter_name: str) -> str:
...
+
+def bootstrap_resamples(
+    metric_name: str,
+    aggregation: Optional[str],
+    configured_iters: Optional[int],
+) -> Optional[int]:
+    """How many resamples went into a bootstrapped standard error."""
+    if configured_iters is None or aggregation not in BOOTSTRAP_AGGREGATIONS:
+        return None
+    if metric_name in CAPPED_BOOTSTRAP_METRICS:
```
mrshu, line 77: **RAV-RUN1-R1-F002 [medium] — suggestion:** lm-eval applies this 100-resample cap to the aggregation callable, so an aliased metric can use `aggregation: bleu` without being named `bleu`. Keying this check on `aggregation` (as `standard_error_method()` already does) would keep `num_bootstrap_samples` faithful for custom metric names too.

every_eval_ever/converters/inspect/utils.py
```diff
@@ -337,6 +346,10 @@ def extract_model_info_from_model_path(model_path: str) -> ModelInfo:
     "has_unknown_level",
     "min_score",
     "max_score",
+    "metric_id",
```
mrshu, line 349: **RAV-RUN1-R1-F003 [medium] — suggestion:** These four new fields look intentionally overrideable, but the strict `SupplementalMetricConfig` does not declare them, so exact probes reject each as `extra_forbidden` before this allowlist runs. Adding matching model fields and an end-to-end override test would make this useful correction path reachable; a `metric_id` override can also refresh its generated status/revision atomically.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — This is a careful and well-tested improvement to
metric provenance. The focused suite and all CI matrices are green; four data
semantics edge cases and two small verification-tool refinements remain as
constructive suggestions.
**Result:** Metric bounds, units, uncertainty, sample counts, and identities now
reflect the upstream harnesses much more faithfully, with unknowns made visible.
**Scope:** Adds shared metric metadata policy, converter-specific uncertainty and
sample-count handling, canonical/namespaced metric IDs, and an offline registry
verification tool.
**Checks:** 6 concerns checked; all 6 reproduced, with 2 lowered to low priority
after impact analysis.
**Second check:** targeted, 2 Codex reviewers; 4 medium findings upheld and 2
tooling findings bounded to low priority.

## Earlier review comments

None.

## What I noticed

|---|---|---|---|---|
| RAV-RUN1-R1-F001 | medium | unknown metric polarity | `every_eval_ever/converters/common/metrics.py:204` | Unknown metric names receive `lower_is_better=false`, while only their bounds are marked unknown. A custom loss/error metric can therefore look affirmatively higher-is-better. (inline) |
| RAV-RUN1-R1-F002 | medium | lm-eval bootstrap count | `every_eval_ever/converters/lm_eval/utils.py:77` | The 100-resample cap is selected by the reported metric name rather than its aggregation. An aliased metric using lm-eval's `bleu`, `chrf`, or `ter` aggregation can report the configured count although lm-eval used only 100. (inline) |
| RAV-RUN1-R1-F003 | medium | Inspect metric overrides | `every_eval_ever/converters/inspect/utils.py:349` | Four fields are newly listed as supplement-overridable, but the strict `SupplementalMetricConfig` declares none of them, so every attempted override is rejected before this allowlist runs. (inline) |
| RAV-RUN1-R1-F004 | medium | Inspect uncertainty routing | `every_eval_ever/converters/inspect/adapter.py:234` | If both `stderr` and `bootstrap_stderr` are present, insertion order selects which one survives and the other is removed. The selected metric's parameters are also dropped, including bootstrap counts or clustering information. (inline) |

<details>
<summary>Non-blocking low-priority findings</summary>

- **RAV-RUN1-R1-F005 [low] registry gap report** — The verifier removes all
  dispersion names from its informational unregistered list, although Inspect
  can still emit `var` and lone dispersion metrics as namespaced results. Its
  actual stale/newly-resolvable/ambiguous checks still cover them.
- **RAV-RUN1-R1-F006 [low] registry provenance** — The tool accepts any seed
  file but always prints the hard-coded registry revision without checking the
  seed checkout's Git revision. Content mismatches are still detected, so this
  is primarily a provenance-labeling refinement.

</details>

## Things to try

- **[medium] unknown metric polarity** — Keep the schema-required boolean, but
  add a `polarity: unknown` marker for names without a verified direction and
  distinguish those from known higher-is-better metrics. Harness-provided
  direction can then clear the marker. (`RAV-RUN1-R1-P001`; covers
  `RAV-RUN1-R1-F001`)
- **[medium] lm-eval bootstrap count** — Apply the cap using `aggregation`, as
  `standard_error_method()` already does, and add an aliased-name/`bleu`
  aggregation regression test. (`RAV-RUN1-R1-P002`; covers
  `RAV-RUN1-R1-F002`)
- **[medium] Inspect metric overrides** — Add the four fields to
  `SupplementalMetricConfig` and test a real override. For `metric_id`, updating
  the generated registration status/revision atomically would keep identity
  metadata consistent. (`RAV-RUN1-R1-P003`; covers `RAV-RUN1-R1-F003`)
- **[medium] Inspect uncertainty routing** — Handle multiple standard-error
  metrics deterministically (or preserve the alternate in details), and carry
  the chosen metric's bootstrap count/clustering parameters into uncertainty
  metadata. (`RAV-RUN1-R1-P004`; covers `RAV-RUN1-R1-F004`)
- **[low] registry gap report** — Exclude only dispersion metrics guaranteed
  not to be emitted, or list fallback-emitted dispersion IDs separately.
  (`RAV-RUN1-R1-P005`; covers `RAV-RUN1-R1-F005`)
- **[low] registry provenance** — Compare the seed checkout's revision with the
  pinned constant, or label the printed revision explicitly as the map's
  expected provenance. (`RAV-RUN1-R1-P006`; covers `RAV-RUN1-R1-F006`)

<details>

- Target: PR #246 at `f10d4d7aab6df0454a9f957dd82c58b438976f5e`
  (16 files, +1865/-129)
- Run ordinal: 1
- Mix: 3 Codex reviewers
- Focus: bounds, polarity, uncertainty, sample counts, metric identity,
  registry verification, and positive suggestion-oriented language
- Finding counts: 0 critical, 0 high, 4 medium, 2 low, 0 nit
- Verification: exact-head focused converter run — 55 passed, 24
  optional-extra skips; GitHub Actions passed all four core/full locked/loose
  matrices
- Reproduction: one batched Codex verifier confirmed 3 single-reviewer medium
  candidates and lowered 2 verifier-tool candidates to low; consensus covered
  the remaining medium candidate
- Second check: targeted; reviewers=2; kept=4/lowered=2/removed=0; approval
  changed no

</details>

---

every_eval_ever/converters/inspect/adapter.py
```diff
@@ -208,29 +231,50 @@ def _extract_evaluation_results(
                     ),
                 )
 
-            stderr_value = next(
+            stderr_value, stderr_method = next(
```
mrshu: **RAV-RUN1-R1-F004 [medium] — suggestion:** If a scorer carries both `stderr` and `bootstrap_stderr`, this `next(...)` makes the published value and method depend on mapping order while both source entries are later removed. Could the collision have deterministic policy (or preserve the alternate in details), and could the chosen metric's params populate bootstrap count/clustering metadata? That would retain the useful uncertainty provenance this PR is adding.

borgr:
Went through all six findings. This run targeted `f10d4d7`; two of the four medium items were already resolved by later commits on the branch, one is a real fix, and one I'm leaving with reasoning. Both low items are now addressed.

**RAV-RUN1-R1-F001 — unknown-direction metrics look higher-is-better.** Resolved at the current head. `metric_bounds_fields` now marks an unrecognized, non-dispersion metric with `{'polarity': 'unknown'}` in `additional_details` (via `_UNKNOWN_POLARITY`) alongside the schema-required `lower_is_better=false`, exactly the explicit marker suggested — a `LOWER_IS_BETTER` or dispersion metric keeps its known answer. The reviewed commit predated that.

**RAV-RUN1-R1-F003 — four override fields not declared on the strict model.** Resolved at the current head. `SupplementalMetricConfig` now declares `metric_id`, `metric_kind`, `metric_unit`, and `metric_parameters`, so the `SYNTHETIC_METRIC_CONFIG_FIELDS` allowlist entries are reachable rather than rejected as `extra_forbidden`. Also predated by the reviewed commit.

**RAV-RUN1-R1-F004 — both stderrs present, insertion order decides.** Fixed (`93c516a05`). When a scorer reports both `stderr` and `bootstrap_stderr`, the analytic one is now chosen deterministically as `standard_error` (preference `('analytic', 'bootstrap')`) instead of by dict order, and the unselected method — its value and its `params` (bootstrap counts, clustering) — is preserved in `score_details.details` rather than dropped. A regression test feeds both with bootstrap first and asserts the analytic value survives while the bootstrap value is kept in details.

**RAV-RUN1-R1-F002 — bootstrap cap keyed on metric name, not aggregation.** Left as is: keying on the name is what matches lm-eval. In `lm_eval/evaluator_utils.py` (0.4.12) the cap is `bootstrap_iters=min(bootstrap_iters, 100) if metric in ["bleu", "chrf", "ter"] else bootstrap_iters` — the condition reads the metric-name string; the aggregation callable is only what gets bootstrapped. So lm-eval does **not** cap an aliased metric whose name is not one of those three, even when it uses `aggregation: bleu` — it runs the full configured count. `bootstrap_resamples` mirrors that by keying `CAPPED_BOOTSTRAP_METRICS` on the name. Switching to the aggregation would make the converter report `num_bootstrap_samples = 100` for that aliased metric while lm-eval actually used the configured count — i.e. it would diverge from the harness it's meant to reflect. If a future lm-eval moves the cap onto the aggregation, the check should move with it, but for the pinned behavior the name is correct.

**RAV-RUN1-R1-F005 — dispersion names dropped from the gap report.** Fixed (`84649b08e`). `verify_metric_ids` still keeps dispersion out of the "wanting a registry entry" count (it belongs in `score_details.uncertainty`), but it no longer drops the five names from view: `bootstrap_stderr`, `std`, `stddev`, `stderr`, `var` now print in their own "namespaced when published" section, so the ones a converter can emit namespaced (e.g. `var`) stay visible.

**RAV-RUN1-R1-F006 — printed revision implied the seed's revision.** Fixed (`84649b08e`). The line now states the revision is the map's own pin and that the `--seed` file is read as given, its checkout revision not verified — rather than implying the seed sits at that revision.

borgr:
Reviewed the 4 unresolved automated review threads against the current head. All four are dismissed: the unknown-polarity marker (`metrics.py:212`) and the four override fields on `SupplementalMetricConfig` (`inspect/utils.py:349`) are both present in this PR's own diff; the bleu/chrf/ter resample cap (`lm_eval/utils.py:77`) is correctly keyed on the metric *name* — lm-eval's `_collect_results` caps by the metric key, not the aggregation callable, so the finding's premise is wrong; and the stderr-preference ordering (`inspect/adapter.py`) is deterministic with the unselected method preserved. Resolving all 4.

PR #247, opened 2026-08-11
[Feature] DRAFT/DO NOT MERGE - Add a daily adapter refresh cron

evijit:
## What this does

Runs each source adapter once a day and sends its records to that adapter's own
pull request on [`EEE_datastore`](https://huggingface.co/datasets/evaleval/EEE_datastore).

Closes the cron ticket. Design notes live in
[`every_eval_ever/cron/README.md`](every_eval_ever/cron/README.md).

**This is a structural change and a new subsystem, so it belongs in the
"needs a human" lane** — I'm opening it as the concrete proposal to react to, not
as something to fast-track. The four decisions below were pre-agreed with
@evijit before implementation; everything else is up for discussion.

## The four decisions, as agreed

| Decision | Choice |
|---|---|
| Raw data storage | Permanent, in a **separate private HF dataset** with a ledger (revised after @nelaturuharsha's feedback — was artifacts) |
| Daily coverage | Every eligible adapter, one job each |
| Raw capture mechanism | One hook in `helpers/fetch.py`; **no adapter was modified** |
| De-duplication | Skip a run when the source has not changed; no record-level dedup |

## How it works

`every_eval_ever/cron/schedule.py` declares which adapters the cron may run and
how to run each one — adapter CLIs are not uniform, and this records the
differences in one place instead of spreading them through a workflow file.
**17 of 25 adapters are scheduled** now that the API keys are in place. The other 10 are listed with the
reason each is excluded:

- `arc_agi`, `livecodebenchpro`, `mercor_eval` — legacy, upstream unusable
- `bfcl`, `cocoabench`, `sciarena` — need a hand-supplied input file
- `vectara_hallucination_leaderboard` — pinned to `SOURCE_COMMIT`, so a refresh
  is a code change and a daily run could only re-emit identical records
- ~~`artificial_analysis`, `llm_stats` — waiting on API keys~~ → **now enabled**,
  their secrets are in place
- `hfopenllm_v2` — emits >4500 records per run. Enabling it before record-level
  dedup exists would add thousands of near-duplicate files to its PR every time
  the leaderboard moves, which is the problem @nelaturuharsha flagged. One-line
  change to enable once dedup lands.

`runner.py` then runs one adapter end to end: run it into a scratch tree,
archive what it fetched, stamp the records, validate them through the same CLI
the datastore bot runs, and commit.

### Tracking cron-generated records

Every record gets this in `source_metadata.additional_details`:

```json
"type_of_addition": "cron",
"cron_run_date": "2026-08-11",
"cron_adapter": "hle",
"cron_run_url": "https://github.com/evaleval/every_eval_ever/actions/runs/...",
"cron_unknown_inferred_fields": "deployment_type,model_availability"
```

The run URL is there so a record traces back to the exact run whose artifact
holds its raw data.

**On the inferred fields:** the ticket asks for `deployment_type` and
`model_availability` to be marked `unknown/unknown`. That already happens —
`ModelInfo.default_model_metadata` (added by `post_codegen.py`) defaults both.
So the cron does not fill them; it *names* the ones that came out `unknown`, per
record, which is what a post-fix actually needs to find them. It never
overwrites a value an adapter did state.

### Raw data — permanent, with a ledger

`helpers/raw_capture.py` archives response bodies when `EEE_RAW_CAPTURE_DIR` is
set, hooked into `fetch_json`/`fetch_csv`. Because those are the shared helpers,
8 adapters got raw capture with **no adapter changes at all**. Adapters with
their own `--save-raw-*` flag are pointed at the same directory. HF-dataset and
git-cloned sources are deliberately **not** re-archived — upstream already
versions them, which is the "don't download what's already in place" requirement.

Payloads then go to a private dataset of their own — `evaleval/EEE_raw`,
overridable with an `EEE_RAW_REPO_ID` repository variable:

```
blobs/<ab>/<sha256>.<ext>              # one copy per distinct payload, ever
ledger/<adapter>/<date>-<run>.jsonl    # one row per payload per run
```

Payloads are **content-addressed**, so a source that has not changed since
yesterday costs nothing but a ledger row — that is what makes keeping every day
affordable rather than duplicating a leaderboard 365 times a year. The ledger is
a dataset in its own right (`load_dataset('json',
data_files='ledger/**/*.jsonl')`), one row per payload per run carrying the
adapter, run date, run URL, source URL, SHA-256, size and blob path. A row is
written even for a payload too large to store, and even for a run that published
nothing, so it says what was fetched on **every** date rather than only the days
something changed.

Two ordering decisions worth a look:

- **Archive before publish, and fail closed.** If raw data cannot be stored, the
  run stops and publishes nothing. Records should not reach the datastore without
  their provenance stored somewhere permanent.
- **The ledger replaced the Actions cache** as the memory of the previous run's
  fingerprint. With a cache, an eviction would make an adapter forget last
  night's state and republish its entire set — precisely the duplicate flood this
  is meant to prevent. The workflow now has no cache steps at all.

Capture is best-effort and never breaks a fetch, and happens *before* parsing, so
a body that fails to parse is still archived — the smoke run below shows a
27 KB CSV kept from an adapter that then crashed.

### Not republishing unchanged data

A run fingerprints the source and stops if it matches what the *last successful
publish* recorded (`state/<adapter>.json` in the raw dataset — written only
after the datastore commit succeeds, which is what makes a failed publish retry
tomorrow instead of being skipped as "unchanged"). It is **run-level, not record-level**: a run publishes everything it produced or
nothing at all, so no individual record is ever dropped. That keeps the ticket's
"hold off on dedup" while still preventing a daily duplicate flood.

One subtlety worth flagging, because it cost a rewrite: only response bodies
stored *exactly as served* can be compared across runs. `hle`'s own
`--save-raw-json` wraps its payload in a `fetched_at` timestamp and `vals_ai`'s
is a normalized form, so fingerprinting those would report a change every single
day and the skip would never fire. Those adapters are archived-but-gated-on-output
instead. I found this by running it against the live source, not by reading it.

## What guarantees one source can't break another

Asked for explicitly, so it is enforced in three places:

1. **Job level** — one matrix job per adapter with `fail-fast: false`, so a
   failure or a hang is confined to that adapter. Each job has its own 6-hour
   budget and a 120-minute timeout, and `max-parallel: 6` means nothing queues
   behind a slow adapter. (The 6-hour ceiling is *per job*, so the matrix already
   avoids a single chokepoint — the parallelism is what removes the queueing.)
2. **Invocation level** — `helm` runs five leaderboards; one failing does not
   stop the other four, and their records still publish.
3. **Partial conversions** — this fixed a real bug in my first draft. Per
   `adapters/README.md`, an adapter reports a partial conversion by writing every
   valid record and *then* exiting non-zero. Treating non-zero as fatal would
   have thrown away good records. A partial refresh now publishes what converted,
   says so in the commit description, and raises a workflow warning
   annotation.

`tests/test_cron_workflow.py` asserts these workflow properties, because a YAML
regression here is silent — it shows up as a day of missing data.

## Verification

555 tests pass, `ruff check` clean. 186 new tests across schedule, stamping, raw
capture, archiving, publishing, preflight, the runner, and the workflow. No test
can reach the Hub, and the runner tests share a per-test state store between
consecutive refreshes instead of stubbing the change gate — stubbing the gate is
exactly what let the v3 gating bugs ship.

Smoke-tested against live sources with `--dry-run`:

| Adapter | Result |
|---|---|
| `terminal_bench_2` | 142 records, 0 errors, 0 warnings, 562 KB raw archived |
| `hle` | 50 records, 0 errors, 0 warnings, 27 KB raw archived |
| `hle` (second run) | `unchanged` — published nothing, as intended |
| `mmlu_pro` | Failed, correctly (see below) |

The `mmlu_pro` failure is a **pre-existing adapter bug this surfaced**, not
something in this PR: a leaderboard row has the model name
`seed-oss-36b-base-w-syn.`, and a trailing dot is rejected by
`_required_path_component`. The cron reported it, archived the raw CSV anyway,
and correctly classified the run as failed. Worth its own issue.

## What this does not solve

Stated plainly, since the ticket framed this as an MVP:

- **No record-level dedup.** Records accumulate in each adapter's PR until
  someone merges it. This is why `hfopenllm_v2` stays off.
- **3 adapters archive no raw data** (`hal`, `lexam`, `mt_bench`) — they call
  `requests` directly with no raw-dump flag. They are gated on output instead, so
  they still will not republish unchanged records.
- **The workflow installs `--all-extras`** because three adapters need
  `datasets`, which reaches the lock only via the `helm` extra. A dedicated
  `cron` extra would be lighter, but that needs a `uv.lock` regeneration I could
  not do here.

## Before this can run for real

- [x] `ARTIFICIAL_ANALYSIS_API_KEY` and `LLM_STATS_API_KEY` — added, so both
      adapters are now scheduled
- [x] `HF_TOKEN` — present
- [ ] **Confirm `HF_TOKEN` has the access this needs**: write on
      `EEE_datastore`, plus permission to create/write the private raw dataset.
      Worth an explicit check because the token predates this PR
      (last updated 2026-02-05) and there is also an `EEE_INGESTION_HF_TOKEN`
      secret that may be the intended one — say the word and I will switch to it.
      `uv run python -m every_eval_ever.cron preflight` answers this in one call,
      and the workflow runs it before planning.
- [ ] Merge this, since a `workflow_dispatch` only becomes available once the
      workflow exists on the default branch
- [ ] Then: dispatch with `dry_run: true` on one adapter, check the step summary,
      and only after that let the schedule run

### Verification still owed

`artificial_analysis` and `llm_stats` are the only scheduled adapters I have not
smoke-tested end to end — I never had their keys locally, and a dry run cannot
reach them. They are also two of the five that will *not* fall back to an output
fingerprint, since they fetch through `helpers.fetch` and so get a verbatim
capture. First real run for those two is worth watching.

### Revision history

- **v5** — review-anvil round 3 (all round-2 findings confirmed fixed upstream).
  Adapter subprocesses no longer receive the cron's write-capable HF token
  (scrubbed from the child env; an explicit `source_hf_token` + separate
  `EEE_SOURCE_HF_TOKEN` read token exists for sources that need authenticated
  reads). Mid-batch publish failures are retry-safe: an attempt record written
  before the first batch lets the next run delete the dead attempt's files from
  the PR and republish one clean copy. Capture-failure evidence (sibling
  captures, error rows, failure reports) is archived *before* the run fails.
  Adapter failure reports — which embed raw source rows — moved from the public
  artifact into the private dataset (`reports/<adapter>/…`); the artifact now
  carries only `summary.json`. An enabled adapter missing its credential fails
  its own job instead of silently vanishing from the matrix. A state write that
  fails after retry returns non-zero with the PR URL kept in the summary.

- **v4** — a full 8-angle review pass plus @mrshu's review-anvil findings. The
  big one: the change gate was broken three ways (a run compared against the
  ledger row it had just written; zero-capture adapters never persisted a
  fingerprint; a failed publish poisoned the next day's gate). The gate now
  reads per-adapter state written only after a successful publish. Also fixed:
  capture filenames are content-addressed so one URL serving two bodies cannot
  corrupt the blob store; a public raw dataset fails preflight **and** every
  archive commit re-verifies privacy; PR adoption is filtered by author so a
  stranger's identically-titled PR is never committed onto; the PR-creating
  commit carries the bare title (a batch suffix would have retitled the PR and
  opened a fresh one daily for >300-file adapters); nothing-new exits 3, not 2,
  so an argparse usage error can never read as a quiet success; plus the reuse/
  efficiency items (validator-shared record canonicalization, one shared record
  serializer, batched blob existence checks, dead code removed).

- **v3** — added a `preflight` command, run ahead of planning, after the API
  keys landed: it checks token identity, both destinations, and per-adapter
  credentials. It also covers a gap — a `--dry-run` refresh does not archive, so
  nothing else exercised raw-dataset access without publishing.
- **v2** — raw data moved from workflow artifacts to a permanent private dataset
  with a content-addressed store and a ledger, per @nelaturuharsha. The ledger
  also replaced the Actions cache for fingerprint memory. Artifacts remain as a
  90-day convenience copy of the reports, not as retention.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/adapter_cron.yml
+++ b/.github/workflows/adapter_cron.yml
@@ -31,17 +31,9 @@ permissions:
   contents: read
 
 env:
-  # One place for the credentials both planning and refreshing need. An
-  # adapter's requires_env name must appear here (tests enforce it).
-  HF_TOKEN: ${{ secrets.HF_TOKEN }}
-  ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
-  LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
+  # Not a secret; safe at workflow scope. Secrets are deliberately passed only
+  # to the command steps that use them, never to checkout/setup/artifact steps.
   EEE_RAW_REPO_ID: ${{ vars.EEE_RAW_REPO_ID || 'evaleval/EEE_raw' }}
-
-concurrency:
-  # Never let two refreshes of the same adapter race for its pull request.
-  group: adapter-cron
-  cancel-in-progress: false
 
 jobs:
   plan:
@@ -62,6 +54,11 @@ jobs:
       - name: Check credentials and destinations
         env:
           UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+          # Adapter keys are present so preflight reports which adapters are
+          # held back. requires_env names must appear here (tests enforce it).
+          ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
+          LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
         run: |
           uv run python -m every_eval_ever.cron preflight \
             --raw-repo-id "$EEE_RAW_REPO_ID" \
@@ -72,6 +69,10 @@ jobs:
         env:
           UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
           REQUESTED: ${{ inputs.adapter }}
+          # Presence-only: the schedule uses these to tell a credentialed
+          # adapter apart from an unconfigured one.
+          ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
+          LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
         run: |
           set -euo pipefail
           if [ -n "${REQUESTED}" ]; then
@@ -94,6 +95,14 @@ jobs:
     name: ${{ matrix.adapter }}
     needs: plan
     if: needs.plan.outputs.adapters != '[]'
+    # Scoped to one adapter and one mode: two publishes of the same adapter
+    # must never race for its pull request, but unrelated adapters, and dry
+    # runs against publishes, should not queue behind each other — GitHub also
+    # keeps only one pending run per group and silently replaces the older one,
+    # so a wide group can drop a refresh entirely.
+    concurrency:
+      group: adapter-cron-${{ matrix.adapter }}-${{ inputs.dry_run && 'dry-run' || 'publish' }}
+      cancel-in-progress: false
     runs-on: ubuntu-latest
     # Well under the 6-hour job ceiling: an adapter that hangs should fail
     # today rather than burn a runner and collide with tomorrow's run.
@@ -118,6 +127,10 @@ jobs:
       - name: Refresh ${{ matrix.adapter }}
         env:
           UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+          # Each adapter job receives only its own source credential.
+          ARTIFICIAL_ANALYSIS_API_KEY: ${{ matrix.adapter == 'artificial_analysis' && secrets.ARTIFICIAL_ANALYSIS_API_KEY || '' }}
+          LLM_STATS_API_KEY: ${{ matrix.adapter == 'llm_stats' && secrets.LLM_STATS_API_KEY || '' }}
         run: |
           set -uo pipefail
           # The default shell is `bash -e`, which would abort the step before
@@ -166,15 +179,15 @@ jobs:
                    "invocation(s) failed; published \(.records) record(s)"
                  else empty end' "$summary"
 
-      # A short-lived convenience copy. Permanent retention is the raw dataset;
-      # this is here so a failed run's reports are one click away.
+      # Reports and the run summary only — never cron-work/raw: artifacts on a
+      # public repository are downloadable by anyone signed in, and raw source
+      # bodies belong solely in the private raw dataset.
       - name: Upload reports and summary
         if: always()
         uses: actions/upload-artifact@v4
         with:
           name: cron-${{ matrix.adapter }}-${{ github.run_number }}
           path: |
-            ${{ runner.temp }}/cron-work/raw/
             ${{ runner.temp }}/cron-work/adapter_reports/
             ${{ runner.temp }}/cron-work/summary.json
           if-no-files-found: warn
--- a/.github/workflows/adapter_cron.yml
+++ b/.github/workflows/adapter_cron.yml
@@ -30,6 +30,14 @@ on:
 permissions:
   contents: read
 
+env:
+  # One place for the credentials both planning and refreshing need. An
+  # adapter's requires_env name must appear here (tests enforce it).
+  HF_TOKEN: ${{ secrets.HF_TOKEN }}
+  ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
+  LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
+  EEE_RAW_REPO_ID: ${{ vars.EEE_RAW_REPO_ID || 'evaleval/EEE_raw' }}
+
 concurrency:
   # Never let two refreshes of the same adapter race for its pull request.
   group: adapter-cron
@@ -54,22 +62,15 @@ jobs:
       - name: Check credentials and destinations
         env:
           UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
-          HF_TOKEN: ${{ secrets.HF_TOKEN }}
-          ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
-          LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
         run: |
           uv run python -m every_eval_ever.cron preflight \
-            --raw-repo-id "${{ vars.EEE_RAW_REPO_ID || 'evaleval/EEE_raw' }}" \
+            --raw-repo-id "$EEE_RAW_REPO_ID" \
             --markdown "$GITHUB_STEP_SUMMARY"
 
       - name: Decide which adapters to refresh
         id: plan
         env:
           UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
-          # Present here only so the schedule can tell a credentialed adapter
-          # apart from an unconfigured one.
-          ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
-          LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
           REQUESTED: ${{ inputs.adapter }}
         run: |
           set -euo pipefail
@@ -115,32 +116,28 @@ jobs:
         run: uv sync --locked --all-extras
 
       - name: Refresh ${{ matrix.adapter }}
-        id: refresh
         env:
           UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
-          HF_TOKEN: ${{ secrets.HF_TOKEN }}
-          ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
-          LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
         run: |
           set -uo pipefail
-          # The default shell is `bash -e`, which would abort the step before the
-          # exit code can be read. Exit 2 is a successful no-op, not a failure.
+          # The default shell is `bash -e`, which would abort the step before
+          # the exit code can be read. Exit 3 is a successful no-op ("nothing
+          # new"); anything else non-zero — including argparse's usage-error
+          # exit 2 — must fail the job.
           set +e
           uv run python -m every_eval_ever.cron run "${{ matrix.adapter }}" \
             --work-dir "${{ runner.temp }}/cron-work" \
             --summary "${{ runner.temp }}/cron-work/summary.json" \
-            --raw-repo-id "${{ vars.EEE_RAW_REPO_ID || 'evaleval/EEE_raw' }}" \
+            --raw-repo-id "$EEE_RAW_REPO_ID" \
             --run-id "${{ github.run_id }}-${{ github.run_attempt }}" \
             --run-url "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" \
             ${{ inputs.dry_run && '--dry-run' || '' }} \
             ${{ inputs.force && '--force' || '' }}
           code=$?
           set -e
-          case "$code" in
-            0) echo 'published=true' >> "$GITHUB_OUTPUT" ;;
-            2) echo 'published=false' >> "$GITHUB_OUTPUT" ;;
-            *) echo 'published=false' >> "$GITHUB_OUTPUT"; exit "$code" ;;
-          esac
+          if [ "$code" -ne 0 ] && [ "$code" -ne 3 ]; then
+            exit "$code"
+          fi
 
       - name: Summarise the run
         if: always()
--- a/.github/workflows/adapter_cron.yml
+++ b/.github/workflows/adapter_cron.yml
@@ -1,9 +1,10 @@
 name: Daily adapter refresh
 
-# Refreshes each source adapter once a day and sends its records to that
-# adapter's own datastore pull request. One matrix job per adapter, so each has
-# its own 6-hour job budget, its own failure, and its own PR: a slow or broken
-# adapter cannot hold up the others.
+# Refreshes each source adapter once a day, stores what it fetched in the
+# private raw dataset, and sends its records to that adapter's own datastore
+# pull request. One matrix job per adapter, so each has its own 6-hour job
+# budget, its own failure, and its own PR: a slow or broken adapter cannot hold
+# up the others.
 #
 # Design notes and how to add an adapter: every_eval_ever/cron/README.md
 
@@ -100,14 +101,6 @@ jobs:
           UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
         run: uv sync --locked --all-extras
 
-      - name: Restore the previous run fingerprint
-        uses: actions/cache/restore@v4
-        with:
-          path: ${{ runner.temp }}/fingerprint
-          key: eee-cron-fingerprint-${{ matrix.adapter }}-${{ github.run_id }}
-          restore-keys: |
-            eee-cron-fingerprint-${{ matrix.adapter }}-
-
       - name: Refresh ${{ matrix.adapter }}
         id: refresh
         env:
@@ -122,8 +115,9 @@ jobs:
           set +e
           uv run python -m every_eval_ever.cron run "${{ matrix.adapter }}" \
             --work-dir "${{ runner.temp }}/cron-work" \
-            --fingerprint "${{ runner.temp }}/fingerprint/${{ matrix.adapter }}" \
             --summary "${{ runner.temp }}/cron-work/summary.json" \
+            --raw-repo-id "${{ vars.EEE_RAW_REPO_ID || 'evaleval/EEE_raw' }}" \
+            --run-id "${{ github.run_id }}-${{ github.run_attempt }}" \
             --run-url "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" \
             ${{ inputs.dry_run && '--dry-run' || '' }} \
             ${{ inputs.force && '--force' || '' }}
@@ -145,10 +139,12 @@ jobs:
             echo
             jq -r '"- Status: **\(.status)** — \(.detail)",
                    "- Records: \(.records) in \(.collections | join(", "))",
-                   "- Raw payloads archived: \(.raw_payloads) (\(.raw_bytes) bytes, policy \(.raw_policy))",
+                   "- Raw payloads captured: \(.raw_payloads) (\(.raw_bytes) bytes, policy \(.raw_policy))",
+                   "- Raw archive: \(.raw_archive.status)" +
+                     (if .raw_archive.ledger_path then " → `\(.raw_archive.repo_id)` `\(.raw_archive.ledger_path)` (\(.raw_archive.uploaded) new, \(.raw_archive.reused) reused)" else "" end),
                    "- Fingerprint source: \(.fingerprint_source // "n/a")"' "$summary"
-            jq -r 'if (.defaulted_fields | length) > 0
-                   then "- Defaulted to unknown: " + ([.defaulted_fields | to_entries[] | "\(.key) (\(.value))"] | join(", "))
+            jq -r 'if (.unknown_inferred_fields | length) > 0
+                   then "- Inferred axes still unknown: " + ([.unknown_inferred_fields | to_entries[] | "\(.key) (\(.value))"] | join(", "))
                    else empty end' "$summary"
             jq -r 'if .pr_url then "- Pull request: \(.pr_url)" else empty end' "$summary"
           } >> "$GITHUB_STEP_SUMMARY"
@@ -160,7 +156,9 @@ jobs:
                    "invocation(s) failed; published \(.records) record(s)"
                  else empty end' "$summary"
 
-      - name: Upload raw data, reports, and summary
+      # A short-lived convenience copy. Permanent retention is the raw dataset;
+      # this is here so a failed run's reports are one click away.
+      - name: Upload reports and summary
         if: always()
         uses: actions/upload-artifact@v4
         with:
@@ -171,10 +169,3 @@ jobs:
             ${{ runner.temp }}/cron-work/summary.json
           if-no-files-found: warn
           retention-days: 90
-
-      - name: Store the fingerprint for tomorrow
-        if: steps.refresh.outputs.published == 'true' && !inputs.dry_run
-        uses: actions/cache/save@v4
-        with:
-          path: ${{ runner.temp }}/fingerprint
-          key: eee-cron-fingerprint-${{ matrix.adapter }}-${{ github.run_id }}
--- a/.github/workflows/adapter_cron.yml
+++ b/.github/workflows/adapter_cron.yml
@@ -49,6 +49,19 @@ jobs:
         env:
           UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
         run: uv sync --locked
+      # Runs first because the refresh fails closed: a token that cannot store
+      # raw data would otherwise fail every adapter at its last step.
+      - name: Check credentials and destinations
+        env:
+          UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+          ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
+          LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
+        run: |
+          uv run python -m every_eval_ever.cron preflight \
+            --raw-repo-id "${{ vars.EEE_RAW_REPO_ID || 'evaleval/EEE_raw' }}" \
+            --markdown "$GITHUB_STEP_SUMMARY"
+
       - name: Decide which adapters to refresh
         id: plan
         env:
--- a/.github/workflows/adapter_cron.yml
+++ b/.github/workflows/adapter_cron.yml
@@ -69,10 +69,6 @@ jobs:
         env:
           UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
           REQUESTED: ${{ inputs.adapter }}
-          # Presence-only: the schedule uses these to tell a credentialed
-          # adapter apart from an unconfigured one.
-          ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
-          LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
         run: |
           set -euo pipefail
           if [ -n "${REQUESTED}" ]; then
@@ -162,7 +158,7 @@ jobs:
             echo
             jq -r '"- Status: **\(.status)** — \(.detail)",
                    "- Records: \(.records) in \(.collections | join(", "))",
-                   "- Raw payloads captured: \(.raw_payloads) (\(.raw_bytes) bytes, policy \(.raw_policy))",
+                   "- Raw payloads captured: \(.raw_payloads) (\(.raw_bytes) bytes, policy \(.raw_policy); \(.raw_skipped) over-ceiling, \(.capture_errors | length) capture error(s))",
                    "- Raw archive: \(.raw_archive.status)" +
                      (if .raw_archive.ledger_path then " → `\(.raw_archive.repo_id)` `\(.raw_archive.ledger_path)` (\(.raw_archive.uploaded) new, \(.raw_archive.reused) reused)" else "" end),
                    "- Fingerprint source: \(.fingerprint_source // "n/a")"' "$summary"
@@ -179,16 +175,15 @@ jobs:
                    "invocation(s) failed; published \(.records) record(s)"
                  else empty end' "$summary"
 
-      # Reports and the run summary only — never cron-work/raw: artifacts on a
-      # public repository are downloadable by anyone signed in, and raw source
-      # bodies belong solely in the private raw dataset.
-      - name: Upload reports and summary
+      # The run summary only. Raw bodies live solely in the private raw
+      # dataset, and adapter failure reports go there too — they embed raw
+      # source rows, and artifacts on a public repository are downloadable by
+      # anyone signed in.
+      - name: Upload the run summary
         if: always()
         uses: actions/upload-artifact@v4
         with:
           name: cron-${{ matrix.adapter }}-${{ github.run_number }}
-          path: |
-            ${{ runner.temp }}/cron-work/adapter_reports/
-            ${{ runner.temp }}/cron-work/summary.json
+          path: ${{ runner.temp }}/cron-work/summary.json
           if-no-files-found: warn
           retention-days: 90
--- a/.github/workflows/adapter_cron.yml
+++ b/.github/workflows/adapter_cron.yml
@@ -1,0 +1,180 @@
+name: Daily adapter refresh
+
+# Refreshes each source adapter once a day and sends its records to that
+# adapter's own datastore pull request. One matrix job per adapter, so each has
+# its own 6-hour job budget, its own failure, and its own PR: a slow or broken
+# adapter cannot hold up the others.
+#
+# Design notes and how to add an adapter: every_eval_ever/cron/README.md
+
+on:
+  schedule:
+    # 04:17 UTC. Off the hour, where scheduled-workflow queues are shortest.
+    - cron: '17 4 * * *'
+  workflow_dispatch:
+    inputs:
+      adapter:
+        description: 'Single adapter to refresh (blank = every scheduled adapter)'
+        required: false
+        type: string
+      dry_run:
+        description: 'Run everything except the datastore commit'
+        type: boolean
+        default: false
+      force:
+        description: 'Publish even if the source has not changed since the last run'
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+concurrency:
+  # Never let two refreshes of the same adapter race for its pull request.
+  group: adapter-cron
+  cancel-in-progress: false
+
+jobs:
+  plan:
+    name: Plan
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    outputs:
+      adapters: ${{ steps.plan.outputs.adapters }}
+    steps:
+      - uses: actions/checkout@v6.0.2
+      - uses: astral-sh/setup-uv@v7.6.0
+      - name: Install dependencies
+        env:
+          UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
+        run: uv sync --locked
+      - name: Decide which adapters to refresh
+        id: plan
+        env:
+          UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
+          # Present here only so the schedule can tell a credentialed adapter
+          # apart from an unconfigured one.
+          ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
+          LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
+          REQUESTED: ${{ inputs.adapter }}
+        run: |
+          set -euo pipefail
+          if [ -n "${REQUESTED}" ]; then
+            printf 'adapters=["%s"]\n' "${REQUESTED}" >> "$GITHUB_OUTPUT"
+            echo "Refreshing only ${REQUESTED} (manual dispatch)." >> "$GITHUB_STEP_SUMMARY"
+            exit 0
+          fi
+          uv run python -m every_eval_ever.cron list > plan.json
+          jq -c '.adapters' plan.json | sed 's/^/adapters=/' >> "$GITHUB_OUTPUT"
+          {
+            echo '## Planned refresh'
+            echo
+            echo "Adapters: $(jq -r '.adapters | join(", ")' plan.json)"
+            echo
+            echo '### Not run today'
+            jq -r '.skipped | to_entries[] | "- **\(.key)**: \(.value)"' plan.json
+          } >> "$GITHUB_STEP_SUMMARY"
+
+  refresh:
+    name: ${{ matrix.adapter }}
+    needs: plan
+    if: needs.plan.outputs.adapters != '[]'
+    runs-on: ubuntu-latest
+    # Well under the 6-hour job ceiling: an adapter that hangs should fail
+    # today rather than burn a runner and collide with tomorrow's run.
+    timeout-minutes: 120
+    strategy:
+      fail-fast: false
+      # Raise this to widen the daily wave; it only exists so a refresh leaves
+      # runner capacity for pull request CI.
+      max-parallel: 6
+      matrix:
+        adapter: ${{ fromJSON(needs.plan.outputs.adapters) }}
+    steps:
+      - uses: actions/checkout@v6.0.2
+      - uses: astral-sh/setup-uv@v7.6.0
+      - name: Install dependencies
+        # --all-extras because three adapters read HuggingFace datasets through
+        # `datasets`, which reaches the lock only via the helm extra.
+        env:
+          UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
+        run: uv sync --locked --all-extras
+
+      - name: Restore the previous run fingerprint
+        uses: actions/cache/restore@v4
+        with:
+          path: ${{ runner.temp }}/fingerprint
+          key: eee-cron-fingerprint-${{ matrix.adapter }}-${{ github.run_id }}
+          restore-keys: |
+            eee-cron-fingerprint-${{ matrix.adapter }}-
+
+      - name: Refresh ${{ matrix.adapter }}
+        id: refresh
+        env:
+          UV_PROJECT_ENVIRONMENT: ${{ runner.temp }}/eee-venv
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+          ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
+          LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
+        run: |
+          set -uo pipefail
+          # The default shell is `bash -e`, which would abort the step before the
+          # exit code can be read. Exit 2 is a successful no-op, not a failure.
+          set +e
+          uv run python -m every_eval_ever.cron run "${{ matrix.adapter }}" \
+            --work-dir "${{ runner.temp }}/cron-work" \
+            --fingerprint "${{ runner.temp }}/fingerprint/${{ matrix.adapter }}" \
+            --summary "${{ runner.temp }}/cron-work/summary.json" \
+            --run-url "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" \
+            ${{ inputs.dry_run && '--dry-run' || '' }} \
+            ${{ inputs.force && '--force' || '' }}
+          code=$?
+          set -e
+          case "$code" in
+            0) echo 'published=true' >> "$GITHUB_OUTPUT" ;;
+            2) echo 'published=false' >> "$GITHUB_OUTPUT" ;;
+            *) echo 'published=false' >> "$GITHUB_OUTPUT"; exit "$code" ;;
+          esac
+
+      - name: Summarise the run
+        if: always()
+        run: |
+          summary="${{ runner.temp }}/cron-work/summary.json"
+          [ -f "$summary" ] || { echo "No summary written." >> "$GITHUB_STEP_SUMMARY"; exit 0; }
+          {
+            echo "## ${{ matrix.adapter }}"
+            echo
+            jq -r '"- Status: **\(.status)** — \(.detail)",
+                   "- Records: \(.records) in \(.collections | join(", "))",
+                   "- Raw payloads archived: \(.raw_payloads) (\(.raw_bytes) bytes, policy \(.raw_policy))",
+                   "- Fingerprint source: \(.fingerprint_source // "n/a")"' "$summary"
+            jq -r 'if (.defaulted_fields | length) > 0
+                   then "- Defaulted to unknown: " + ([.defaulted_fields | to_entries[] | "\(.key) (\(.value))"] | join(", "))
+                   else empty end' "$summary"
+            jq -r 'if .pr_url then "- Pull request: \(.pr_url)" else empty end' "$summary"
+          } >> "$GITHUB_STEP_SUMMARY"
+          # A partial refresh publishes real data, so the job is green. Annotate
+          # it anyway, or a source that half-converts every day goes unnoticed.
+          jq -r 'if (.failed_invocations | length) > 0 then
+                   "::warning title=Partial refresh::\(.adapter): " +
+                   "\(.failed_invocations | length) of \(.invocations) " +
+                   "invocation(s) failed; published \(.records) record(s)"
+                 else empty end' "$summary"
+
+      - name: Upload raw data, reports, and summary
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: cron-${{ matrix.adapter }}-${{ github.run_number }}
+          path: |
+            ${{ runner.temp }}/cron-work/raw/
+            ${{ runner.temp }}/cron-work/adapter_reports/
+            ${{ runner.temp }}/cron-work/summary.json
+          if-no-files-found: warn
+          retention-days: 90
+
+      - name: Store the fingerprint for tomorrow
+        if: steps.refresh.outputs.published == 'true' && !inputs.dry_run
+        uses: actions/cache/save@v4
+        with:
+          path: ${{ runner.temp }}/fingerprint
+          key: eee-cron-fingerprint-${{ matrix.adapter }}-${{ github.run_id }}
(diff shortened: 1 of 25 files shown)


every_eval_ever/cron/stamp.py
```diff
@@ -0,0 +1,197 @@
...
+    Instance-level ``*_samples.jsonl`` companions carry no ``source_metadata``;
+    they are reachable from the aggregate that references them, which is what
+    the stamp goes on.
+    """
+    data_root = Path(root)
+    return sorted(
+        path
+        for path in data_root.glob('*/*/*/*.json')
+        if not path.name.endswith('_samples.json')
+    )
```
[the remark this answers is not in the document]
evijit: Fixed in c0ac615f8 — and you're right that it was misleading, though not in the way it looks.

The glob is `*/*/*/*.json`, so `.jsonl` companions were never matched in the first place: the `_samples.json` filter was dead code, and the docstring described a guarantee it wasn't providing. The fragility is real, though — this package uses `*.json*` elsewhere (`runner.validate_records`, and the validator's own documented patterns), so widening the glob here is a plausible edit that would have silently started stamping companions.

It now matches on the stem instead of the full filename, which holds either way:

```python
if path.suffix == '.json' and not path.stem.endswith(SAMPLES_SUFFIX)
```

Added `test_a_companion_named_with_a_json_extension_is_not_an_aggregate` (covering the `_samples.json` case the old filter was nominally for) and `test_stamping_a_tree_leaves_companions_untouched`.

every_eval_ever/cron/runner.py
```diff
@@ -0,0 +1,682 @@
...
+            'arguments': ' '.join(item.arguments) or '(no arguments)',
+            'returncode': item.returncode,
+        }
+        for item in outcome.failed
+    ]
+
+    raw_capture.index_unlisted_payloads(raw_dir)
+    manifest = raw_capture.read_manifest(raw_dir)
+    summary.raw_payloads = sum(1 for entry in manifest if entry.get('file'))
+    summary.raw_bytes = sum(entry.get('bytes') or 0 for entry in manifest)
```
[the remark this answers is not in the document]
evijit: Good catch — this was a real bug, fixed in c0ac615f8.

`archive()` deliberately writes a ledger row for an oversized payload, and `test_a_payload_too_large_to_store_still_gets_a_ledger_row` asserts exactly that. But it calls `archive()` directly, so it passed while the runner's guard made the case unreachable: `raw_payloads` counted only entries with a local `file`, so a run whose only payload was over the ceiling returned `nothing_captured` and never archived. The documented behaviour ("a row is written even for a payload too large to store") was therefore false end to end.

The guard now asks whether anything was *captured* rather than whether anything *landed*:

```python
summary.raw_archive = _archive_raw(raw_dir, captured=len(manifest), ...)
```

Also: `raw_skipped` now reports fetched-but-not-stored, `raw_bytes` counts only stored bytes (it previously summed skipped payloads too, so bytes could be non-zero while payloads was 0), and there's a warning when a payload is dropped for size. `test_an_oversized_payload_is_still_recorded_in_the_ledger` covers the runner path the unit test missed.

Writing that test also surfaced that `publish` and the ledger read weren't in the no-network fixture, so a missing stub could make a live Hub call — all three entry points are stubbed by default now.

every_eval_ever/cron/preflight.py
```diff
@@ -0,0 +1,174 @@
...
+
+    Creating it here rather than mid-refresh means a token that cannot create
+    repositories is reported before any adapter has run.
+    """
+    try:
+        info = api.repo_info(repo_id=repo_id, repo_type='dataset')
+    except Exception:
+        info = None
+
+    if info is not None:
```
mrshu, line 109: **Existing public raw datasets pass preflight**

`check_raw_dataset()` reports success for a public repository. A refresh can then upload permanent raw payloads where anyone can read them.

**What to change**

- Make public visibility a required preflight failure.
- Verify that the raw repository is private again immediately before each archive commit.
- Add focused tests that public visibility blocks both preflight and archive.

every_eval_ever/helpers/raw_capture.py
```diff
@@ -0,0 +1,278 @@
...
+        _logger.warning(
+            'not capturing %s: %d bytes exceeds the %d byte ceiling',
+            url,
+            len(body),
+            ceiling,
+        )
+        _record(directory, filename, checksum, entry)
+        return None
+
+    directory.mkdir(parents=True, exist_ok=True)
```
mrshu, line 189: **Changed responses from one URL overwrite earlier captured bytes**

`_capture()` replaces the URL-derived file but keeps both checksums in the manifest. The archive can store later bytes under an earlier checksum path.

**What to change**

- Keep distinct local payloads for each URL identity and checksum so every manifest row resolves to matching bytes.
- Add a capture-to-archive test for two different bodies from one URL and equal bodies from two URLs.

every_eval_ever/cron/runner.py
```diff
@@ -0,0 +1,743 @@
...
+
+    stamped = stamp_tree(
+        data_root,
+        adapter=adapter.name,
+        run_date=run_date,
+        run_url=run_url,
+    )
+    summary.unknown_inferred_fields = stamped.unknown_inferred
+
+    previous = _previous_fingerprint(
```
mrshu: **Archived fingerprints can suppress first and failed publications**

`refresh()` archives the current fingerprint before it reads the latest ledger. A run can compare against itself, and an archived run that fails publication can block its next retry.

**What to change**

- Compare against durable last-successful-publication state while keeping raw archiving before datastore publication.
- Add shared-backend tests for first, unchanged, changed, and publish-failure-then-retry runs.

every_eval_ever/cron/preflight.py
```diff
@@ -0,0 +1,174 @@
...
+    name = identity.get('name') or identity.get('fullname') or 'unknown'
+    role = (identity.get('auth') or {}).get('accessToken', {}).get('role')
+    return Check(
+        'hugging face token',
+        True,
+        f'authenticated as {name}' + (f' (role: {role})' if role else ''),
+    )
+
+
+def check_datastore(api: HfApi, repo_id: str) -> Check:
```
mrshu, line 70: **Datastore preflight proves read access only**

`check_datastore()` calls only `repo_info()`. A readable token can pass planning, then every job that reaches pull-request publication can fail late.

**What to change**

- Verify the exact pull-request publication capability during preflight without creating repository artifacts.
- Add a test for a token that can read the datastore but cannot publish through the intended pull-request path.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — Comment-only review; three confirmed high-priority findings need attention.
**Result:** The isolation and provenance structure is promising. Four confirmed issues are worth addressing before rollout.
**Scope:** Add a daily adapter-refresh workflow with permanent raw capture, change gating, validation, stamping, and datastore PR publication.
**Second check:** 3 reviewers checked 4 findings and 4 plans; 4 findings kept, 3 plans clarified, and 1 implementation path set aside.

## Earlier review comments

- **Fixed** — The aggregate companion filter now excludes `*_samples.jsonl` through the `_samples` stem check. [Earlier thread](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3761339302)
- **Still present** — `raw_payloads` counts stored files while `raw_skipped` separately counts oversized manifest rows. This remains in the earlier thread, so this review does not open a duplicate. [Earlier thread](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3761339342)

## What I noticed

|---|---|---|---|---|
| RAV-RUN1-R1-F001 | high | raw-data-privacy | `every_eval_ever/cron/preflight.py:82` | An existing public raw dataset passes preflight, so refreshes can upload permanent source payloads publicly. |
| RAV-RUN1-R1-F002 | high | raw-provenance | `every_eval_ever/helpers/raw_capture.py:166` | Changed responses from one URL overwrite earlier local bytes while the manifest retains both checksums. |
| RAV-RUN1-R1-F003 | high | change-gating | `every_eval_ever/cron/runner.py:457` | Archived fingerprints can suppress a first publication and retries after a publication failure. |
| RAV-RUN1-R1-F004 | medium | rollout-preflight | `every_eval_ever/cron/preflight.py:56` | Datastore preflight proves read access only, so jobs can fail late when PR publication starts. |

## Things to try

- **[high] raw-data-privacy** — Block public raw repositories during preflight and immediately before archive commits. (`RAV-RUN1-R1-P001`; covers `RAV-RUN1-R1-F001`)
- **[high] raw-provenance** — Store each URL and checksum pair separately, then verify archive content against its ledger checksum. (`RAV-RUN1-R1-P002`; covers `RAV-RUN1-R1-F002`)
- **[high] change-gating** — Gate on durable last-successful-publication state while keeping raw archiving before datastore publication. (`RAV-RUN1-R1-P003`; covers `RAV-RUN1-R1-F003`)

- **RAV-RUN1-R1-P004 [medium] rollout-preflight** — The concrete permission-check method is set aside because the pinned Hub client lacks a suitable non-mutating write check. The required outcome remains actionable.

<details>

- Target: PR #247 (`feature/daily-adapter-cron`, 24 files, +5183/-0)
- Reviewed head: `c0ac615f83f21a60aa6f035a144aef5bda160a1f`
- Run ordinal: 1
- Rounds: 1/1 completed; adaptive review off
- Mix: 3 codex-exec
- Reviewer signal: 1 substantive review; 2 reviewers could not access the exact head and were recorded as failed
- Focus: correctness, maintainability, simplicity, production blast-radius, and constructive suggestion language
- Earlier review comments: 2 comments; 1 fixed and 1 still present
- Finding counts: 0 critical, 3 high, 1 medium, 0 low, 0 nit
- Checks: concerns=4; confirmed=3, narrowed=1, ruled-out=0, set-aside=0
- Second check: full; reviewers=3; findings kept=4; plans clarified=3; implementation paths set aside=1
- Action-lock: 2 independent auditors; 4/4 inline bodies passed both audits

</details>

---

every_eval_ever/cron/preflight.py
evijit: Agreed, and fixed in c48087138 — I'd wrongly made this a courtesy warning. Now:

- `check_raw_dataset` **fails** preflight when the raw dataset is public (and after creating one, reads it back and refuses if it doesn't read back private). A transient `repo_info` error is also no longer treated as "repo missing" — creating or blessing anything on the basis of a 500 could mask a public dataset behind a green check; it now fails with "could not read".
- `archive()` re-checks visibility via `_require_private()` immediately before **every** commit, not only at preflight, and raises before anything is uploaded. Since archiving happens before publishing and fails closed, a repo flipped public mid-stream stops the whole refresh.
- Tests: `test_a_public_raw_dataset_fails_preflight`, `test_a_transient_error_is_not_treated_as_a_missing_raw_dataset`, `test_a_public_raw_dataset_is_refused` (archive-level, asserts zero commits).

Still deliberately not done: flipping visibility automatically — reporting is the cron's job, changing a repo's visibility is a human's.

every_eval_ever/helpers/raw_capture.py
evijit: Fixed in c48087138. The capture filename now includes a content digest alongside the URL digest, so one URL serving two bodies keeps **both** payloads on disk with a manifest row each pointing at exactly the bytes it hashed; refetching identical bytes stays idempotent (same name, one row). This removed the in-process dedupe memo too — dedupe now reads the on-disk manifest, so a cleared-and-reused capture directory starts genuinely empty.

Both tests you asked for exist: `test_one_url_serving_two_bodies_keeps_both` + `test_one_url_serving_two_bodies_archives_both_correctly` (capture→archive, asserts each committed blob's bytes hash to the address it's stored under), and `test_two_urls_sharing_a_last_segment_do_not_collide` covers equal-ish names from two URLs.

every_eval_ever/cron/runner.py
evijit: Fixed in c48087138 — and this was the most serious finding on the PR; my own review pass converged on the same three failure modes (self-comparison, zero-capture adapters never persisting state, and failed-publish-then-skip losing records).

The gate no longer reads the ledger at all. It reads `state/<adapter>.json` in the raw dataset, written **only after a successful datastore commit**, which gives exactly the property you named: raw archiving stays before publication (fail-closed provenance), while the comparison target is durable last-successful-publication state. A partial publish records `partial: true`, and the next run treats that as "publish regardless" so records that failed conversion get retried.

Shared-backend tests as requested: the runner test fixture now shares a state store between consecutive `refresh()` calls instead of stubbing the gate (the stubbing is what had hidden this), covering first (`test_a_run_never_compares_against_its_own_fingerprint`), unchanged (`test_the_previous_fingerprint_comes_from_the_publish_state`), changed (`test_a_changed_source_publishes_again`), and publish-failure-then-retry (`test_a_failed_publish_does_not_update_the_state` + `test_the_run_after_a_failed_publish_publishes`). A two-run simulation against the live hle source publishes once, then skips.

every_eval_ever/cron/preflight.py
evijit: Partially addressed in c48087138, with the limit stated rather than papered over.

`check_token` now fails preflight when the token's role is definitively read-only (`whoami().auth.accessToken.role == 'read'`) — that's the common misconfiguration and it's knowable up front. What I did *not* find is a way to prove the PR-publication capability itself without side effects: the Hub has no dry-run commit, and probing by opening/closing a discussion would leave artifacts on the datastore, which your note rules out. Fine-grained token scopes are visible in `whoami()` but interpreting repo-level grants reliably across token shapes felt like more false confidence than signal, so a fine-grained token passes with its identity reported and the first real commit remains the proof.

Test: `test_a_read_only_token_fails_preflight`. If you know a side-effect-free capability probe I missed, very happy to add it.

.github/workflows/adapter_cron.yml
```diff
@@ -0,0 +1,181 @@
...
+  # adapter's requires_env name must appear here (tests enforce it).
+  HF_TOKEN: ${{ secrets.HF_TOKEN }}
+  ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
+  LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
+  EEE_RAW_REPO_ID: ${{ vars.EEE_RAW_REPO_ID || 'evaleval/EEE_raw' }}
+
+concurrency:
+  # Never let two refreshes of the same adapter race for its pull request.
+  group: adapter-cron
+  cancel-in-progress: false
```
mrshu: **Static workflow concurrency can replace an unrelated pending refresh**

The `adapter-cron` group covers every workflow run. GitHub keeps one running and one pending item, then replaces the older pending item when another run arrives.

**What to change**

- Move concurrency to the refresh job and key it by workflow, `matrix.adapter`, and dry-run versus publish mode while keeping `cancel-in-progress: false`.
- Update the workflow test to require adapter-keyed job concurrency and reject the workflow-wide static group.

.github/workflows/adapter_cron.yml
```diff
@@ -0,0 +1,181 @@
...
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: cron-${{ matrix.adapter }}-${{ github.run_number }}
+          path: |
+            ${{ runner.temp }}/cron-work/raw/
+            ${{ runner.temp }}/cron-work/adapter_reports/
+            ${{ runner.temp }}/cron-work/summary.json
+          if-no-files-found: warn
+          retention-days: 90
```
mrshu: **The workflow artifact exposes raw response bodies outside the private dataset**

The public repository uploads `cron-work/raw` on every outcome and keeps it for 90 days. Signed-in repository readers can download those captured responses.

**What to change**

- Remove `cron-work/raw` from the artifact while keeping non-sensitive reports and summaries.
- Update the workflow test to reject raw capture paths in `upload-artifact` inputs.

every_eval_ever/cron/archive.py
```diff
@@ -0,0 +1,357 @@
...
+    except (RepositoryNotFoundError, EntryNotFoundError):
+        return None
+    except Exception as error:
+        _logger.warning(
+            'could not read %s from %s (%s); treating this run as new',
+            state_path(adapter),
+            repo_id,
+            error,
+        )
+        return None
```
mrshu: **Unreadable state can republish an unchanged record set**

`read_state()` returns `None` for download, permission, parse, and state-shape failures. The runner treats that result as a first run and can publish the full set under new UUIDv4 paths.

**What to change**

- Treat only a confirmed missing state entry as first-run state, and raise `ArchiveError` for other read or state-shape failures.
- Add transient-download and malformed-state tests that prove `publish()` is not called.

every_eval_ever/cron/runner.py
```diff
@@ -0,0 +1,793 @@
...
+    state = archive_module.read_state(adapter, repo_id=raw_repo_id, token=token)
+    if not state:
+        return None
+    if state.get('partial'):
+        _logger.info(
+            '%s: previous publish was partial; publishing regardless of the '
+            'fingerprint so the missing records get another attempt',
+            adapter,
+        )
+        return None
```
mrshu: **_previous_fingerprint() disables gating after every partial publication**

A successful partial publication writes partial true. On every later run, _previous_fingerprint returns None whenever that flag remains true. Identical partial output therefore bypasses comparison and publishes again under new UUIDv4 paths.

**What to change**

Retry partial conversions on every run. Persist a separate canonical output fingerprint and failed-invocation identity for partial publications. Skip publication only when both the successful output and failure outcome match the prior partial run. Publish when either changes. Add tests for identical, changed, and fully recovered partial runs.

every_eval_ever/cron/runner.py
```diff
@@ -0,0 +1,793 @@
...
+        in {RawPolicy.VIA_FETCH_HELPERS, RawPolicy.VIA_ADAPTER_FLAG}
+        and not manifest
+    ):
+        # Declared as archived but nothing was captured at all: report it rather
+        # than let the run look like it saved raw data.
+        _logger.warning(
+            '%s declares raw policy %s but captured no payloads',
+            adapter.name,
+            adapter.raw_policy.value,
+        )
```
mrshu: **A capture failure can still publish records without archived source bytes**

`capture_response()` swallows capture exceptions. The runner only warns about an empty manifest, and one successful capture can also hide another failed capture.

**What to change**

- Add a durable capture-error signal across the adapter subprocess boundary and stop publication after any declared-policy capture error or unexplained empty manifest.
- Add runner tests for capture exceptions, empty declared captures, mixed-success captures, and deliberate oversized ledger-only rows.

`NOT_CAPTURED`, `UPSTREAM_VERSIONED`, and deliberate oversized ledger-only rows can remain unchanged.

.github/workflows/adapter_cron.yml
```diff
@@ -0,0 +1,181 @@
...
+permissions:
+  contents: read
+
+env:
+  # One place for the credentials both planning and refreshing need. An
+  # adapter's requires_env name must appear here (tests enforce it).
+  HF_TOKEN: ${{ secrets.HF_TOKEN }}
+  ARTIFICIAL_ANALYSIS_API_KEY: ${{ secrets.ARTIFICIAL_ANALYSIS_API_KEY }}
+  LLM_STATS_API_KEY: ${{ secrets.LLM_STATS_API_KEY }}
+  EEE_RAW_REPO_ID: ${{ vars.EEE_RAW_REPO_ID || 'evaleval/EEE_raw' }}
```
mrshu, line 36: **Workflow-level environment values expose every secret to unrelated steps**

`HF_TOKEN` and both adapter API keys reach checkout, setup, install, summary, artifact upload, and every matrix adapter. Those steps do not need all three credentials.

**What to change**

- Pass `HF_TOKEN` only to preflight and refresh commands.
- Pass each adapter API key only to preflight, planning, and the matching adapter refresh.
- Give no secrets to checkout, setup, install, summary, or artifact steps.
- Update the workflow test to verify command-level credential coverage and reject workflow-level secret exposure.

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — The revision is a strong step forward, with two high-priority and four medium-priority suggestions worth addressing.
**Result:** The new head fixes most earlier findings. This round found focused follow-ups in raw-data privacy, provenance, state gating, workflow isolation, and secret scope.
**Scope:** This PR adds a daily adapter refresh subsystem with raw capture, change gating, validation, datastore publication, and isolated scheduled jobs.
**Checks:** 7 concerns checked; 3 confirmed, 3 narrowed, and 1 lowered in priority.
**Second check:** Full Codex review with 3 reviewers; 3 findings kept, 4 clarified, 1 clarified carry-forward set aside, and 0 removed.

## Earlier review comments

<details>
<summary>Earlier review comments (9 items)</summary>

- The aggregate companion filter is fixed on this head.
- Raw payload and skipped-payload counts are now separate.
- `RAV-RUN1-R1-F001` is fixed by preflight and archive privacy checks.
- `RAV-RUN1-R1-F002` is fixed by checksum-addressed capture files.
- `RAV-RUN1-R1-F003` is fixed by post-publication state persistence.
- `RAV-RUN1-R1-F004` remains a narrower rollout limitation and is set aside below without a duplicate inline thread.
- `RAV-RUN1-R1-P001` is complete.
- `RAV-RUN1-R1-P002` is partly complete. Separate checksum-addressed files are present; archive checksum verification remains outside this round's actionable findings.
- `RAV-RUN1-R1-P003` is complete.

</details>

## What I noticed

|---|---|---|---|---|
| RAV-RUN2-R1-F002 | high | raw-data-privacy | `.github/workflows/adapter_cron.yml:171` | The public workflow artifact includes raw response bodies that the design otherwise stores privately. (inline) |
| RAV-RUN2-R1-F005 | high | raw-provenance | `every_eval_ever/cron/runner.py:403` | A capture exception or hidden mixed capture failure can still allow publication without archived source bytes. (inline) |
| RAV-RUN2-R1-F001 | medium | workflow-isolation | `.github/workflows/adapter_cron.yml:41` | One static concurrency group can serialize unrelated runs and replace an older pending refresh. (inline) |
| RAV-RUN2-R1-F003 | medium | change-gating | `every_eval_ever/cron/archive.py:310` | An unreadable or malformed state file is treated as a first run and can republish unchanged records. (inline) |
| RAV-RUN2-R1-F004 | medium | partial-refresh | `every_eval_ever/cron/runner.py:257` | An identical persistent partial result republishes its successful subset under new UUIDv4 paths. (inline) |
| RAV-RUN2-R1-F006 | medium | secret-scope | `.github/workflows/adapter_cron.yml:33` | Workflow-level secrets reach setup, install, summary, artifact, and unrelated adapter steps. (inline) |

<details>

- **[high] raw-data-privacy** — Keep raw captures out of public workflow artifacts while preserving safe reports and summaries. (`RAV-RUN2-R1-P002`; covers `RAV-RUN2-R1-F002`)
- **[high] raw-provenance** — Carry capture failures across the subprocess boundary and stop publication when a declared capture policy fails. (`RAV-RUN2-R1-P005`; covers `RAV-RUN2-R1-F005`)
- **[medium] workflow-isolation** — Scope concurrency to the refresh job and separate adapter and run-mode queues. (`RAV-RUN2-R1-P001`; covers `RAV-RUN2-R1-F001`)
- **[medium] change-gating** — Treat only confirmed missing state as a first run and reject other state errors. (`RAV-RUN2-R1-P003`; covers `RAV-RUN2-R1-F003`)
- **[medium] partial-refresh** — Compare durable partial output and failure identity without suppressing recovered conversions. (`RAV-RUN2-R1-P004`; covers `RAV-RUN2-R1-F004`)
- **[medium] secret-scope** — Pass each secret only to the command steps that need it. (`RAV-RUN2-R1-P006`; covers `RAV-RUN2-R1-F006`)

</details>

<details>

- Target: PR #247 (`feature/daily-adapter-cron`, 25 files, +5670/-7), head `c4808713813737d2f5d2a588cdaffc60e9b5f937`
- Run ordinal: 2
- Mix: 3 codex-exec
- Focus: correctness, maintainability, simplicity, production blast radius, and constructive positive suggestions
- Earlier review comments: 9 items; 7 fixed, 1 partly fixed, 1 narrower carry-forward set aside
- Finding counts: 0 critical, 2 high, 4 medium, 0 low, 0 nit
- Checks: concerns=7; confirmed=3, narrowed=3, lowered=1; none ruled out
- Second check: full; reviewers=3; kept=3, clarified=4, set-aside=1, removed=0; approval unchanged
- Set aside: 1 item; reason=no reliable non-mutating capability probe in the pinned client

</details>

---

---

### Earlier review comments

- **RAV-RUN1-R1-F004 [medium] rollout-preflight** — The current client can prove reachability and reject a definite read-only role, but it has no proven non-mutating write-capability probe. Keep the documented limit and use a controlled rollout canary rather than adding a speculative probe. _(This is still present. Source: https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762294429; id=RAV-RUN1-R1-F004)_

.github/workflows/adapter_cron.yml
evijit: Fixed in 53257e14e — and thank you, this was the worst finding of the round: it quietly undid the private-dataset design for 90 days per run. The artifact now carries `adapter_reports/` and `summary.json` only; `test_raw_payloads_never_reach_a_public_artifact` rejects any raw path in `upload-artifact` inputs.

every_eval_ever/cron/runner.py
evijit: Fixed in 53257e14e. `capture_response` now writes a durable **error row** into the manifest when a capture fails (the log line dies with the subprocess; the manifest crosses the boundary). The runner then fails before publishing when a declared-capture adapter produced records alongside any capture error, or with an empty manifest — one successful capture no longer hides a failed one.

All four requested tests exist: `test_a_capture_failure_stops_publication` (mixed success — asserts the state is not advanced either), `test_records_without_any_captured_payload_stop_publication`, `test_a_capture_failure_leaves_a_durable_error_row` + `test_error_rows_do_not_count_as_captured_payloads` (capture level), and `test_an_oversized_only_capture_still_publishes` (deliberate ledger-only rows proceed, as you specified). `NOT_CAPTURED`/`UPSTREAM_VERSIONED` are unaffected.

.github/workflows/adapter_cron.yml
evijit: Fixed in 53257e14e. Concurrency is now on the refresh job, keyed `adapter-cron-${{ matrix.adapter }}-${{ dry-run|publish }}` with `cancel-in-progress: false` kept — the actual invariant is "two publishes of one adapter never race for its PR", and the workflow-wide group was both over-serializing and, as you note, silently replacing the older pending run. Tests: `test_two_publishes_of_one_adapter_never_run_at_once`, `test_dry_runs_do_not_queue_behind_publishes`, `test_no_workflow_wide_concurrency_group`.

every_eval_ever/cron/archive.py
evijit: Fixed in 53257e14e — you're right that my "fail open is safe" reasoning only holds for *confirmed-missing* state. `read_state` now returns `None` solely on `RepositoryNotFoundError`/`EntryNotFoundError` and raises `ArchiveError` for transient download errors, unreadable files, malformed JSON, and non-object payloads — the run fails and retries tomorrow rather than republishing the set. Tests: `test_a_transient_state_read_error_raises`, `test_a_malformed_state_file_raises` (and the runner's failure path is covered by the shared-store fixture, which asserts publish is never reached when the gate raises).

every_eval_ever/cron/runner.py
evijit: Fixed in 53257e14e, implementing exactly the shape you proposed. Partial state now records a `failure_identity` — an order-independent digest of which invocations failed and with what exit codes — alongside the gating fingerprint. `_unchanged_since` skips a partial-successor run only when **both** match; any change on either side (source moved, failure recovered, new failure appeared) publishes. So a persistently half-broken source costs one publish, not one per day, while recovery is never suppressed — a recovered invocation's records change the output fingerprint, which forces the publish.

Tests: `test_an_identical_partial_run_is_not_republished`, `test_a_changed_failure_set_publishes_again`, `test_a_recovered_partial_run_publishes`.

.github/workflows/adapter_cron.yml
evijit: Fixed in 53257e14e. No secrets at workflow scope any more: preflight and the plan step carry the adapter keys (they report credential coverage), each matrix job receives `HF_TOKEN` plus only its own adapter's key via `${{ matrix.adapter == '...' && secrets.X || '' }}`, and checkout/setup/install/summary/artifact steps get none. `EEE_RAW_REPO_ID` (not a secret) stays workflow-level.

Tests as requested: `test_no_secrets_at_workflow_scope`, `test_secrets_are_scoped_away_from_non_command_steps`, and `test_every_declared_adapter_credential_reaches_the_command_steps` keeps the schedule's `requires_env` tied to the step env blocks so the scoping can't silently strand an adapter. (This deliberately reverses part of an earlier "hoist the env" cleanup — least privilege wins.)

mrshu (commented):
# ⚒️ review-anvil report

**Review decision:** COMMENT — Constructive suggestions remain for privacy, retry safety, credential scope, and refresh reliability.
**Result:** The daily refresh design is well covered by tests; these focused changes would make failure recovery and operational boundaries safer.
**Scope:** Add a daily adapter matrix that archives raw source data, skips unchanged runs, validates records, and publishes each adapter to its datastore PR.
**Checks:** 6 concerns checked; 3 confirmed, 1 clarified, and 2 lowered in priority.
**Second check:** 2 Codex reviewers checked 9 findings; 7 kept, 1 clarified, and 1 set aside for a policy decision.

<details>
<summary>Earlier review comments (12 coalesced topics)</summary>

- **Fixed:** Companion detection now excludes `_samples` stems. [Earlier thread](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3761339302)
- **Still present as low-priority guidance:** `raw_skipped` is durable but absent from the two main reviewer-facing summaries. [Earlier thread](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3761339342)
- **Fixed:** Raw-dataset visibility is checked during preflight and before archive commits. [Finding](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762294418) · [Earlier report](https://github.com/evaleval/every_eval_ever/pull/247#pullrequestreview-4911423812)
- **Fixed:** Capture and blob paths are content-addressed. [Finding](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762294423) · [Earlier report](https://github.com/evaleval/every_eval_ever/pull/247#pullrequestreview-4911423812)
- **Fixed:** Change gating uses state written only after successful publication. [Finding](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762294425) · [Earlier report](https://github.com/evaleval/every_eval_ever/pull/247#pullrequestreview-4911423812)
- **Still present:** Datastore preflight proves read access, not repository-specific write authority. [Earlier thread](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762294429)
- **Fixed:** Concurrency is scoped by adapter and run mode. [Finding](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762564998) · [Earlier report](https://github.com/evaleval/every_eval_ever/pull/247#pullrequestreview-4911738032)
- **Still present:** Public artifacts can include raw source rows from failure reports. [Finding](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762565001) · [Earlier report](https://github.com/evaleval/every_eval_ever/pull/247#pullrequestreview-4911738032)
- **Fixed:** Only confirmed missing state is treated as a first run. [Finding](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762565004) · [Earlier report](https://github.com/evaleval/every_eval_ever/pull/247#pullrequestreview-4911738032)
- **Fixed:** Partial refresh state includes output and failure identity. [Finding](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762565009) · [Earlier report](https://github.com/evaleval/every_eval_ever/pull/247#pullrequestreview-4911738032)
- **Fixed for publication:** Capture errors stop publication. This run found a separate durability gap for successful sibling captures. [Finding](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762565011) · [Earlier report](https://github.com/evaleval/every_eval_ever/pull/247#pullrequestreview-4911738032)
- **Fixed at workflow scope:** Secrets are step-scoped. This run found a separate child-process scope issue. [Finding](https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762565014) · [Earlier report](https://github.com/evaleval/every_eval_ever/pull/247#pullrequestreview-4911738032)

</details>

## What I noticed

ID legend: `RUN` is the observed PR review run, `R` is the origin round, `F` is a finding, and `P` is a plan.

| ID | Priority | Topic | Code location | Constructive suggestion |
|---|---|---|---|---|
| RAV-RUN3-R1-F001 | high | publish-retry | `every_eval_ever/cron/publish.py:136` | Make a later-batch retry reconcile only its incomplete attempt, so earlier completed refreshes stay intact and records do not duplicate. (inline) |
| RAV-RUN3-R1-F002 | high | secret-scope | `every_eval_ever/cron/runner.py:151` | Keep datastore and raw-dataset write authority in the cron parent instead of every adapter subprocess. (inline) |
| RAV-RUN3-R1-F004 | medium | raw-provenance | `every_eval_ever/cron/runner.py:466` | Archive successful sibling captures and capture-error evidence before returning failure. Keep publication blocked. (inline) |
| RAV-RUN3-R1-F005 | medium | credential-monitoring | `every_eval_ever/cron/preflight.py:163` | Give an enabled adapter with missing credentials an isolated failing signal while other configured adapters continue. (inline) |
| RAV-RUN3-R1-F007 | medium | state-persistence | `every_eval_ever/cron/runner.py:620` | Return a visible failure if publication state cannot be saved, while retaining the published PR URL for reconciliation. (inline) |

<details>

- **RAV-RUN3-R1-F006 [low] error-visibility** — Consider showing `raw_skipped` in the workflow Step Summary and datastore publication description. Detailed reasons can remain in `summary.json` and the private ledger.

</details>

## Things to try

<details>
<summary>Suggested implementation groups (8 items)</summary>

- **[high] raw-data-privacy** — Generate a dedicated public report with allowlisted counts, references, and safe reason categories. Test the complete uploaded directory. (`RAV-RUN2-R1-P002`; covers `RAV-RUN2-R1-F002`)
- **[high] publish-retry** — Track attempt-specific paths or progress, then resume or remove only the incomplete attempt. (`RAV-RUN3-R1-P001`; covers `RAV-RUN3-R1-F001`)
- **[high] secret-scope** — Build child environments without the parent write token. Give explicit read-only source credentials only where required. (`RAV-RUN3-R1-P002`; covers `RAV-RUN3-R1-F002`)
- **[medium] raw-provenance** — Keep capture-error metadata in the private ledger and archive available evidence before returning failure. (`RAV-RUN3-R1-P004`; covers `RAV-RUN3-R1-F004`)
- **[medium] credential-monitoring** — Keep enabled adapters visible to the matrix and fail only each uncredentialed adapter job. (`RAV-RUN3-R1-P005`; covers `RAV-RUN3-R1-F005`)
- **[low] error-visibility** — Render the existing `raw_skipped` count on both reviewer-facing summaries. (`RAV-RUN3-R1-P006`; covers `RAV-RUN3-R1-F006`)
- **[medium] state-persistence** — Retry state persistence, then return a distinct failure with the PR URL if it still fails. (`RAV-RUN3-R1-P007`; covers `RAV-RUN3-R1-F007`)
- **[medium] rollout-preflight** — Use authoritative target-repository scope metadata when available. Treat uncertain fine-grained authority as a rollout prerequisite. (`RAV-RUN3-R1-P008`; covers `RAV-RUN1-R1-F004`)

</details>

- **RAV-RUN3-R1-F003 [medium] raw-provenance** — The code deliberately permits oversized payloads to leave only ledger metadata, while broader PR prose promises permanent raw bytes. Please choose the intended ceiling policy and storage route before making this path fail closed.

<details>

- Target: PR #247 (`feature/daily-adapter-cron`, 25 files, +6079/−7) at `53257e14e18e5f2132408bc06a36b74465a3e44e`
- Run ordinal: 3
- Mix: 3 Codex (`codex-exec`) reviewers
- Focus: correctness, production blast radius, simplicity, and maintainability
- Earlier review comments: 21 ledger entries coalesced into 12 root causes; 9 fixed and 3 still present
- Finding counts: 0 critical, 3 high, 5 medium, 1 low, 0 nit
- Checks: concerns=6; confirmed=3, clarified=1, lowered=2
- Second check: targeted; reviewers=2 Codex; kept=7, clarified=1, set aside=1, removed=0
- Review event: COMMENT (user-requested comment-only output)

</details>

---

---

### Earlier review comments

| RAV-RUN2-R1-F002 | high | raw-data-privacy | `.github/workflows/adapter_cron.yml:185` | Keep raw source values out of public artifacts by generating an allowlisted public report. Preserve full local/private failure provenance. | _(This is still present. Source: https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762565001; id=RAV-RUN2-R1-F002)_
| RAV-RUN1-R1-F004 | medium | rollout-preflight | `every_eval_ever/cron/preflight.py:70` | Validate datastore-specific write authority when token metadata proves it. Treat ambiguous fine-grained access as inconclusive and use a controlled rollout check. | _(This is still present. Source: https://github.com/evaleval/every_eval_ever/pull/247#discussion_r3762294429; id=RAV-RUN1-R1-F004)_

every_eval_ever/cron/publish.py
```diff
@@ -0,0 +1,214 @@
...
+
+    root = Path(data_root)
+    api = api or HfApi(token=token)
+    existing = find_open_pr(api, repo_id, adapter)
+    batches = _batched(files, max(1, files_per_commit))
+    total = len(batches)
+
+    pr_number = existing
+    pr_url = None
+    for index, batch in enumerate(batches, start=1):
```
mrshu, line 191: **A later batch failure duplicates records on retry**

`publish()` commits each batch at once. If a later batch fails, a fresh run reuses the PR and adds the full set under new UUID4 paths.

**What to change**

- Track one attempt's batch progress, or remove only that incomplete attempt's paths before restarting, while keeping earlier completed refreshes unchanged.
- Add a test that fails a later batch, retries from a fresh work directory against one persistent PR tree, and verifies one copy of each logical record.

every_eval_ever/cron/runner.py
```diff
@@ -0,0 +1,855 @@
...
+    directory.
+
+    A non-zero exit is not treated as fatal here. An adapter that hits a source
+    row it cannot represent writes every valid record, writes a report under
+    ``adapter_reports/``, and *then* exits non-zero — so a run is reported,
+    never discarded, and one bad leaderboard does not cost the others their
+    records.
+    """
+    work_dir.mkdir(parents=True, exist_ok=True)
+    child_environment = {
```
mrshu, line 163: **Adapter subprocesses inherit datastore write credentials**

`run_adapter()` copies the cron parent's full environment. Adapter code and dependencies therefore receive `HF_TOKEN` authority for the datastore and private raw dataset. Scheduled adapters otherwise declare only adapter-specific source credentials.

**What to change**

- Exclude the write-capable `HF_TOKEN` from child environments while keeping it in the parent for archive, state, and publication calls.
- Give an adapter a separate least-privilege read-only token only when it explicitly needs authenticated Hugging Face source access.
- Add a child-environment test that rejects datastore write credentials and requires source tokens to be explicit, read-only, and otherwise least-privileged.

every_eval_ever/cron/runner.py
```diff
@@ -0,0 +1,855 @@
...
+        if summary.capture_errors:
+            problems.append(
+                f'{len(summary.capture_errors)} capture failure(s): '
+                + ', '.join(summary.capture_errors[:5])
+            )
+        if not manifest:
+            problems.append(
+                f'raw policy {adapter.raw_policy.value} captured nothing'
+            )
+        if problems:
```
mrshu: **Capture failures discard successful sibling provenance**

`refresh()` returns before `_archive_raw()` when one declared capture fails. Successful sibling bytes and the error evidence remain only in the ephemeral work directory, which the workflow does not upload.

**What to change**

- Preserve capture-error metadata in `archive.ledger_rows()` and archive successful captures and capture-error rows before returning failure, while keeping datastore publication blocked.
- Extend the mixed-capture test to verify permanent archiving and no publication.

every_eval_ever/cron/preflight.py
```diff
@@ -0,0 +1,220 @@
...
+    reason ``scheduled_adapters`` formats — rewording a message must never make
+    a credential problem invisible.
+    """
+    runnable = []
+    held_back: list[Check] = []
+    for adapter in CRON_ADAPTERS:
+        if not adapter.enabled:
+            continue
+        missing = adapter.missing_env(environment)
+        if missing:
```
mrshu, line 163: **Missing credentials remove enabled adapters without a failing signal**

Preflight classifies a missing enabled-adapter credential as a non-required warning. `scheduled_adapters()` removes that adapter from the matrix. The Step Summary names the omission, but the workflow stays green. A direct request also reports a successful no-op.

**What to change**

- Keep enabled adapters in the matrix and give each uncredentialed adapter an isolated failing result, while allowing unrelated configured adapters to run.
- Reserve quiet skips for adapters declared disabled.
- Add scheduled and manual tests for an enabled adapter with missing required credentials.

every_eval_ever/cron/runner.py
```diff
@@ -0,0 +1,855 @@
...
+                    'run_url': run_url,
+                    'partial': bool(summary.failed_invocations),
+                    'failure_identity': current_failures,
+                    'pr_number': result.pr_number,
+                    'pr_url': result.pr_url,
+                },
+                repo_id=raw_repo_id,
+                token=environment.get('HF_TOKEN'),
+            )
+        except archive_module.ArchiveError as error:
```
mrshu: **State-write failures still report a published success**

After `write_state()` raises `ArchiveError`, `refresh()` appends a warning and returns `EXIT_PUBLISHED`. The summary already has the PR URL, but the stale gate can let the next run publish the same logical set again.

**What to change**

- Retry the state write within the run and, if it still fails, write a distinct failed summary with the published PR URL before returning a nonzero result.
- Add a test that makes `write_state()` fail and verifies the nonzero result and retained publication URL.

every_eval_ever/cron/publish.py
evijit: Fixed in 62d27b447, with the durable-progress shape you proposed. Before the first batch, the run writes `state/<adapter>.attempt.json` in the raw dataset listing exactly the paths it is about to commit. A later run that finds a dangling attempt hands those paths to `publish(stale_paths=...)`, which deletes whichever of them actually reached the PR (existence-checked via one `get_paths_info` call against the PR ref) in the same commit as the first new batch — so earlier *completed* refreshes are untouched and the retry replaces the dead attempt instead of stacking beside it. A dangling attempt also forces a publish even when the fingerprint matches, since half a dead attempt is sitting on the PR. The attempt is cleared atomically with the state write (same repo, same commit), so the two can never disagree.

Your exact test exists: `test_a_failed_later_batch_retried_leaves_one_copy_of_each_record` fails batch 2 of 3 against a persistent fake PR tree, retries from a fresh work directory with fresh filenames, and asserts the final tree holds exactly the five retry files and nothing from the dead attempt. Plus `test_stale_paths_absent_from_the_pr_are_not_deleted`, `test_stale_paths_without_an_open_pr_are_ignored`, and runner-level `test_a_failed_publish_leaves_the_attempt_for_reconciliation` / `test_a_dangling_attempt_forces_a_publish_even_when_unchanged`. Also verified with a live three-run simulation against the real hle source.

every_eval_ever/cron/runner.py
evijit: Fixed in 62d27b447. `run_adapter` scrubs `HF_TOKEN`, `HUGGING_FACE_HUB_TOKEN` and `HF_HUB_TOKEN` from the child environment — every spelling the hub client reads. Archiving, state and publication all happen in the parent, which is the only holder of write authority. For a source that genuinely needs authenticated HF *read* access, the schedule gains a `source_hf_token` flag: the child then receives the separate `EEE_SOURCE_HF_TOKEN` (a least-privilege read token an operator provisions) as its `HF_TOKEN` — never the cron's own. No scheduled adapter currently needs it, so nothing changes for them beyond losing authority they never should have had.

Tests: `test_the_write_token_never_reaches_the_adapter_subprocess` (asserts all three names are absent while source keys and ordinary variables pass through) and `test_a_source_hf_token_is_forwarded_only_when_declared` (asserts the forwarded value is the read token, and that undeclared adapters get nothing).

every_eval_ever/cron/runner.py
evijit: Fixed in 62d27b447. The capture-failure check now runs *after* `_archive_raw`, so by the time the run returns failure, the successful sibling captures, the error rows (now carried into `ledger_rows()` as an `error` field) and the adapter's failure reports are already permanent. Publication stays blocked, and the state is not advanced.

The mixed-capture test is extended as requested: `test_capture_failure_evidence_is_archived_before_the_run_fails` asserts the archive call happened, that its ledger rows contain both the error row and the successful sibling, that the failure report was included, and that nothing was published. `test_ledger_rows_carry_capture_errors` covers the ledger field itself.

every_eval_ever/cron/preflight.py
evijit: Fixed in 62d27b447, exactly along the lines you drew. `scheduled_adapters` no longer filters by environment — every enabled adapter stays in the matrix — and `refresh()` fails a run with missing credentials (`EXIT_FAILED`, status `missing_credentials`, detail naming the variables), so the workflow shows one isolated red job per uncredentialed adapter while configured adapters proceed (`fail-fast: false` already guarantees the isolation). A manual dispatch of a single uncredentialed adapter fails the same way rather than reporting a green no-op. Quiet skips are now reserved for adapters deliberately declared `enabled=False`.

Tests: `test_missing_credentials_fail_the_adapters_own_run` (covers both the scheduled and direct-request paths, since both go through `refresh`), `test_uncredentialed_enabled_adapters_stay_scheduled`, `test_disabled_adapters_are_never_scheduled`.

every_eval_ever/cron/runner.py
evijit: Fixed in 62d27b447. `write_state` is retried once; if it still fails, the run returns `EXIT_FAILED` with status `published_state_unrecorded` and the PR URL retained in both `summary.pr_url` and the detail text — the records are on the PR, only the gate is stale, and a green exit would have hidden that. The detail also says what happens next: the attempt record is still standing (it is only cleared in the same commit as the state), so the next run reconciles this exact attempt rather than blindly stacking a duplicate — the F001 machinery closes the loop you flagged.

Tests: `test_a_persistent_state_write_failure_fails_the_run_keeping_the_pr` (nonzero + URL retained) and `test_a_transient_state_write_failure_is_retried` (one failure then success stays green).

evijit:
Closing, see #249

PR #258, opened 2026-08-18
fix(schema): do not invent max attempts

### LINKED ISSUE #236: generation_args.max_attempts defaults to 1, so records assert an attempt count no source stated
`generation_args.max_attempts` has a schema default of `1`, so **every record with a
`generation_config` claims the run allowed one submission attempt, whether or not the
source said anything about attempts.** Should the default be `null`?

It is the only field in `eval.schema.json` with a non-null default. It is not required.
Because pydantic serialises defaults, it materialises in output even when no converter
touched it — this is committed lm_eval output, converted from a log that never mentions
attempts:

```
$ uv run every_eval_ever convert lm_eval \
    --log_path tests/data/lm_eval/results_2026-01-21T03-44-18.458309.json --output_dir /tmp/out/data
$ grep -o '"max_attempts": [0-9]*' /tmp/out/data/*/*/*/*.json
"max_attempts": 1
"max_attempts": 1
```

Suggested fix: `"default": 1` → `"default": null` in the `generation_args` subschema, then
regenerate `eval_types.py`. The field stays optional, and its absence then means "the
source did not say" instead of "one attempt".

@nelaturuharsha — worth doing? It changes the output of every converter that builds a
`generation_config`, so it wants a maintainer's call rather than a drive-by PR. Happy to
open that PR if you agree.

<details>
<summary>Where it comes from, and why it reads as data</summary>

The concept is Inspect's: `max_attempts` is how many times a solver may submit an answer.
Only the inspect converter ever observes a real value, and even there it is uncertain —
`converters/inspect/adapter.py:407` carries `# TODO not sure if max_attempts == max_retries
in this case`.

For lm_eval, HELM, lighteval and every adapter, nothing in the source corresponds to it, so
the `1` is not a measurement. A consumer filtering on `max_attempts == 1` today selects
records where the value was never observed, together with the Inspect records where it was.

Found while writing the lighteval converter, where the same `1` appears in every record.
</details>

reacher-z:
## What / source

Changes the optional generation_args.max_attempts schema default from 1 to null and regenerates eval_types.py. Sources that do not report an attempt limit no longer materialize a claim that one attempt was allowed. Explicitly observed Inspect values remain unchanged.

Fixes #236.

## Review lane

- [ ] Fast
- [x] Needs a human — this is a material output semantics change across converters.

Design agreed in: #236, including collaborator confirmation that this is worth doing.

## Checklist

- [x] Validation is clean — 2 lm-eval aggregate records and 1 Inspect aggregate record, with 0 warnings
- [x] Offline regression tests added; full test suite green: 759 passed, 20 skipped
- [x] Ruff check clean
- [x] Content spot-checked: absent lm-eval value is omitted; explicit Inspect value remains 1

The adapter report, registry, and source-row checklist items are not applicable because this changes schema default serialization and drops no source records.

## Decisions & coverage

- Decision / where: schema_version
  Chose / instead of: keep the current version because the field remains optional and accepts the same integer values; only its non-required default annotation changes
  Confidence: high
  General: no
- Decision / where: generated model
  Chose / instead of: regenerate eval_types.py from the schema and run post_codegen, rather than hand-editing generated behavior
  Confidence: high
  General: yes
- Decision / where: explicit values
  Chose / instead of: preserve source-provided Inspect max_attempts values while omitting unobserved values
  Confidence: high
  General: yes

Coverage: 2 lm-eval source tasks to 2 valid records, 0 dropped; 1 Inspect source run to 1 valid record, 0 dropped. The lm-eval outputs contain no max_attempts key; the Inspect fixture that explicitly reports 1 still emits 1.

Commands run:

- uv run ruff check every_eval_ever tests
- env -u NO_COLOR TERM=xterm-256color uv run pytest tests -q
- uv run --extra inspect pytest tests/test_inspect_adapter.py -q
- real lm-eval and Inspect CLI conversions followed by validation at final collection/developer/model paths

## AI assistance

OpenAI Codex assisted with implementation, tests, and validation. I reviewed the diff and the generated output.

### DIFF
--- a/every_eval_ever/eval_types.py
+++ b/every_eval_ever/eval_types.py
@@ -232,7 +232,7 @@ class GenerationArgs(BaseModel):
     )
     sandbox: Sandbox | None = None
     max_attempts: int | None = Field(
-        1, description='Maximum number of submission attempts (default 1).'
+        None, description='Maximum number of submission attempts.'
     )
     incorrect_attempt_feedback: str | None = Field(
         None, description='Feedback from the model after incorrect attempt.'
--- a/every_eval_ever/schemas/eval.schema.json
+++ b/every_eval_ever/schemas/eval.schema.json
@@ -531,8 +531,8 @@
                                     },
                                     "max_attempts": {
                                         "type": "integer",
-                                        "description": "Maximum number of submission attempts (default 1).",
-                                        "default": 1
+                                        "description": "Maximum number of submission attempts.",
+                                        "default": null
                                     },
                                     "incorrect_attempt_feedback": {
                                         "type": "string",
--- a/tests/test_inspect_adapter.py
+++ b/tests/test_inspect_adapter.py
@@ -338,6 +338,10 @@ def test_gaia_eval():
     assert converted_eval.model_info.inference_platform == 'openai'
     assert converted_eval.model_info.inference_engine is None
 
+    generation_config = converted_eval.evaluation_results[0].generation_config
+    assert generation_config is not None
+    assert generation_config.generation_args.max_attempts == 1
+
     results = converted_eval.evaluation_results
     assert len(results) > 0
     assert results[0].evaluation_name == 'gaia'
--- a/tests/test_lm_eval_adapter.py
+++ b/tests/test_lm_eval_adapter.py
@@ -237,6 +237,10 @@ def test_transform_from_file_generation_config():
     assert gen is not None
     assert gen.generation_args.temperature == 0.0
     assert gen.generation_args.max_tokens == 512
+    assert gen.generation_args.max_attempts is None
+    assert 'max_attempts' not in gen.generation_args.model_dump(
+        mode='json', exclude_none=True
+    )
     assert gen.additional_details['num_fewshot'] == '0'
 
 


borgr:
CI was never run here either — the workflow was queued as `action_required` for a first-time contributor. Approved, and all four rows are green.

This closes #236. The change is right: a record carrying `max_attempts: 1` asserted an attempt count no source stated, and the two new tests pin both directions — Inspect keeps a real 1 it read from the log, lm-eval omits the field entirely rather than inventing it.

Flagging the lane rather than the content: this edits `eval.schema.json` and the generated types, and it changes what existing converters emit, so by CONTRIBUTING's review policy it wants a human read and not only an agent pass — the diff is four files but the blast radius is every record with a `generation_args`. Labelled `auto-review` so the pass happens too.

borgr:
Rebased onto current main so this can be evaluated. The only conflict was the datamodel-codegen header in `every_eval_ever/eval_types.py`, where main no longer carries a `#   timestamp:` line and this branch had regenerated one. Took main's headerless form; the schema change itself is untouched.

Local run on the rebased branch: 1335 passed, 1 skipped, ruff clean.

The change reads correctly to me. `max_attempts` defaulting to 1 asserted a single-attempt protocol for every eval whose source never reported one, and the two adapter tests now pin it to `None` for inputs that say nothing about retries.

borgr (approved):
Reviewed and approving.

The change is right. `max_attempts` defaulting to 1 asserted a single-attempt protocol for every eval whose source never reported one, which is the same class of claim as an invented metric bound — a schema default silently becoming published data. `None` is the honest value for "the source did not say", and the two adapter tests pin it for inputs that mention no retries.

`"default": null` in the JSON schema is redundant with the field being optional, but it is explicit about the intent and costs nothing, so I left it.

Rebased onto current main (the only conflict was the datamodel-codegen header timestamp, which main no longer carries). All five checks green on `ce60b218f`, and locally 1335 passed, 1 skipped, ruff clean.

Issue #256, opened 2026-08-14
Update changed cron records instead of adding new files


tommasocerruti:
The cron skips unchanged records, but when a record changes, it adds a new file and leaves the old one in the datastore. Over time, this can create multiple versions of the same evaluation and unnecessarily grow the datastore.

Could cron records use a stable identity and update the existing file when their content changes? 
(Git history would still preserve the previous version)

evijit:
We should in general find a deduplication method.

nelaturuharsha:
(1) We agree on for immutability of records available on the datastore. We however, need a retention policy and then a cold storage where people who use it can download it, before finally deleting the same.

(2) You're right that this history of change is preserved via git - but in view of the above, I would not modify a record in place, but treat it as changed. We have no yet guaranteed full immutability due to functional reasons. 

(3) It also brings in a good discussion for when the schema changes and moves forward

(4) We should be careful about what we do for de-duplication. We need a grounded method to decide what entails such a thing - and shouldn't rush through designing this, as we might miss out on data we can include otherwise, we've discussed before on slack.

borgr:
**The blocker is enforced, not conventional, and this turns out to gate every data correction rather than only datastore growth.**

`converters.common.publication.publish_evaluation_logs` calls `require_uuid4` on every filename it is handed. So a record *cannot* be published under a name derived from its identity — the publisher rejects it. That is why a changed cron record lands beside the old one instead of replacing it.

**What that costs, measured.** I sampled the datastore for a separate reason and the number is stark. Of **51,802** aggregate records on `main`, **26,611 (51.4%)** carry a `metric_id` that cannot identify a metric — missing, generic, or an exact copy of `evaluation_name`. Four adapters produce 23,373 of them, all four on the daily cron. Method was three records per collection, which settled 98 of the 107 collections because one adapter writes every record in a collection, plus a full scan of the 9 that disagreed.

Fixing those four adapters today would publish ~20,702 corrected records **beside** the wrong ones. The datastore would then hold both, the wrong half would still fail the gate, and the correction would have made the count worse before better. The same applies to #287, which corrects declared metric bounds in three cron adapters.

So this is not only about unnecessary growth. **It is the mechanism by which any adapter fix reaches published data**, and without it every correction is a duplication.

## Shape I would propose

- A **cron** record's filename derives from its identity — collection, developer, model, and the evaluation identity within the record — so a re-run overwrites in place and git history keeps the previous version, which is what the issue already suggests.
[...]

PR #260, opened 2026-08-19
[Feature] Add reviewed PwC DrugBank protocol adapter

yananlong:
## Summary

This PR adds a manual-only adapter that converts selected DrugBank leaderboard cells from a local Papers With Code custom-format PostgreSQL dump into EEE aggregate evaluation logs.

A reviewed YAML qualification manifest identifies the exact source cells to convert and supplies reviewed model and benchmark identities, source-scale decisions, split and protocol semantics, provenance, and the registry revision used during review. Canonical metric metadata — including metric identity, kind, unit, direction, and canonical bounds — is resolved and checked against the same vendored `eval-card-registry` snapshot used by the general Papers With Code adapter. Observed metric ranges are derived from the source values being converted rather than from the manifest.

The manifest distinguishes `transductive` (both drugs seen), `inductive-s1` (both drugs unseen), and `inductive-s2` (one drug unseen), following the split convention described by the [[DSN-DDI primary source](https://academic.oup.com/bib/article/24/1/bbac597/6966537)](https://academic.oup.com/bib/article/24/1/bbac597/6966537). Scores themselves come from the pinned Papers With Code dump; the manifest supplies source-cell selection and protocol evidence.

## Design and safety

The adapter parses the dump strictly and fails closed on dump-hash drift, missing or ambiguous anchors, malformed scores, incompatible source scales, registry-revision drift, unknown or ambiguous registry metrics, duplicate YAML keys, duplicate metric selections, conflicting benchmark protocols, coercive source IDs, or invalid output paths.

Conversion is atomic: the reviewed selection is treated as one qualified bundle, and no output is published unless the complete bundle passes registry, schema, and path validation.

Aggregate and result identities include the selected metric results, canonical metric configuration, and protocol semantics. Output records include `split_id`, the generalization regime, entity-overlap evidence, and the protocol digest. DrugBank remains the trusted `source_data` value, while the raw Papers With Code URL and archive URL are retained as provenance metadata.

Papers With Code `±` values are preserved without coercing an unspecified dispersion type into a typed standard-error field. When score rescaling applies, the source-reported dispersion and its canonical-scale value remain distinguishable.

## Relationship to the general Papers With Code adapter

The general Papers With Code adapter can emit the same underlying evaluation rows. `pwc_evaluation_id` identifies the shared PwC evaluation row, while an exact source-cell match uses:

`(pwc_evaluation_id, metric_config.metric_name)`

This PR makes that overlap explicit and joinable without attempting to solve global cross-collection deduplication. Broader deduplication policy remains part of #256 rather than expanding the scope of this adapter.

## Scope and follow-up

No production qualification manifest or generated datastore records are included. The adapter remains `runnable=False` because it requires local inputs and has no live fetch path.

The adapter converts the source cells selected by the reviewed manifest; it does not independently verify the cited paper content or reconstruct train/test membership from the aggregate Papers With Code dump, and it does not compute performance deltas between splits.

## Verification

* The GitHub Actions `Tests` workflow passes on the current PR head across Ruff, core/full, and locked/loose configurations.
* Registry-contract tests cover the pinned registry revision, canonical metric parity with the general Papers With Code adapter, fail-closed handling of unknown or ambiguous metrics, and source-derived observed ranges.
* Integration coverage exercises a real `pgdumplib` custom-format dump through the adapter CLI path.
* Source-scale mismatches fail closed instead of being guessed.
* No dependency or lockfile changes are included.

### DIFF
--- a/every_eval_ever/adapters/README.md
+++ b/every_eval_ever/adapters/README.md
@@ -34,8 +34,8 @@ An adapter that automation runs must therefore:
   `save_failure_report` + a non-zero exit, which is what lets a partial refresh be
   told apart from a crash.
 
-`bfcl`, `cocoabench` and `sciarena` are registered as `runnable=False`: they need a
-local input file and have no live fetch path.
+`bfcl`, `cocoabench`, `paperswithcode_drugbank`, and `sciarena` are registered as
+`runnable=False`: they need local inputs and have no live fetch path.
 
 ### Skipping a run whose source has not changed
 
@@ -90,12 +90,46 @@ re-hosting bytes that are already durably stored.
 | `terminal_bench_2` | tbench.ai | Fetches Terminal-Bench 2.0 agentic coding benchmark results. |
 | `hle` | Scale SEAL leaderboard | Converts the Scale SEAL Humanity's Last Exam leaderboard into `data/hle/`. Emits per-model accuracy (with 95% CI) and calibration error. |
 | `mmlu_pro` | TIGER-Lab leaderboard CSV | Converts the MMLU-Pro leaderboard (`TIGER-Lab/mmlu_pro_leaderboard_submission`) into `data/mmlu-pro/`. Emits per-model overall + 14 per-subject accuracies. |
+| `paperswithcode_drugbank` | Local Papers with Code PostgreSQL dump + reviewed YAML manifest | Manually converts only DrugBank score cells with reviewed model, metric-scale, split, and protocol semantics. Writes `data/paperswithcode-drugbank/`. The same cells also appear in `data/paperswithcode/` from the general adapter over the same dump, unreviewed; both records carry the PwC `pwc_evaluation_id` so the pair is joinable. |
 | `lexam` | LEXam project website | Converts the LEXam legal-reasoning leaderboard (open-question judge scores + 4-choice MCQ accuracy) into `data/lexam/`. |
 | `vectara_hallucination_leaderboard` | HuggingFace (`vectara/results`) | Converts the Vectara Hallucination Leaderboard result files, pinned to a source commit, into `data/vectara-hallucination-leaderboard/`. Emits 4 aggregate metrics plus per-category and per-text-complexity breakdowns (40 scores per model). |
 | `benchpress` | HuggingFace (`microsoft/benchpress-score-matrix`) | Converts the BenchPress score matrix into `data/benchpress/`. An aggregator: every cell keeps its own citation, and logs split by `evaluator_relationship`. See [`benchpress/README.md`](benchpress/README.md). |
 | `bountybench` | BountyBench run logs (local JSON tree) | Converts BountyBench cybersecurity agent logs into `data/bountybench/`. Emits one aggregate per (model, workflow, configuration) plus a per-bounty `*_samples.jsonl` sidecar with the full agent transcript. |
 | `paperswithcode` | Papers with Code PostgreSQL dumps | Converts PwC leaderboard entries into `data/paperswithcode/`. Metric bounds and direction are resolved against a vendored eval-card-registry snapshot; unknown metrics fail the run rather than getting invented bounds. Needs the `paperswithcode` extra. |
 | `wild` | HuggingFace (`kensho/WILD-raw`) | Converts the WILD-raw item-level eval responses (65 models × 27 benchmarks, run with Inspect AI) into `data/wild/`: aggregate accuracy per model×benchmark and per subtask, with optional per-item `_samples.jsonl` sidecars (`--include-instances`). See [`wild/README.md`](wild/README.md). |
+
+### Papers with Code DrugBank
+
+To run this manual adapter, provide a local custom-format PostgreSQL dump and a
+reviewed, non-empty YAML qualification manifest (schema version 2). It is not
+scheduled because
+neither input has a live fetch path, and the repository does not include a
+production manifest. The manifest must supply reviewed canonical model,
+metric, and benchmark IDs plus the registry commit against which they were
+reviewed. The adapter does not resolve or invent identities, and conversion is
+atomic: an anchor, hash, or schema mismatch aborts before any output is written.
+For split comparisons, each manifest entry must declare `transductive`,
+`inductive-s1`, or `inductive-s2`, with matching manifest-declared overlap
+labels. A
+canonical benchmark ID cannot be reused for conflicting protocols anywhere in
+the manifest, so the adapter cannot emit conflicting split scores under one
+benchmark key. The adapter preserves the scores and protocol evidence supplied
+by the reviewed manifest; it does not independently verify cited source content
+or train/test membership from the aggregate dump, and it does not compute a
+performance delta between splits.
+
+```bash
+uv run --extra paperswithcode python -m \
+  every_eval_ever.adapters.paperswithcode_drugbank.adapter \
+  --dump /path/to/paperswithcode.dump \
+  --overlay /path/to/reviewed-drugbank.yaml \
+  --output-dir /tmp/paperswithcode-drugbank/data/paperswithcode-drugbank
+```
+
+As of 2026-08-19, the archived Papers with Code DrugBank URL redirects away
+from its dataset record. Generated logs retain that URL as raw provenance,
+link `source_data` to DrugBank, and record the official Papers with Code archive
+in source metadata.
 
 ### Mercor Evaluation Exports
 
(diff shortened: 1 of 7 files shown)


borgr:
@yananlong this rebases mechanically everywhere except one spot, and that spot is a design question I do not want to answer inside a conflict marker.

`main` now carries the Papers with Code adapter from #209: `every_eval_ever/adapters/paperswithcode/`, collection `paperswithcode`, converting the whole postgres dump from the `huggingface/paperswithcode-backups` bucket weekly, with a `paperswithcode = ["pgdumplib>=4.0.0"]` extra and a vendored eval-card-registry snapshot for metric bounds.

This PR adds a second package over the same dump — `paperswithcode_drugbank`, collection `paperswithcode-drugbank`, extra `paperswithcode-drugbank = ["pgdumplib>=4.0.0,<5.0.0"]`. The conflict is both sides editing the extras block and `all`, and resolving it either way encodes an answer:

1. **Two packages, one extra.** Drop the `paperswithcode-drugbank` extra, depend on `paperswithcode`, keep the packages separate. Cheapest, and leaves two adapters reading one dump.
2. **One adapter, a drugbank mode.** Fold this in as a mode of `adapters/paperswithcode/`, which is where the dump reading, the bucket resolution and the registry snapshot already live.

Also worth settling before either: whether the two overlap on rows. If the general adapter already emits DrugBank protocol leaderboards, the same source rows land in two collections under two identities, and that is a dedup problem rather than a packaging one.

Which do you want? Once that is decided the rest is a `uv lock` regeneration and a one-line table row, and I will do that part.

The `<5.0.0` cap is the other small difference from main's specifier — intentional, or just what was current when you wrote it?

borgr:
Rather than leave this on you: #269 continues these four commits with `main` merged in, the extras conflict resolved, and the overlap question answered rather than left open.

Short version. The duplicate `pgdumplib` extra now depends on main's `paperswithcode` one. And the overlap with the general adapter is joinable, not silent — both sides already record the PwC `evaluations` row id as `additional_details.pwc_evaluation_id`, so I documented that and the key in the docstring and the README row instead of restructuring anything. 951 tests pass, ruff clean, and it is labelled `auto-review` since this PR never got a pass (Copilot hit its quota).

Your `<5.0.0` cap on pgdumplib is the one thing I dropped — if that was deliberate rather than what was current, say so and it goes back.

Leaving this open until #269 lands, so the work is never in limbo. Happy to hand the commits back if you would rather land it from your fork.

borgr:
Pushed to this branch (`c4d0790b5`) rather than opening a second PR. @yananlong's four commits are still the base of the history, with two on top.

**`main` merged in.** The only conflict was the extras block: main gained the general Papers with Code adapter (#209) over the same postgres dump, with `paperswithcode = ["pgdumplib>=4.0.0"]`. This branch declared `paperswithcode-drugbank = ["pgdumplib>=4.0.0,<5.0.0"]` — the same package a second time. It now depends on the existing extra, and the two `uv run --extra` lines point at it. @yananlong, the `<5.0.0` cap is the one thing that went: deliberate, or just what was current? It comes back in a line if deliberate.

**The overlap, stated rather than left implicit.** The scheduled `paperswithcode` run is `--all --best-effort`, so it converts every dataset in the dump — the DrugBank cells included, without the reviewed split and protocol semantics this adapter adds. So the same measurement appears in two collections.

That is joinable, not silent: both adapters record the PwC `evaluations` row id as `additional_details.pwc_evaluation_id` (`paperswithcode/adapter.py:884` here, `paperswithcode_drugbank/adapter.py:692` there), so a consumer reading both can tell one measurement from two. The second commit says so in the adapter docstring and the `adapters/README.md` row. What was missing was the statement, not the key.

If maintainers would rather have one adapter with a drugbank mode, or have the general adapter skip the reviewed cells, that is a larger change and belongs with #256's dedup discussion rather than in this PR.

`uv run pytest tests` 951 passed, 32 skipped; `uv run ruff check` clean. The regenerated `uv.lock` has a zero-line diff against main.

Labelled `auto-review`: this PR never got a review pass — Copilot hit its quota — so that would be a first read, not a re-read.

borgr:
Review — one bug, one divergence, two smaller notes. No blockers on the protocol modelling, which is careful.

**1. The uncertainty is parsed, then dropped and left on the wrong scale.** `parse_metric_value` pulls `±` out of a source string, `_build_result` sets `score_details.uncertainty=None`, and puts the raw figure in `details.reported_uncertainty` unscaled while the score is multiplied by `scale_factor`. Reproduced on this head:

```
'91.2% ± 1.4'  -> score=0.912, details.reported_uncertainty='1.4'
'62.65 ± 0.31' -> score=62.65 (or 0.6265 with source_scale: percent), reported_uncertainty='0.31'
```

A reader of the second field sees a spread wider than the score's whole range. No test covers this path — `uncertainty` and `±` appear nowhere in the 1366-line test file, which is why it survived.

Scaling it is unambiguous. Putting it in the typed field is not: PwC's `±` may be a standard error, a standard deviation or a CI half-width, and `ScoreDetails.uncertainty.standard_error` would assert one. Cleanest is an overlay field declaring the dispersion kind for the reviewed cells, then fill the typed field; failing that, scale the number and name the field for what it is.

**2. Bounds and direction come from the overlay, while the sibling adapter over the same dump takes them from the registry.** `adapters/paperswithcode` resolves `min_score`/`max_score`/`lower_is_better`/`score_type` against a vendored registry snapshot and fails closed on an unknown metric; here they are hand-reviewed YAML fields. Same dump, same metric, two possible answers — and the overlap means both adapters can publish the same PwC row. Suggest resolving through the same snapshot and keeping the overlay for the protocol semantics only, which is what it is uniquely good at.

**3. `export` splits `model_info.id` by hand.** `log.model_info.id.split('/', 1)` raises on a flat id, and bypasses `datastore_path_components`, which is the documented way to derive those components (#271 §path). `valid_model_id` currently guarantees a slash, so this is latent rather than live.

**4. Fail-fast with no accounting is a choice worth stating.** Every rejection raises, so one malformed cell aborts the run and nothing records exclusions. For a small hand-reviewed manifest that is defensible — unlike a scrape — but `SourceConversionResult` is what the skill asks for, so say in the docstring that the manifest is the ledger.

@yananlong — (1) needs a decision from you about what the `±` is; the rest I can push if you'd rather not.

borgr:
Fixed the two that were unambiguous, in `806087e53` — no point leaving them as your to-do when the answer is not in question.

**Uncertainty scaling.** `parse_metric_value` now returns a validated number rather than text, rejecting a negative or non-finite one, and `_build_result` applies the same scale factor it applies to the score. `'99.49 ± 0.31'` on a percent-scale metric now records `score=0.9949` with `reported_uncertainty=0.0031`, and `raw_value` still holds the source's own spelling of both figures. Four tests cover the path, which had none — percent scale, identity scale, no dispersion, and three unusable values.

**The path split.** `export` uses `datastore_path_components` instead of `model_info.id.split('/', 1)`, so a flat id can no longer raise there and the developer folder comes from the one place that owns that rule.

957 passed, 32 skipped; ruff clean.

**Still yours, and still a real question:** `score_details.uncertainty` stays unset. Filling `standard_error` would assert that PwC's `±` is a standard error, and it may be a standard deviation or a CI half-width depending on the paper. If the reviewed cells all have a known dispersion kind, an overlay field is the clean way to say so and then the typed field can be filled. Happy to add it if you tell me what the manifest should carry.

The registry-vs-overlay question for bounds (finding 2) I have deliberately not touched — that one is a design call about two adapters over one dump, not a fix.

borgr:
Follow-up in `421d495a0`, and an answer to my own question about the `±`.

**It is not private knowledge, and we already had the answer.** `adapters/paperswithcode` reads the same dump and documents the same conclusion independently — *"PwC's '±' does not identify the spread as a standard error, a standard deviation, or a CI half-width, so the caller keeps it verbatim rather than coercing it into a typed Uncertainty."* So leaving `score_details.uncertainty` unset is not a gap here, it is the house rule, and this PR already matched it. Nothing to decide.

What is knowable per cell is narrower: your overlay carries `protocol_evidence_url` and `source_locator`, so for a hand-reviewed manifest of a handful of cells, whoever reads the paper can record what its `±` is. Worth doing later if the typed field matters; not a blocker.

**Comparing the two adapters did turn up something, though: the same defect on main.** The general adapter rescales scores onto the registry's canonical scale and leaves the `±` verbatim, so a published record reads `score: 0.337` beside `reported_uncertainty: "0.82"`. #276 fixes that, and settles the convention — `reported_uncertainty` is what the source printed, `reported_uncertainty_canonical` is the same number on the score's scale, present only when a rescale applied.

`421d495a0` moves this PR onto that convention, so the same key does not mean verbatim in one adapter and canonical in the other. 957 passed, 32 skipped; ruff clean.

PR #268, opened 2026-08-26
[Submission] tau-bench leaderboard adapter, review answered (supersedes #192)

borgr:
Supersedes #192, continuing @benshi34's branch rather than replacing it — their `Add tau-bench leaderboard adapter` commit is the base of this history, with `main` merged in and the review answered on top. Opened here because #192's branch lives on a fork I cannot push to.

## Why #192 could not just be updated

That branch was 172 commits behind and predates #218, which moved adapters from `utils/<name>/` to `every_eval_ever/adapters/<name>/`. The merge is a modify/delete collision on a tree that no longer exists, so the first two commits here are the merge (resolved by taking main's deletion of `utils/README.md`) and the relocation.

## The four review-anvil findings

**RAVF001, metric identity (high).** Pass^k was published as `tau_bench.pass_at_k`. Those are different quantities — tau-bench's Pass^k requires all k trials to succeed, pass@k requires one — so a consumer joining on that id would have averaged incompatible scores. Now `tau_bench.pass_hat_k`, with `metric_semantics` in `additional_details` saying which it is. The registry has no `pass-hat-k` entry (only `pass-at-1` and `pass-at-8`, both 0–1), so the source-namespaced id is the correct form here.

**RAVF002, provenance (medium).** A remote run cited `raw.githubusercontent.com/…/main/…`, which serves different bytes next week, and a local replay invented that same URL for bytes read off disk. `pin_base_url` now resolves the ref to a commit once, before anything is fetched, so every URL in the run is immutable and the batch records one `submission_commit`. Every submission also carries the sha256 of its own payload, and a local run records `local_input_path` and drops the invented URL. `--allow-unpinned-source` keeps a run possible when the API cannot be reached, and the record then says the ref was not pinned.

**RAVF003, score validation (medium).** `--input-dir` and `--base-url` accept inputs the leaderboard never serves. Pass^k is bounded 0–100 and cost at 0 or above, booleans and non-finite values are rejected through `require_finite_number`, and the message names the submission, domain and metric.

**RAVF004, developer identity (medium).** `Z.ai` was unmapped and became developer `z.ai` with its own datastore directory, while `Zhipu AI` mapped to `zhipu-ai`. Provider names are now matched normalized, so a new spelling of a mapped provider cannot split it.

## The four older threads

All one shape: `methodology`, `voice_config`, `model_release`, `trajectory_files` and `references` were read through `x or {}`, which covers a missing block but not a malformed one, so a submission carrying a string there raised `AttributeError`. One `_mapping` guard covers them. Separately `--limit` sliced *after* every selected submission had been downloaded; it now bounds the download.

## What current conventions also required

- **Publication.** `save_evaluation_log` per record became `save_evaluation_logs`, which validates and serializes the batch before creating a file and removes what it created if a later write fails.
- **Accounting.** `convert_logs` returns a `SourceConversionResult`. A submission that raises is a failure and does not cost the rest of the batch; one the leaderboard lists with no score in any domain is an exclusion instead of a silent drop. `run` writes the report beside the output and exits non-zero on failures.
- **Automation.** `parse_args(argv)` so the catalog entry is checked against the real parser, a `tau_bench` entry writing the `tau-bench` collection, and the row in `adapters/README.md`. Raw capture comes free through `fetch_json`.

## Verification

`uv run pytest tests` 878 passed, 31 skipped — the adapter's own file went 5 → 16 tests, one per fix above. `uv run ruff check` clean.

@benshi34 @nelaturuharsha — if you would rather land this from the fork, say so and I will hand the commits over instead.


borgr:
Not needed — #192 itself now carries this work, pushed to @benshi34's branch through the maintainer-edit permission the PR already grants. The discussion belongs on #192's existing threads, not on a second PR.

PR #269, opened 2026-08-26
[Feature] Reviewed PwC DrugBank protocol adapter, rebased (supersedes #260)

borgr:
Supersedes #260, continuing @yananlong's four commits with `main` merged in — their authorship is the base of this history, not a rewrite of it. Opened here because #260's branch is on a fork I cannot push to.

## What the rebase had to decide

`main` gained the general Papers with Code adapter (#209) after this branch was written: `every_eval_ever/adapters/paperswithcode/`, over the same postgres dump, with a `paperswithcode = ["pgdumplib>=4.0.0"]` extra. This branch declared `paperswithcode-drugbank = ["pgdumplib>=4.0.0,<5.0.0"]` — the same package a second time, with a cap main does not have. Resolved by depending on the existing extra and dropping the duplicate, and pointing the two `uv run --extra` lines at it. One line to revert if separate extras are wanted.

## The overlap, and why it is not a blocker

The scheduled `paperswithcode` run is `--all --best-effort`, so it converts every dataset in the dump — DrugBank cells included, without the reviewed split and protocol semantics this adapter adds. The same measurement therefore appears in `data/paperswithcode/` and `data/paperswithcode-drugbank/`.

That is joinable rather than silent: both adapters record the PwC `evaluations` row id as `additional_details.pwc_evaluation_id` (`paperswithcode/adapter.py:884` and `paperswithcode_drugbank/adapter.py:692`), so a consumer reading both collections can tell one measurement from two. The last commit states this in the adapter docstring and the `adapters/README.md` row, which is what was missing — not the dedup itself.

If maintainers would rather have one adapter with a drugbank mode, or have the general adapter skip the reviewed cells, that is a larger change and belongs in its own PR against #256's dedup discussion.

## Verification

`uv run pytest tests` 951 passed, 32 skipped; `uv run ruff check` clean. The regenerated `uv.lock` has a zero-line diff against main, since the drugbank extra is gone and `pyyaml` was already a core dependency.

Labelled `auto-review`: #260 never got a pass — Copilot hit its quota and review-anvil never ran — so this is a first review, not a re-review.

@yananlong — say the word if you would rather land it from your fork and I will hand the commits over.


borgr:
Not needed — #260 itself now carries this work, pushed to @yananlong's branch through the maintainer-edit permission the PR already grants. Continuing on #260.

PR #271, opened 2026-08-27
Write down which name the datastore's developer folder comes from

borgr:
The publisher-directory question that has been blocking #230 turns out to be already decided, already implemented, and documented in exactly one place nobody looks. This writes it where readers land, and corrects one piece of guidance that pointed the wrong way.

## The rule

`data/<collection>/<developer>/<model>/` takes `<developer>` from **`model_info.id`'s prefix** — not from `model_info.developer`, and not from the eval-card-registry's `org_id`. `datastore_path_components` implements it (a slashed id decides both folders; `developer` is read only for a flat id), `check_model_identity_path` enforces it, and `adapters/lexam` and `adapters/open_medical_llm` both follow it. `lexam` says so in its own docstring.

It was written down once, in the skill's `registry.md`, as the tail of a bullet about model ids: *"Note the EEE datastore path takes the developer folder from the id prefix, not from `model_info.developer`."*

## Where it wasn't

- **The schema** described `developer` as "Name of organization that provides the model (e.g. 'OpenAI')", which reads as though that field picks the directory.
- **`datastore-gate.md` §path** listed the five components and their character rules without saying what `<developer>` refers to.
- **`datastore_path_components`** documented the flattening, not the choice.
- **`fields.md` pointed the wrong way.** It blamed per-adapter developer maps for the datastore holding one model under two orgs, citing `qwen` vs `alibaba` — but `get_developer` produces that pair by itself:

      Qwen/Qwen3-32B  -> Qwen          qwen3-32b     -> alibaba
      meta-llama/...  -> meta-llama    llama-3.1-8b  -> meta
      zai-org/GLM-4.6 -> zai-org       glm-4.6       -> zhipu-ai

  Prefix for a slashed name, company fold for a bare one. An author following "use the shared helper, not a private map" to the letter still splits the directory when their source writes bare names. That guidance now says to resolve a bare name through the registry and take the resolved canonical id's prefix, as `open_medical_llm` does.

## Two consequences now stated in §path

A record whose id prefix and `model_info.developer` disagree is usually **correct** — `Qwen/Qwen3-32B` with developer `alibaba` is the registry's own two-tier model, and "fixing" the prefix to match the company is the one thing `registry.md` tells you not to do. The real defect is one publisher reachable under two prefixes.

## What this settles elsewhere

**#230 is wrong at the root, not merely conflicting.** Its `check_developer_slug` warns whenever a publisher directory is a registry alias rather than the canonical org, which is the prefix-fixing this rule forbids — and under the two-tier model it would fire on correct records. That is why it turned `test_conversion_passes_the_merge_gate[inspect]` red on `mistral/` rather than finding a bug. I will close it against this, and the converter migration I was about to propose is not needed.

The live split in the published datastore is real and is a separate, narrower bug — `get_developer`'s two branches disagreeing. Filing that with the evidence above; it changes emitted data, so it wants its own PR and a human read.

No behaviour change here. `eval_types.py` is regenerated for the one description, through the exact `regenerate_types.yml` command chain. `uv run pytest tests` 859 passed, 31 skipped, including `test_skill_conversion.py`; `uv run ruff check` clean.


borgr:
Tightened after a fair objection about skill bloat.

The first pass added 18 lines to a skill that is already 850 across SKILL.md and seven references, and — worse — it left the `fields.md` bullet whose claim was wrong sitting directly above a second bullet correcting it. Two bullets, one of them misleading, is how a reference gets unusable.

Now: §path states the rule in four lines and names `check_model_identity_path` rather than re-explaining the registry's two-tier model, the `fields.md` bullet is **rewritten in place** rather than joined, and `registry.md` carries a pointer instead of a restatement. Net +6 lines on the skill, and one wrong claim removed.

The code docstrings carry the explanation instead, which is where someone reading `datastore_path_components` or `get_developer` will hit it anyway.

borgr:
Added `3ebf316db`, the same defect one field over.

`source_type`'s schema description was *"Whether the data comes from a direct evaluation run or from documentation"*, which does not decide the case people actually hit — raw per-item outputs that somebody else ran. The skill carried the sharper test (the artifact you hold, not who ran the compute) plus a parenthetical noting that the schema under-specifies it. A doc apologising for another doc is the same thing that happened with the developer folder, so the schema now states the test and #274 removes the apology.

Both description changes go through the same regeneration, so this PR carries one `eval_types.py` diff for the two.

